PR #5, opened 2018-06-19
Lsl copy phrase bug

lawhead:
* Updated copy_phrase to use the acquisition public properties; this fixes a bug where fs was not being set since it was getting a value of undefined; this was reported in Pivotal Tracker bug #158405692.

* Updated process_data_for_decision method to ensure that output np.array has dtype of float. All downstream code is expecting a float array parameter. This is the default return value when all channel data are floats, however, when we write markers directly to LSL, these values are strings and need to be explicitly converted to sentinel float values.

### DIFF
--- a/bcipy/helpers/bci_task_related.py
+++ b/bcipy/helpers/bci_task_related.py
@@ -1,6 +1,7 @@
 import os
 from psychopy import visual, event
 import numpy as np
+from typing import Any
 
 
 def fake_copy_phrase_decision(copy_phrase, target_letter, text_task):
@@ -75,7 +76,7 @@ def process_data_for_decision(sequence_timing, daq):
     """Process Data for Decision.
 
     Processes the raw data (triggers and eeg) into a form that can be passed to
-        signal processing and classifiers.
+    signal processing and classifiers.
 
     Parameters
     ----------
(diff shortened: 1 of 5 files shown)


bcipy/helpers/bci_task_related.py
```diff
@@ -122,9 +122,20 @@ def process_data_for_decision(sequence_timing, daq):
             if len(raw_data) < data_limit:
                 raise Exception("Not enough data received")
 
-        # Take only the sensor data from raw data and transpose it
-        raw_data = np.array([np.array(raw_data[i][0]) for i in
-                             range(len(raw_data))]).transpose()
+
+        def float_val(col):
```
tab-cmd: This can be moved outside of this function.

bcipy/helpers/bci_task_related.py
```diff
@@ -122,9 +122,20 @@ def process_data_for_decision(sequence_timing, daq):
...
+            """Convert marker data to float values so we can put them in a
+            typed np.array. The marker column has type float if it has a 0.0
+            value, and would only have type str for a marker value."""
+            if type(col) is str:
```
tab-cmd: use isinstance. 

```
if isinstance(col, str):
```

bcipy/helpers/bci_task_related.py
tab-cmd: This can be of any type correct? Or at least str/float? Add typing to it

```
from typing import Any
```

tab-cmd (approved):
Some small things to change

bcipy/helpers/bci_task_related.py
lawhead: Thanks, good suggestions.

PR #14, opened 2018-07-16
Rsvp calibration breaks

tab-cmd:


bcipy/bci_tasks/rsvp/calibration.py
```diff
@@ -103,6 +108,24 @@ def execute(self):
                                       self.wait_screen_message_color):
                     break
 
+                #Take a break every number of trials defined in parameters.json
+                if self.enable_breaks:
+                    if not self.trials_before_break == 0:
```
tab-cmd: This check isn't needed

bcipy/bci_tasks/rsvp/calibration.py
```diff
@@ -103,6 +108,24 @@ def execute(self):
...
+                        #Check whether any trials have taken place, and, if so,
+                        #whether the number of trials performed is divisible
+                        #by the number of trials before a break set in parameters
+                        if (idx_o != 0) and (idx_o % self.trials_before_break) == 0:
+                            #Update countdown every second
+                            for counter in range(0,self.break_len):
```
tab-cmd: This can be `for counter in range(self.break_len):`

bcipy/bci_tasks/rsvp/calibration.py
```diff
@@ -103,6 +108,24 @@ def execute(self):
...
+                                time = self.break_len - counter
+                                message = f'{self.break_message} {time}s'
+                                self.rsvp.update_task_state(
+                                    text=(message),
```
tab-cmd: does message need to be in brackets? If not, remove

bcipy/bci_tasks/rsvp/calibration.py
```diff
@@ -103,6 +108,24 @@ def execute(self):
...
+                                    color_list=task_color[idx_o])
+                                self.rsvp.draw_static()
```
tab-cmd: We generally don't flip or draw static inside our tasks. 

Let's make a helper function in `bci_task_related` like we do for the `get_user_input`. 

It should be called if breaks are enabled and take in the current index, trials before break, rsvp display, and break length. Then we can just import the function everytime we need it

bcipy/bci_tasks/rsvp/calibration.py
```diff
@@ -72,6 +72,11 @@ def __init__(self, win, daq, parameters, file_save, fake):
         self.is_txt_sti = parameters['is_txt_sti']
         self.eeg_buffer = parameters['eeg_buffer_len']
 
+        self.enable_breaks = parameters['enable_breaks']
+        self.break_len = parameters['break_len']
```
tab-cmd, line 76: 👍

PR #21, opened 2018-08-11
Language model

tab-cmd:


shiranD:
One comment I have is regarding this file: https://github.com/BciPy/BciPy/blob/language_model/bcipy/helpers/lang_model_related.py
I haven't written it and it seems a good script for the old LM, but not a suitable one for the current OCLM. Just heads up if BciPy highly relies on it - this could be the first place to check.

tab-cmd:
Hi @shiranD, I wrote that blob when we implemented the old lm bc you had a lot going on and we wanted to test the integration quickly. I'd like for this PR to include the necessary integration logic as well. You will likely need to add a parameter dictating which lm to use and then route it to the correct lm class.  Extending what I've written is okay or if you think something a little fancier would be better to interface  :) that's ok as well. This might be a good discussion to have with Steven (does he have a Github account we should add?)

The only thing you will have to worry about is:

- Including all the necessary params in the `parameters.json`. Most of them, except for supporting this lm, should already be there.
- If you change using the lang_model_related helper, be sure whatever you do is implemented in `bci_main`.
- If you use the helper, adding the if statements and routing to the correct class with all the needed params.


If you start doing this and something isn't straightforward please slack me!

bcipy/oclm_language_model/demo/eeg_utils.py
```diff
@@ -0,0 +1,67 @@
+from __future__ import division
```
tab-cmd, line 1: please delete this and consider using MNE or if random is okay the fake data  generated from the acquisition module: http://martinos.org/mne/stable/auto_examples/index.html#real-time-m-eeg-acquisition

PR #27, opened 2018-10-09
Alert Tone Task

lawhead:


bci_main.py
```diff
@@ -58,11 +59,17 @@ def execute_task(task_type: dict, parameters: dict, save_folder: str) -> bool:
         save_folder (str): path to save folder
     """
 
+    exp_type = ExperimentType(task_type['exp_type'])
+
+    # Initialize Display Window
+    display = init_display_window(parameters)
```
tab-cmd, line 65: I know it's tempting to have the window come first, however when loading all these modules if something causes an exception to be thrown (depending on the exception) it causes the screen to freeze in a unrecoverable state. If not in full screen or two monitors used not a problem.

bcipy/display/display_main.py
```diff
@@ -57,6 +57,14 @@ def init_display_window(parameters):
     return display_window
 
 
+def print_message(window: visual.Window, message: str="Initializing..."):
```
tab-cmd: This should go in a helper. It could replace: `trial_complete_message`

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -110,75 +97,82 @@ def load_items_from_txt(self, event):
...
-gui.add_image(
-    path='bcipy/static/images/gui_images/ohsu.png', position=(10, 0), size=100)
-gui.add_image(
-    path='bcipy/static/images/gui_images/neu.png', position=(530, 0), size=100)
-
-
-# Make the GUI Show now
-gui.show_gui()
-app.MainLoop()
+def main():
```
tab-cmd: maybe call this function `run_rsvp_gui`?

bcipy/parameters/parameters.json
```diff
@@ -289,6 +289,14 @@
     "recommended_values": "",
     "type": "directorypath"
   },
+  "alert_sounds_path": {
```
tab-cmd, line 292: We could have a more generic path to sounds directory parameter? I could imagine this implementation being used elsewhere. This is more explicit which I imagine the clinical team would like.

bcipy/tasks/rsvp/alert_tone_calibration.py
```diff
@@ -0,0 +1,56 @@
+"""Calibration Task that uses alert tones to help the user retain focus."""
+
+from bcipy.helpers.stimuli_generation import play_sound, soundfiles
+from bcipy.tasks.task import Task
+from bcipy.tasks.rsvp.calibration import RSVPCalibrationTask
+
+TASK_NAME = 'RSVP Alert Tone Calibration Task'
```
tab-cmd: I agree this should be a static variable, but I think it should belong to the class and called via self.

bcipy/tasks/task_registry.py
```diff
@@ -0,0 +1,68 @@
...
+
+# NOTE:
+# In the future we may want to consider dynamically retrieving all subclasses
+# of Task and use these to populate a registry. We could also provide
+# functionality for bcipy users to define their own tasks and register them so
+# they would appear as options in the GUI.
+#
+# However, this approach is currently problematic for the GUI interface. Due
+# to the tight coupling of the display code with the Tasks, importing any of
+# the Task subclasses causes pygame (a psychopy dependency) to create a GUI,
```
tab-cmd, line 13: on setup we should be `pip uninstall pygame`. We don't need it at all and PsychoPy only keeps it for historical reasons

bcipy/tasks/rsvp/icon_to_icon.py
```diff
@@ -20,6 +20,8 @@
 import logging
 from os import path
 
+TASK_NAME = 'RSVP Icon to Icon Task'
```
tab-cmd: Same here as in the Tone Task

tab-cmd (approved):
A nice refactor of a few areas and the tone task implementation is solid. Take these as suggestions as needed and merge when ready.

bcipy/display/display_main.py
lawhead: Good suggestion. I think it's sufficiently different from trial_complete_message to warrant having both for now.

bci_main.py
lawhead: Hm, interesting. I have not encountered this yet. I did some testing and tried to throw exceptions at various places in the code and the window seemed to cleanup okay. Do you have a more specific example of when you have had this issue? 

The main problems I was trying to solve here was first of all, to feel more responsive during loading, since there is a lot of initialization that happens and it is easy to think that the click did not happen correctly. Secondly, I wanted to fail immediately in the main thread if the user selected an invalid experiment type, rather than waiting for all that loading to occur and have to clean it all up (which requires waiting for multiple threads to stop).

My inclination is to leave it as is for now and then fix the initialization issues as they come, but if you think that would make the system harder to manage let me know and I can revert back.

bcipy/tasks/rsvp/alert_tone_calibration.py
lawhead: Sounds good, thanks.

bcipy/tasks/task_registry.py
lawhead: Good to know.

PR #29, opened 2018-10-19
Alert tone task delay

lawhead:
#161291446 ; alert tone task sound delay; adds a parameterized delay between playing a sound and presenting the target.

### DIFF
--- a/bci_main.py
+++ b/bci_main.py
@@ -66,10 +66,6 @@ def execute_task(task_type: dict, parameters: dict, save_folder: str) -> bool:
 
     exp_type = ExperimentType(task_type['exp_type'])
 
-    # Initialize Display Window
-    display = init_display_window(parameters)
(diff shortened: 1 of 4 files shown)


bci_main.py
```diff
@@ -104,6 +100,13 @@ def execute_task(task_type: dict, parameters: dict, save_folder: str) -> bool:
         language_model = None
         filename = None
 
+    # Initialize Display Window
+    # We have to wait until after the prompt for the prompt to load the signal
+    # model before displaying the window, otherwise in fullscreen mode this 
+    # throws an error.
+    display = init_display_window(parameters)
+    print_message(display, "Initializing...")
+
```
lawhead, line 109: @sci-tab , you were right about this one. I added a comment to document why display has to go here in case this gets refactored later. The new language model takes a long time to load so it may be worth exploring some other options.

PR #39, opened 2019-01-15
Added analysis channels for DSI VR-300

lawhead:


bcipy/helpers/data_vizualization.py
```diff
@@ -19,7 +19,8 @@ def generate_offline_analysis_screen(
         fs=300,
         plot_x_ticks=8,
         plot_average=False,
-        show_figure=False) -> None:
+        show_figure=False,
+        channel_names = None) -> None:
```
tab-cmd, line 24: You could default to an empty dict here

tab-cmd (approved):

bcipy/helpers/data_vizualization.py
lawhead: Yeah, I was avoiding a pylint dangerous-default-value warning, but in this case the channel_names dict is not mutated so a default arg would be fine.

PR #40, opened 2019-01-16
Level Feedback Display and Task

tab-cmd:
- Level Feedback
- Clean up bar graph
- Inter Sequence Feedback Task
- Move LM tests to integration


bcipy/parameters/parameters.json
```diff
@@ -630,14 +646,54 @@
...
+  "feedback_padding": {
+    "value": "0.37",
+    "section": "feedback_config",
+    "readableName": "feedback_padding",
+    "helpTip": "feedback_padding",
+    "recommended_values": "",
+    "type": "float"
+  },
+  "feedback_target_line_width": {
+    "value": "100",
```
lawhead, line 711: I tried this locally and 100 seemed way too wide; maybe 10?

bcipy/feedback/visual/level_feedback.py
```diff
@@ -0,0 +1,114 @@
...
+    def _determine_height_and_width(self):
+        return self.width, self.height
+
+    def _validate_levels(self):
+        # validate the number of levels and colors supplied are equal
+        if len(self.level_colors) != self.level_number:
+            raise ValueError('Level colors must be equal to the number of levels')
+
+
+if __name__ == '__main__':
```
lawhead: I like the demo code here; this is helpful.

bcipy/feedback/visual/level_feedback.py
```diff
@@ -0,0 +1,114 @@
...
+    from bcipy.helpers.load import load_json_parameters
```
lawhead, line 126: I like the demo here; this is helpful.

bcipy/feedback/visual/level_feedback.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def administer(self, position=1):
+        # construct the bar indicator and validate the levels / colors supplied
+        #   are accruate
+        self._construct_bar_indicator()
+        self._validate_levels()
+
+        # indexing starts at zero but position does not ;)
+        position -= 1
+
```
lawhead, line 51: It seems like this class could be simplified somewhat. For instance, instead of maintaining separate values for level_number and level_colors and then validating that output is the correct length, we could just enumerate through the colors to construct the rectangles. Also, is there any reason to store the stimuli? And instead of maintaining state regarding the position-x, this could be calculated based on the index.

So maybe something like:

```        
pos_x = self.parameters['feedback_pos_x']

# loop through the levels wanted and construct the stimuli
for index, color in enumerate(self.level_colors):
    rect = visual.Rect(
        win=self.display,
        width=self.width,
        height=self.height,
        fillColor=color,
        lineColor=self.feedback_indicator_color if position == index else self.line_color,
        pos=((pos_x + (index * self.padding)), self.position_y),
        lineWidth= self.target_line_width if position == index else self.line_width,
        ori=0.0)
    rect.draw()
```

bcipy/tasks/rsvp/calibration/inter_sequence_feedback_calibration.py
```diff
@@ -0,0 +1,176 @@
...
+                # Do the sequence
+                last_sequence_timing = self.rsvp.do_sequence()
+
+                # Write triggers for the sequence
+                _write_triggers_from_sequence_calibration(
+                    last_sequence_timing, self._task.trigger_file)
+
+                # TODO implement feedback decision maker
+                position = self._get_feedback_decision()
+                timing = self.visual_feedback.administer(position=position)
```
lawhead: It looks like this is the only code that differs from the calibration.py execute method. We could add a hook in the Calibration class to call a present_feedback method that could be overridden or externally provided. Then this code could be executed in the callback and the _task.execute method could be used as is.

lawhead (approved):

bcipy/parameters/parameters.json
lawhead: Never mind; this seems very display-specific and 100 was too narrow during testing (I believe 150 worked better). So probably just leave this as is for now.

bcipy/feedback/visual/level_feedback.py
tab-cmd: done

PR #41, opened 2019-01-16
Freq decomposition

tab-cmd:


bcipy/signal/processing/decomposition/psd.py
```diff
@@ -0,0 +1,80 @@
...
+        data: np.ndarray,
+        band: tuple,
+        fs: float=100,
+        window_length: float=4,
+        plot: bool=False,
+        method: PSD_TYPE=PSD_TYPE.MULTITAPER,
+        relative=False):
+    """Power spectral density:
+
+    Many thanks to: https://raphaelvallat.github.io/bandpower.html
```
lawhead, line 34: It might be more clear to name the parameter window_seconds, or to provide parameter documentation. Also, it seems like the types of the default arguments should match the declared variable types (ex. 4.0).

bcipy/signal/processing/decomposition/psd.py
```diff
@@ -0,0 +1,80 @@
...
+from scipy import signal
+from scipy.signal import welch
+from scipy.integrate import simps
+from mne.time_frequency import psd_array_multitaper
+
+
+class PSD_TYPE(Enum):
+    WELCH = 'Welch'
+    MULTITAPER = 'MutliTaper'
+
```
lawhead, line 22: The referenced blog article has some good points about why you might want to choose one PSD_TYPE over the other (tradeoffs in variance, frequency resolution, computational cost). It may be worth re-stating some of those points here or in the power_spectral_density docstring.

bcipy/parameters/parameters.json
```diff
@@ -731,6 +731,17 @@
...
   },
+  "psd_method": {
+    "value": "welch",
+    "section": "bci_config",
+    "readableName": "Power Spectral Density Method",
+    "helpTip": "The method used to approximate power spectral density bands",
+    "recommended_values": [
+      "welch",
+      "multitaper"
+    ],
```
lawhead, line 742: Do these values need to be capitalized to match the enum strings?

lawhead (approved):

bcipy/parameters/parameters.json
tab-cmd: Done. I have no way to connect the two yet, but I think that makes sense. This will come in a later PR which connects this and #40 .

bcipy/signal/processing/decomposition/psd.py
tab-cmd: I don't want to do too much educating here since there is more to discuss than even the article alludes to, but I did point to the trade off.

bcipy/signal/processing/decomposition/psd.py
tab-cmd: done

PR #44, opened 2019-02-16
Query selection

azizkocana:
This branch covers the new query selection methods for RSVPKeyboard. 

In order to incorporate different query paradigms, `bcipy.tasks.query_selection` is created. This module also includes random querying. Right now each query actor is an **object** (previously was a function). We also want to keep track of the trajectory of decision, hence a query module should also be able to keep track of history and its own parameters. 

To test the branch `branch_test` folder is created and will be deleted before merge. This folder contains 2 key elements;

1. `oracle:` a simulated user with kernel density estimates. Uses a calibration data to mimic EEG collected from a user. A better oracle model was already presented in the previous bugfixes, this oracle only returns a scalar score value.
2. `demo_query_method:` basically a smoke test. Runs the inference with the decision maker and the oracle.

@lawhead @sci-tab , I will update the demo file and the code based on your preference.


bcipy/tasks/main_frame.py
```diff
@@ -228,15 +238,39 @@ def decide_state_update(self):
         self.list_epoch[-1]['decision'] = decision
         return decision
 
-    def prepare_stimuli(self):
+    # TODO: think about where timing, color, is_txt, num_sti should go.
+    # e.g. is_txt is predefined with the alphabet directly
+    def _prepare_stimuli(self, timing=[1, 0.2], color=['red', 'white'],
```
lawhead, line 240: Now that prepare_stimuli is private to the class, we should remove it from the class documentation.

bcipy/tasks/main_frame.py
```diff
@@ -228,15 +238,39 @@ def decide_state_update(self):
...
             Return:
                 stimuli(tuple[list[char],list[float],list[str]]): tuple of
                     stimuli information. [0]: letter, [1]: timing, [2]: color
                 """
-        stimuli = \
-            best_case_rsvp_seq_gen(self.alphabet, self.list_epoch[-1][
-                'list_distribution'][-1], num_sti=1, is_txt=self.is_txt_sti,
-                timing=self.stimuli_timing)
+
+        tmp = self.list_epoch[-1]['list_distribution'][-1][:]
```
lawhead: There are several places in the code that use this selector `self.list_epoch[-1]['list_distribution'][-1]`. It would clarify what this is doing if this was pulled out into another method with a more descriptive name (latest_prob_distribution?) and some documentation. Then you probably don't even need this tmp variable.

bcipy/tasks/main_frame.py
```diff
@@ -228,15 +238,39 @@ def decide_state_update(self):
...
+        stimuli_pos = self.query_method.update_query(tmp[:])
+
+        samples, times, colors = [], [], []
+        for idx_num in range(num_sti):
+            # append a fixation cross. if not text, append path to
+            # image fixation
+            if is_txt:
+                sample = ['+']
+            else:
+                sample = ['bcipy/static/images/bci_main_images/PLUS.png']
```
lawhead, line 259: This is another pattern that we have throughout the code base that would be good to have consolidated in a single place. It would be good to add a helper function (in bci_task_related.py?) with something like:

        def with_fixation(sample: List, is_text: bool = True) -> List:
            """Return a copy of sample with fixation item prepended."""
            fixation = '+' if is_txt else 'bcipy/static/images/bci_main_images/PLUS.png'
            return [fixation] + sample

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,190 @@
...
+                updates the querying method and picks the particular query
+        """
+
+    def __init__(self, len_alp, len_query=4):
+        self.alp = list(range(len_alp))
+        self.len_query = len_query
+
+    def reset(self):
+        """ This object does not require history. Reset is a void function. """
+        tmp = None
```
lawhead: For functions that don't do anything, just use the keyword `pass` (see Task.py).

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,190 @@
...
```
lawhead: Same goes here for using `pass`.

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def reset(self, lam=.9):
+        """ resets the history related items in the query method
+            Args:
+                lam(float): in [0,1] convex combination parameter """
+        self.lam = lam
+        self.momentum = np.zeros(len(self.alp))
+        self.prob_history = []
+
+    def update_query(self, p):
```
lawhead, line 119: We should have more descriptive variable names. Also, rather than adding the type information to the documentation, we can add it inline now. So this becomes:

    from typing import List
    def update_query(self, p: List[float]) -> List[int]:
        ...

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,190 @@
...
+        """ with the final belief over the system, updates the querying method
+            and generates len_query most likely queries.
+            Args:
+                p(list[float]): list of probability distribution over the state
+                    estimates
+            Return:
+                pos_query(list[int]): positions of queries in alphabet """
+        tmp = p[:]
```
lawhead, line 127: Rather than `tmp`, maybe call this something like `probabilities`?

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,190 @@
...
+        if len(self.prob_history) >= 2:
+            # only update momentum for previous terms
+            idx_prev_query = [self.alp.index(self.last_query[i]) for i in
+                              range(len(self.last_query))]
+            self.momentum *= self.gam
+
+            for k in idx_prev_query:
+                # momentum = current_mass * mass_displacement
+                self.momentum[k] += self.prob_history[-1][k] * np.power(10,
+                                                                        5) * (
```
lawhead: There's some odd formatting on this line. Are you using a code formatter for this? I switch back and forth between autopep8 and yapf.

bcipy/tasks/main_frame.py
azizkocana: I removed documentation. 

Stimuli is generated for the main_frame and it is not required to be called outside. Therefore, I switched it to a private class. However, I am not %100 sure about my decision.

bcipy/tasks/query_selection.py
azizkocana: I fixed the formatting.

I am using auto-pep8 in Pycharm.

bcipy/tasks/query_selection.py
azizkocana: `tmp` appears due to assignment problem in python. This is literally a temporary element in order to avoid overwriting of the original element. Therefore, I prefer to keep it as is.

bcipy/tasks/main_frame.py
azizkocana: named as `last_prob_dist`

bcipy/tasks/main_frame.py
tab-cmd: I think it should go outside. It's easier to test in isolation.

bcipy/tasks/query_selection.py
tab-cmd: He's saying to give it a name to help readers understand what you are assigning.

bcipy/tasks/query_selection.py
```diff
@@ -0,0 +1,189 @@
+""" Query selection module. Each query method is a class to keep track of the
```
tab-cmd, line 1: This shouldn't go in tasks....

branch_test/demo_query_method.py
```diff
@@ -0,0 +1,271 @@
+import os
```
tab-cmd, line 1: `branch_test` should be removed...

branch_test/oracle.py
```diff
@@ -0,0 +1,87 @@
+import numpy as np
+from scipy.stats import norm, iqr
+from sklearn.neighbors.kde import KernelDensity
+from sklearn.metrics import auc, roc_curve
+import string
+
+
+class BinaryRSVPOracle(object):
```
tab-cmd, line 8: I've seen you implement this a few times. It could be moved somewhere.

tab-cmd (changes requested):

branch_test/oracle.py
azizkocana: I implemented a more general simulator using RSVP data for the debugging during August. I totally agree need a place for this implementation, which can serve as a broad demo of the system.

branch_test/demo_query_method.py
azizkocana: I will remove it once it is tested and approved by the clinical team. It helps debugging of newly created modules.

PR #45, opened 2019-02-17
Signal viewer

tab-cmd:


bcipy/gui/viewer/filter.py
```diff
@@ -0,0 +1,51 @@
+"""Defines filter functions that can be applied to a viewer stream."""
+import numpy as np
+from bcipy.signal.process.filter import bandpass, notch
+# from bcipy.signal.processing.sig_pro import sig_pro
+
+
+def downsample_filter(factor: int, fs: int):
```
tab-cmd: I think any downsampling / signal modification should happen from the signal processing module. We could add a downsampling function there.

bcipy/gui/viewer/filter.py
```diff
@@ -0,0 +1,51 @@
+"""Defines filter functions that can be applied to a viewer stream."""
+import numpy as np
+from bcipy.signal.process.filter import bandpass, notch
+# from bcipy.signal.processing.sig_pro import sig_pro
```
tab-cmd: This can be removed.. it still exists, just as a text_filter.

bcipy/gui/viewer/ring_buffer.py
```diff
@@ -0,0 +1,40 @@
+"""Defines a RingBuffer with a fixed size; when full additional elements
```
tab-cmd, line 1: This could be moved elsewhere. It's something that might be used outside of viewing data.

bcipy/gui/viewer/pubsub_viewer.py
```diff
@@ -0,0 +1,359 @@
...
+                    logging.debug(("Max attempts to retrieve data exceeded.",
+                                   "Stopping DataPublisher"))
+                    pub.sendMessage(CMD_CHANNEL, msg=STOP_CMD)
+                    return
+                time.sleep(0.1)
+
+        logging.debug("Stopping EEG Viewer DataPublisher.")
+
+
+class EegFrame(wx.Frame):
```
tab-cmd: I would write `EEGFrame`

bcipy/gui/viewer/data_viewer.py
```diff
@@ -0,0 +1,395 @@
...
+
+from bcipy.acquisition.device_info import DeviceInfo
+from bcipy.acquisition.util import StoppableProcess
+from bcipy.gui.viewer.data_source import QueueDataSource
+from bcipy.gui.viewer.launcher import Launcher
+from bcipy.gui.viewer.ring_buffer import RingBuffer
+from bcipy.gui.viewer.filter import downsample_filter, sig_pro_filter
+
+
+class EegFrame(wx.Frame):
```
tab-cmd: This could be `EEGFrame`

bcipy/gui/viewer/file_streamer.py
```diff
@@ -0,0 +1,43 @@
...
+            Used to communicate messages to the main thread.
+    """
+
+    def __init__(self, data_file, data_queue):
+        super(FileStreamer, self).__init__()
+        self.data_file = data_file
+        self.data_queue = data_queue
+
+    def run(self):
+        print("Starting streamer")
```
tab-cmd: use logging

tab-cmd (commented):
Nice work here. I would transfer most print statements to logging. I'd also add a unique identifier in the log message as was done in the feedback module to make them easier to find when needed.

PR #50, opened 2019-04-12
#164785900 ; session cleanup ordering

lawhead:
#164785900 ; temporary work-around to allow acquisition to shutdown properly when using a USB-C external monitor for dual display.

### DIFF
--- a/bci_main.py
+++ b/bci_main.py
@@ -116,8 +116,6 @@ def execute_task(task_type: dict, parameters: dict, save_folder: str) -> bool:
 
 def _clean_up_session(display, daq, server):
     """Clean up session."""
-    # Close the display window
-    display.close()
 
     # Stop Acquisition
     daq.stop_acquisition()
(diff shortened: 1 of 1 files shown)


bci_main.py
```diff
@@ -126,6 +124,12 @@ def _clean_up_session(display, daq, server):
     if server:
         server.stop()
 
+    # Close the display window
+    # NOTE: There is currently a bug in psychopy when attempting to shutdown
+    # windows when using a USB-C monitor. Putting the display close last in
+    # the sequence allows acquisition to properly shutdown.
+    display.close()
```
tab-cmd, line 131: I agree with moving. To prevent hanging screens, we might consider wrapping this and forcing a kill in another way if it fails to close. What error is thrown when it tries to close with the USB-C monitor? We can merge this as is but worth looking into.

lawhead: I don't think it ever actually throws an exception, it just hangs. The current workaround is to force with Ctrl-C. I think it only happens in fullscreen mode. Next time I'm in the lab I'll confirm this behavior. If I can duplicate with a small script I will report the issue to the psychopy project.

PR #56, opened 2019-05-06
Decision criteria

lawhead:
#165452744 ; Refactored DecisionMaker to use a CriteriaEvaluator to determine when to commit to a decision based on a set of configurable criteria.

### DIFF
--- a/bcipy/tasks/rsvp/main_frame.py
+++ b/bcipy/tasks/rsvp/main_frame.py
@@ -2,8 +2,12 @@
 
 from bcipy.helpers.stimuli import best_case_rsvp_seq_gen
 from bcipy.helpers.task import SPACE_CHAR
+from typing import Dict, List
+import logging
 import numpy as np
 import string
+
+log = logging.getLogger(__name__)
 
 
 class EvidenceFusion(object):
@@ -88,6 +92,9 @@ class DecisionMaker:
                         arrays with prob. dist. over alp
             seq_constants(list[str]): list of letters which should appear in
                 every sequence.
+            criteria_evaluator: CriteriaEvaluator - optional parameter to
+                provide alternative rules for committing to a decision.
+
         Functions:
             decide():
                 Checks the criteria for making and epoch, using all
@@ -115,7 +122,8 @@ class DecisionMaker:
                  alphabet=list(string.ascii_uppercase) + ['<'] + [SPACE_CHAR],
                  is_txt_stim=True,
                  stimuli_timing=[1, .2],
-                 seq_constants=None):
+                 seq_constants=None,
+                 criteria_evaluator=None):
         self.state = state
         self.displayed_state = self.form_display_state(state)
         self.stimuli_timing = stimuli_timing
@@ -139,6 +147,8 @@ class DecisionMaker:
 
         # Items shown in every sequence
         self.seq_constants = seq_constants
+
+        self.criteria_evaluator = criteria_evaluator or CriteriaEvaluator.default()
 
     def reset(self, state=''):
         """ Resets the decision maker with the initial state
@@ -189,18 +199,19 @@ class DecisionMaker:
 
         self.list_epoch[-1]['list_distribution'].append(p[:])
 
-        # Check stopping criteria
-        if (self.sequence_counter + 1) < self.min_num_seq or \
-                not ((self.sequence_counter + 1) >= self.max_num_seq or
-                     np.max(p) > self.posterior_commit_threshold):
+        params = dict(
+            min_num_seq=self.min_num_seq,
+            max_num_seq=self.max_num_seq,
+            threshold=self.posterior_commit_threshold)
 
-            stimuli = self.schedule_sequence()
-            commitment = False
-            return commitment, {'stimuli': stimuli}
-        else:
+        # Check stopping criteria
+        if self.criteria_evaluator.should_commit(self.list_epoch[-1], params):
             self.do_epoch()
-            commitment = True
-            return commitment, []
+            return True, []
+        else:
+            stimuli = self.schedule_sequence()
+            return False, {'stimuli': stimuli}
+
 
     def do_epoch(self):
         """ Epoch refers to a commitment to a decision.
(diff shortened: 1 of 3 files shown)


bcipy/tasks/tests/test_decision_maker.py
```diff
@@ -0,0 +1,111 @@
...
+        likelihood = np.array([0.05, 0.05, 0.1, 0.8])
+        decision, _arg = decision_maker.decide(likelihood)
+        self.assertFalse(decision, "Must exceed threshold")
+        self.assertEqual('..', decision_maker.state)
+
+        likelihood = np.array([0.05, 0.05, 0.09, 0.81])
+        decision, _arg = decision_maker.decide(likelihood)
+        self.assertTrue(decision)
+        self.assertEqual('..d', decision_maker.state)
+        self.assertEqual('d', decision_maker.displayed_state)
```
lawhead, line 38: I wrote these tests prior to the refactor to ensure that the functionality remained the same. I noticed that the threshold is currently doing a strictly greater than comparison. Should this be greater than or equal to?

tab-cmd: Great question. I'd bring that up at today's meeting!

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -143,6 +148,8 @@ def __init__(self,
         # Items shown in every sequence
         self.seq_constants = seq_constants
 
+        self.criteria_evaluator = criteria_evaluator or CriteriaEvaluator.default()
```
tab-cmd: I think it may be clearer to move this default to the constructor as opposed to using None and then setting it here.

bcipy/tasks/tests/test_evidence_fusion.py
```diff
@@ -0,0 +1,79 @@
...
+import unittest
+import numpy as np
+import bcipy.tasks.rsvp.main_frame as mf
+from bcipy.helpers.load import load_json_parameters
+
+
+class TestEvidenceFusion(unittest.TestCase):
+    """Tests for EvidenceFusion"""
+
+    def setUp(self):
```
tab-cmd: This can be removed if nothing is setup

bcipy/tasks/tests/test_evidence_fusion.py
```diff
@@ -0,0 +1,79 @@
...
+            self.assertSequenceEqual(list(erp_evidence[i]), list(val))
+
+        for i, val in enumerate(conjugator.evidence_history['FRP']):
+            self.assertSequenceEqual(list(frp_evidence[i]), list(val))
+
+        for i, val in enumerate(conjugator.likelihood):
+            self.assertAlmostEqual(expected_posterior[i], val)
+
+        # Reset the conjugator before starting a new epoch for clear history
+        conjugator.reset_history()
```
tab-cmd: We could remove this call and add another test for that method. In this case, it's not providing much to the test of fusion.

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -253,3 +261,77 @@ def prepare_stimuli(self):
...
+    Parameters:
+    -----------
+        continue_criteria: list of functions; if any of these evaluate to true the
+            decision maker continues.
+        commit_criteria: list of functions; if any of these return true and
+            continue_criteria are all false, decision maker commits to a decision.
+    """
+
+    def __init__(self, continue_criteria: list, commit_criteria: list):
+        self.continue_criteria = continue_criteria or []
```
tab-cmd: Same comment as above about pulling the default into the init.

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -253,3 +261,77 @@ def prepare_stimuli(self):
...
+            continue_criteria=[less_than_min_iterations],
+            commit_criteria=[exceeds_max_iterations, meets_commit_threshold])
+
+    def should_commit(self, epoch: Dict, params: Dict):
+        """Evaluates the given epoch; returns true if stoppage criteria has
+        been met, otherwise false.
+
+        Parameters:
+        -----------
+            epoch - Epoch data
```
tab-cmd: Some explanation of what would be needed for an `epoch` may be helpful here. It looks like we're using only the`list_distribution`. If we won't need anything else, we could also have them pass that in directly.

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -253,3 +261,77 @@ def prepare_stimuli(self):
...
+        """
+        if any(criteria(epoch, params) for criteria in self.continue_criteria):
+            return False
+        return any(
+            criteria(epoch, params) for criteria in self.commit_criteria)
+
+
+# Criteria
+
+def less_than_min_iterations(epoch, commit_params):
```
tab-cmd: There is a high possibility of needing these criteria outside of RSVP (also the `CriteriaEvaluator `). It may be worth thinking of where we could pull this out too.

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -253,3 +261,77 @@ def prepare_stimuli(self):
...
+    
+    Parameters:
+    -----------
+        continue_criteria: list of functions; if any of these evaluate to true the
+            decision maker continues.
+        commit_criteria: list of functions; if any of these return true and
+            continue_criteria are all false, decision maker commits to a decision.
+    """
+
+    def __init__(self, continue_criteria: list, commit_criteria: list):
```
tab-cmd: If we made an abstract `DecisionCriteria` with an `apply` method (not attached to the names here!), we could then extend it and place our business logic in `apply`. We could then explicitly declare what we expect in our list: `List[DecisionCriteria]`. It's an improvement to have them as methods, but a further abstraction could prove useful as we continue adding criteria.

bcipy/tasks/tests/test_decision_maker.py
```diff
@@ -0,0 +1,111 @@
...
+        """Test min sequences criteria."""
+
+        decision_maker = mf.DecisionMaker(
+            min_num_seq=2,
+            max_num_seq=10,
+            state='',
+            alphabet=['a', 'b', 'c', 'd'],
+            decision_threshold=0.8)
+
+        likelihood = np.array([0.05, 0.05, 0.09, 0.81])
```
tab-cmd, line 50: This test case might benefit from a setup method. We could pull out a default likelihood, decision maker etc which the unit tests could alter if needed.

tab-cmd (approved):

lawhead:
@sci-tab, thanks for the review. I'll address these comments prior to merging.

bcipy/tasks/rsvp/main_frame.py
lawhead: I think I originally did it this way to avoid putting the source code of the CriteriaEvaluator the DecisionMaker, since it must be declared before its reference. But I agree that it's better as a default param if you don't mind all this coming first. We could also factor this out into a separate module.

bcipy/tasks/rsvp/main_frame.py
lawhead: I considered doing this but didn't want to over-engineer things. But you make a good argument about the documentation aspect of this approach. I'll give it a try and see how it looks.

bcipy/tasks/rsvp/main_frame.py
lawhead: Our current criteria only uses list_distribution, but I want to make all the information available to future criteria. I will add some documentation to the epoch param. It should probably be an object rather than a dict at this point.

bcipy/tasks/tests/test_evidence_fusion.py
lawhead: Good catch. Copy-paste error.

bcipy/tasks/rsvp/main_frame.py
lawhead: I refactored into objects and I think it turned out well. Thanks for the suggestion.

PR #58, opened 2019-06-03
Inter-sequence feedback variables

tab-cmd:
# Inter-sequence feedback variables

## Added

- Feedback parameters for level thresholds and psd range

## Updated

- Fixed documentation typos
- Fixed variable names post-display refactor

### DIFF
--- a/bcipy/tasks/rsvp/calibration/inter_sequence_feedback_calibration.py
+++ b/bcipy/tasks/rsvp/calibration/inter_sequence_feedback_calibration.py
@@ -26,11 +26,11 @@ class RSVPInterSequenceFeedbackCalibration(Task):
 
     Calibration task performs an RSVP stimulus sequence to elicit an ERP.
     Parameters will change how many stim and for how long they present.
-    Parameters also change color and text / image inputs.
+    Parameters also change color and text / image inputs. This task assumes 5 levels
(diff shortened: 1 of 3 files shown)


bcipy/tasks/rsvp/calibration/inter_sequence_feedback_calibration.py
```diff
@@ -81,37 +81,45 @@ def __init__(self, win, daq, parameters, file_save):
...
+        self.psd_export_band = (self.psd_lower_limit, self.psd_upper_limit)
 
         # length of time to use for PSD calculation
         self.trial_length = self.time_flash * self.stim_length
 
+        self.lvl_5_threshold = self.parameters['feedback_level_5_threshold']
+        self.lvl_4_threshold = self.parameters['feedback_level_4_threshold']
+        self.lvl_3_threshold = self.parameters['feedback_level_3_threshold']
+        self.lvl_2_threshold = self.parameters['feedback_level_2_threshold']
+
```
lawhead, line 122: I think it's probably clearer for the end user to have a separate parameter for each category, but if there would ever be a different number of feedback categories it might be more flexible to have a single parameter with a delimited list of values.

lawhead (approved):
Lots of nice cleanup. Thanks!

PR #61, opened 2019-06-28
#166855378 ; shuts down the viewer when data is no longer streaming

lawhead:


lawhead:
https://www.pivotaltracker.com/n/projects/2119090

orbitalhybridization:
I was able to replicate the bug with your instructions, thanks! After cloning the branch and trying to replicate the bug again with the new changes, the viewer window flashes for a second, closes, and the process terminates with the output:

Waiting for LSL EEG data stream...
Sample rate: 256.0
Fast forwarding:
Chomped 0 records.

Is this expected? And has anyone else gotten this behaviour? I might be doing something wrong, but am not sure...

- Ian

[Edit: formatting]

lawhead:
Ian, thanks for testing. Yes, that is the expected output.

tab-cmd (approved):

bcipy/gui/viewer/data_source/lsl_data_source.py
```diff
@@ -60,7 +59,7 @@ def next_n(self, n: int, fast_forward=False):
             print(f'Chomped {chomped_count} records.')
             samples = samples[len(tmp):] + tmp            
 
-        if len(samples) < n:
+        if len(samples) < n and not fast_forward:
```
tab-cmd, line 71: A comment here or an addition to the docstring for the method would be helpful.

bcipy/gui/viewer/data_viewer.py
```diff
@@ -318,6 +318,9 @@ def update_buffer(self, fast_forward=False):
                 self.buffer.append(row)
         except StopIteration:
             self.stop()
+            self.Close()
```
tab-cmd, line 322: Is this closing Wx.Frame? We could have a `close` or `exit` method that wraps this which might be more clear

tab-cmd (approved):

PR #62, opened 2019-07-06
Small changes to clean up after tests and make sure all tests pass

smektala:
-Closes buffer processes to prevent them from running forever in the background
-Closes all display windows to prevent them from running forever in the background
-Moves test_do_epoch into test_decide_without_commit so that it passes and doesn't have to make more decisions based on fake probability data than are necessary 
-Fixes test_decide_with_commit

### DIFF
--- a/bcipy/acquisition/tests/test_client.py
+++ b/bcipy/acquisition/tests/test_client.py
@@ -75,7 +75,10 @@ class TestDataAcquistionClient(unittest.TestCase):
         daq.start_acquisition()
         time.sleep(0.1)
         daq.stop_acquisition()
-
+        
+        #Make sure we are able to stop the buffer process
+        buf_temp = daq._buf
+        
         daq._buf = None
 
         #test get_data
@@ -87,9 +90,10 @@ class TestDataAcquistionClient(unittest.TestCase):
         self.assertEqual(data_length,0)
 
         #test offset
-        offset = daq.offset()
+        offset = daq.offset
         self.assertEqual(offset,None)
-
+        
+        daq._buf = buf_temp
         daq.cleanup()
 
     def test_get_data(self):
--- a/bcipy/display/tests/test_display.py
+++ b/bcipy/display/tests/test_display.py
@@ -39,3 +39,5 @@ class TestInitializeDisplayWindow(unittest.TestCase):
 
         self.assertIsInstance(display_window, type(self.window))
         self.assertEqual(display_window, self.window)
+        
+        display_window.close()
--- a/bcipy/display/tests/test_display.py
+++ b/bcipy/display/tests/test_display.py
@@ -28,16 +28,14 @@ class TestInitializeDisplayWindow(unittest.TestCase):
 
         self.parameters = load_json_parameters(parameters_used,
                                                value_cast=True)
+                                               
+        self.display_window = init_display_window(self.parameters)
 
     def tearDown(self):
+        self.display_window.close()
         unstub()
 
     def test_init_display_window_returns_psychopy_window(self):
         """Test display window is created."""
-
-        display_window = init_display_window(self.parameters)
-
-        self.assertIsInstance(display_window, type(self.window))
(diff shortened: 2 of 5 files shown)


bcipy/display/tests/test_display.py
```diff
@@ -39,3 +39,5 @@ def test_init_display_window_returns_psychopy_window(self):
 
         self.assertIsInstance(display_window, type(self.window))
         self.assertEqual(display_window, self.window)
+        
+        display_window.close()
```
tab-cmd: Add to the `tearDown` method

bcipy/acquisition/tests/test_client.py
```diff
@@ -87,9 +90,10 @@ def test_daq_with_no_buffer(self):
         self.assertEqual(data_length,0)
 
         #test offset
-        offset = daq.offset()
+        offset = daq.offset
```
tab-cmd, line 95: :+1:

bcipy/acquisition/tests/test_client.py
```diff
@@ -75,7 +75,10 @@ def test_daq_with_no_buffer(self):
         daq.start_acquisition()
         time.sleep(0.1)
         daq.stop_acquisition()
-
+        
+        #Make sure we are able to stop the buffer process
+        buf_temp = daq._buf
```
tab-cmd, line 82: I'm confused why this is needed to stop the buffer process? @lawhead @smektala

bcipy/feedback/tests/test_visual_feedback.py
```diff
@@ -71,6 +71,7 @@ def setUp(self):
 
     def tearDown(self):
         # clean up by removing the data folder we used for testing
+        self.display.close()
```
tab-cmd, line 74: :+1:

bcipy/tasks/tests/test_main_frame.py
```diff
@@ -30,10 +30,15 @@ def test_decide_without_commit(self):
         #print(self.decision_maker.list_epoch[-1]['list_distribution'][-1])
         #print(probability_distribution)
         self.assertFalse(decision)
+        """Test do_epoch method"""
```
tab-cmd, line 40: I would leave this as a separate test. A good rule of thumb is one method or one assertion per unittest. Let's remove the print statement as well (I know you didn't add it!)

tab-cmd (changes requested):

tab-cmd:
Nice work! I left a few comments. Most are suggestions, but please clarify the acquisition buffer question. I pinged Matt as well.

bcipy/acquisition/tests/test_client.py
smektala: When the buffer server is started, daq._buf is set to a reference to the queue that is used to communicate with the server. This takes place when the daq object starts data acquisition. However, for the test to work correctly, daq._buf has to be set to None. When daq.cleanup() is run, it tries to send a stop message to the server process, but if the variable that should contain the message queue reference is still set to None, it isn't able to stop the buffer server process, and so it runs forever in the background, causing Pytest to hang when all tests are finished. The change saves the queue reference in buf_temp, so that daq._buf can be set back to the queue reference once the test is finished, and the buffer server process can be stopped.

bcipy/tasks/tests/test_main_frame.py
tab-cmd: This should still be removed. Can you remove the commented out print statements above as well?

bcipy/tasks/tests/test_main_frame.py
```diff
@@ -77,12 +81,13 @@ def test_form_display_state(self):
         self.decision_maker.form_display_state(self.decision_maker.state)
         self.assertEqual(self.decision_maker.displayed_state,'ABE')
         self.decision_maker.reset()
-
+
```
tab-cmd, line 89: extra space

tab-cmd (approved):

PR #64, opened 2019-07-12
#166772877 ; get_data queries

lawhead:
get_data queries in Windows take a long time to execute due to the multiprocessing Manager() call. This patch changes the method of inter-process communication in the buffer_server to make these queries more efficient.


tab-cmd (approved):
Looks good. Reviewing this file and the client, I think we could remove the reference / args to window. It doesn't look like we do anything with it?

lawhead:
Oh yes, good point about the win parameter. I think that was an earlier work-around that's no longer needed. I'll do some additional testing in the lab today and remove it if it's not needed.

PR #67, opened 2019-07-26
Improve parameter saving/loading

smektala:
# Improve parameter saving/loading

This PR changes the way the parameter editing form is launched, allowing parameters to be loaded directly from a file that is not the default parameters.json, without overwriting default parameters.

# Added
- Ability to load parameters from a file that is not the default parameters.json, without overwriting the default parameters.json
- Feature to check loaded parameters file against default parameters.json, and fill any missing parameters with the default values
- Parameter sorting alphabetically by group, displaying group names
- BciPy version number to parameter editing form
- Button to parameter editing form to reload default parameters

# Updated
- Launching parameter editing form (now creates another wxPython frame, rather than launching a subprocess)
- Method of saving current location of parameters file (now passed from parameter editing form to main GUI)
- Locations of parameter 'save' button and boolean parameters


bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -15,7 +18,12 @@
 class RSVPKeyboard(BCIGui):
     """GUI for launching the RSVP tasks."""
     event_started = False
-    PARAMETER_LOCATION = 'bcipy/parameters/parameters.json'
+
+    def __init__(self, title, size, background_color, default_param_location='bcipy/parameters/parameters.json',
+                 params_queue=None):
+        super(RSVPKeyboard, self).__init__(title, size, background_color)
+        self.PARAMETER_LOCATION = default_param_location
```
orbitalhybridization: I like this design, it's a lot more flexible yet leaves the original functionality as a default!

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -15,7 +18,12 @@
...
+        self.params_queue = params_queue
```
orbitalhybridization: Just a clarification question, does params_queue meant for multiple .json params file locations to be stored and accessed?

bcipy/gui/params_form.py
```diff
@@ -348,28 +370,99 @@ def onLoad(self, event: wx.EVT_BUTTON) -> None:
...
+
+            self.params, missing_keys = check_if_parameters_contains_all_keys(self.params, load_file)
+
+            if missing_keys:
+                for each_key in missing_keys:
+                    self.params[each_key] = Parameter(*self.params[each_key].values())
+
+                dialog = wx.MessageDialog(
+                    self, "Parameters file {} is missing keys {}. The default "
+                    "values for these keys will be loaded.".format(
```
orbitalhybridization: Again, super flexible functionality. I like that your design is controls for these custom parameters files that are minimalist and only have included changes that may be specific to a user's needs.

orbitalhybridization (commented):
I like the new additions and updated functionality! I only have a small question about params_queue, but the code looks solid overall. Nice having a full test folder too!

bcipy/gui/mode/RSVPKeyboard.py
smektala: When a parameter JSON file is loaded from a non-default location, params_queue is used by the parameter editing form to send the new parameter location back to the main RSVPKeyboard GUI.

.gitignore
```diff
@@ -22,3 +22,4 @@ htmlcov/
 bcipy.egg-info/
 build/
 dist/
+parameters_location.txt
```
tab-cmd: remove

bcipy/gui/BCInterface.py
```diff
@@ -19,6 +20,11 @@ def launch_mode(self, event) -> None:
                 shell=True)
 
 
+try:
```
tab-cmd: remove

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -97,7 +113,7 @@ def load_items_from_txt(self, _event):
         self.comboboxes[0].AppendItems(saved_users)
 
 
-def run_rsvp_gui():
+def run_rsvp_gui(default_param_location='bcipy/parameters/parameters.json', params_queue=None):
```
tab-cmd: We should store this default in one place and import it elsewhere.

bcipy/gui/params_form.py
```diff
@@ -338,6 +351,15 @@ def __init__(self, parent, title='BCI Parameters',
         self.SetSizer(vbox)
         self.SetupScrolling()
 
+    def get_parameters_from_file(self, file_name: str):
```
tab-cmd: We have a load json parameters function in `helpers.load`. Can we leverage or modify that for this use case?

bcipy/gui/params_form.py
```diff
@@ -348,28 +370,99 @@ def onLoad(self, event: wx.EVT_BUTTON) -> None:
...
+            self.get_parameters_from_file(load_file)
+
+            self.params, missing_keys = check_if_parameters_contains_all_keys(self.params, load_file)
+
+            if missing_keys:
+                for each_key in missing_keys:
+                    self.params[each_key] = Parameter(*self.params[each_key].values())
+
+                dialog = wx.MessageDialog(
+                    self, "Parameters file {} is missing keys {}. The default "
```
tab-cmd: single quotes!

bcipy/parameters/parameters.json
```diff
@@ -876,5 +876,13 @@
...
     "recommended_values": "",
     "type": "str"
+  }, 
+  "section_names": {
+    "value": "{'aaaa_booleans': 'Important BciPy Parameters', 'acq_config': 'ACQ Configuration', 'bci_config': 'BCI Configuration', 'feedback_config': 'Feedback Configuration', 'lang_model_config': 'Language Model Configuration', 'signal_config': 'Signal Configuration'}",
+    "section": "bci_config",
+    "readableName": "Names of parameter sections",
+    "helpTip": "Names of parameter sections",
+    "recommended_values": "",
+    "type": "str"
```
tab-cmd: This is not a string, let's call it an object and modify the cast values function to support his type

bcipy/parameters/parameters.json
```diff
@@ -863,7 +863,7 @@
   },
   "backspace_always_shown": {
     "value": "false",
-    "section": "bci_config",
+    "section": "aaaa_booleans",
```
tab-cmd: Let's name this something else... like, `top_level_config`? something a little more descriptive.

tab-cmd (changes requested):

bcipy/parameters/parameters.json
```diff
@@ -876,5 +876,13 @@
...
     "recommended_values": "",
     "type": "str"
+  }, 
+  "section_names": {
+    "value": "{'top_level_config': 'Important BciPy Parameters', 'acq_config': 'ACQ Configuration', 'bci_config': 'BCI Configuration', 'feedback_config': 'Feedback Configuration', 'lang_model_config': 'Language Model Configuration', 'signal_config': 'Signal Configuration'}",
+    "section": "bci_config",
+    "readableName": "Names of parameter sections",
+    "helpTip": "Names of parameter sections",
+    "recommended_values": "",
+    "type": "object"
```
lawhead: The 'section_names' here is a developer-focused configuration rather than a user-focused parameter. My feeling is that this should go elsewhere, either in a config file or somewhere in the GUI code. 

As for the section names themselves: ACQ is not an abbreviation; it should just be spelled out as Acquisition. All of this is configuration so having 'Configuration' in the section is redundant. The BCI section is really large and we should brainstorm ways of breaking it down; also, maybe this should be named General.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -15,7 +16,10 @@
 class RSVPKeyboard(BCIGui):
     """GUI for launching the RSVP tasks."""
     event_started = False
-    PARAMETER_LOCATION = 'bcipy/parameters/parameters.json'
+
+    def __init__(self, title, size, background_color, default_param_location='bcipy/parameters/parameters.json'):
+        super(RSVPKeyboard, self).__init__(title, size, background_color)
+        self.PARAMETER_LOCATION = default_param_location
```
lawhead: The parameter default_param_location should just be renamed to param_location. Also, there are several places in the code where this location is used as a default value. This makes it error-prone if we ever need to change this location. We should just have a single place where the default value is specified and pass in the parameter for every other method. 

Also, self.PARAMETER_LOCATION shouldn't be capitalized since it's not a constant.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -45,8 +50,8 @@ def launch_bci_main(self, event: wx.Event) -> None:
             username = self.comboboxes[0].GetValue().replace(" ", "_")
             experiment_type = event.GetEventObject().GetId()
             mode = 'RSVP'
-            cmd = 'python bci_main.py -m {} -t {} -u {}'.format(
-                mode, experiment_type, username)
+            cmd = 'python bci_main.py -m {} -t {} -u {} -p {}'.format(
+                mode, experiment_type, username, self.PARAMETER_LOCATION)
```
lawhead: Now that we're using python 3 we can use f-strings.

    cmd = f'python bci_main.py -m {mode} -t {experiment_type} -u {username} -p {self.parameter_location}'

bcipy/gui/params_form.py
```diff
@@ -67,22 +65,37 @@ def createControls(self):
...
-            if param.type == 'bool':
-                form_input = self.bool_input(param)
-            elif 'path' in param.type:
-                form_input = self.file_input(param)
-            elif isinstance(param.recommended_values, list):
-                form_input = self.selection_input(param)
-            # TODO: NumCtrl for numeric input types.
-            # from wx.lib.masked import NumCtrl
-            # elif param['type'] in ['float', 'int']:
+        # Retrieve and alphebatize list of unique parameter sections
```
lawhead: Typo in documentation: 'alphabetize'

bcipy/gui/params_form.py
```diff
@@ -67,22 +65,37 @@ def createControls(self):
...
+        param_sections = sorted(list(set([param[1].section for param in self.params.items()])))
+        # Sort parameters by section
+        key_dict = []
+        for each_section in param_sections:
+            # Get readable names for sections
+            name_dict = ast.literal_eval(self.params['section_names'].value)
```
lawhead: This line throws a KeyError when section_names is missing. I would prefer to have this configuration elsewhere and not require an eval.

bcipy/gui/params_form.py
```diff
@@ -67,22 +65,37 @@ def createControls(self):
...
+            new_key = name_dict[each_section] if each_section in name_dict else each_section
+            item_to_insert = [new_key, sorted(
+                [key for key, param in self.params.items() if param.section == each_section])]
+            if each_section == 'top_level_config':
+                # Move most important variables to top
+                key_dict.insert(0, item_to_insert)
```
lawhead: This whole section should be factored out into another method so we can test and confirm it's behavior. key_dict is an array, not a dict. In Python 3.6 and later the dict type maintains insertion order, however, a better data structure would be the collections.OrderedDict.

bcipy/gui/params_form.py
```diff
@@ -67,22 +65,37 @@ def createControls(self):
...
             else:
-                form_input = self.text_input(param)
+                key_dict.append(item_to_insert)
 
-            self.add_input(self.controls, key, form_input)
+        for each_key, param_list in key_dict:
```
lawhead: When iterating, just use `key` rather than `each_key`.

bcipy/gui/params_form.py
```diff
@@ -67,22 +65,37 @@ def createControls(self):
...
+        param_sections = sorted(list(set([param[1].section for param in self.params.items()])))
+        # Sort parameters by section
+        key_dict = []
+        for each_section in param_sections:
+            # Get readable names for sections
+            name_dict = ast.literal_eval(self.params['section_names'].value)
+            new_key = name_dict[each_section] if each_section in name_dict else each_section
+            item_to_insert = [new_key, sorted(
+                [key for key, param in self.params.items() if param.section == each_section])]
+            if each_section == 'top_level_config':
```
lawhead: 'top_level_config' should probably be a constant.

bcipy/gui/params_form.py
```diff
@@ -348,28 +373,91 @@ def onLoad(self, event: wx.EVT_BUTTON) -> None:
...
             load_file = fd.GetPath()
-            self.loaded_from.SetLabel(f'Loaded from: {load_file}')
-            self.flash_msg.SetLabel('Click the Save button to persist these '
-                                    'changes.')
-            self.vbox.Hide(self.form)
+
+            self.get_parameters_from_file(load_file)
+
+            self.params, missing_keys = check_if_parameters_contains_all_keys(self.params, load_file)
+
```
lawhead, line 388: The `check_if_parameters_contains_all_keys` name is a bit misleading, since this not only checks values but mutates the parameters dict. I think a better approach here would be to separate out this functionality into a few simpler functions that can be independently tested:

1. Get the missing key/values (diff) as a new dict:
https://stackoverflow.com/questions/32815640/how-to-get-the-difference-between-two-dictionaries-in-python
2. Merge the new diff dict into the params
3. Convert all keys to Parameter objects.

lawhead:
As I was testing this locally I found several aspects of the user interaction to be confusing. 

The Save button acts as a Save As; when I save to a location, close the window,  and click Edit Parameters again it doesn't load my saved changes. As a user I'm not sure that the file I just saved is going to be used in the experiment.

Loading a parameters file other than the default changes the ordering of the items. We should use consistent ordering here so it's clear what changed.

Also, there is a bug if you try to Load an existing parameters file with a missing key (section_names). The new key does not get written to the file. When I close the dialog and then try to Edit Parameters again the code throws a KeyError for the missing 'section_names' key.

bcipy/gui/params_form.py
smektala: Fixed! key_dict used to be a dictionary, and I forgot to change its name when I made it an array. It is now an OrderedDict.

tab-cmd (approved):

PR #69, opened 2019-08-07
Raw data file

lawhead:
Refactored to create the raw_data.csv file from the sqlite database at the end of acquisition, rather than during acquisition.


lawhead:
In a subsequent effort I'm going to explore whether or not we can now remove a process from the acquisition client.

tab-cmd:
we should make sure that raw_data is written at the end of existing tasks to preserve functionality.

lawhead:
Yes, functionality should be the same. In the client stop_acquisition method it writes the raw_data.csv file.

tab-cmd:
> Yes, functionality should be the same. In the client stop_acquisition method it writes the raw_data.csv file.

Gotcha! I see that now - thanks for doing this!

tab-cmd (approved):

smektala (approved):

PR #75, opened 2019-08-12
#167850628 Re-enable console output for offline_analysis

lawhead:


bcipy/signal/model/offline_analysis.py
```diff
@@ -111,14 +111,21 @@ def offline_analysis(data_folder: str = None,
 
 if __name__ == "__main__":
     import argparse
+    import sys
```
tab-cmd: Can this be removed?

tab-cmd (approved):
Looks good thanks for setting up logging correctly!

bcipy/signal/model/offline_analysis.py
lawhead: Thanks, good catch.

PR #87, opened 2019-10-08
#168649473 ; Copy Phrase error

lawhead:
https://www.pivotaltracker.com/story/show/168649473


azizkocana (approved):
Good catch!
If we check the distribution history, number of elements is actually always 1 more than actual passed sequences. This is because language model prior information is also stored within that list. We hard-coded to have at least 1 sequence and I assume this caused termination by checking the length of past distributions which was \geq 1 all the time due to prior information.

PR #94, opened 2020-10-10
Upgrade and simplify dependencies

tab-cmd:
# Overview

Updated and cleaned requirements.txt.

## Ticket

https://www.pivotaltracker.com/story/show/174911840

## Contributions

- Upgrade PsychoPy==2020.2.4.post1
- Remove dependencies specific to PsychoPy
- Upgrade numpy, pandas, and scipy
- Update system_utils to get the screen resolution using the updated pyglet (PsychoPy dependency)
- Update load csv method to use `.values`. See https://pandas.pydata.org/pandas-docs/version/0.25.1/reference/api/pandas.DataFrame.as_matrix.html

## Test

- Unit tests
- Calibration
- Re-ran offline analysis on older recordings.. Same AUC and artifacts produced. 

## Documentaion


requirements.txt
```diff
@@ -1,28 +1,12 @@
...
 matplotlib==2.1.1
 pylsl==1.13.1
-psutil==5.4.0
-pandas==0.24.1
-sounddevice==0.3.10
-SoundFile==0.10.1
-PySoundCard==0.5.2
-PySoundFile==0.9.0
-Pillow==4.3.0
-openpyxl==2.6.3
```
lawhead, line 28: We use some of these dependencies directly (ex. sounddevice and psutil). I'm not sure what best practices are but if I'm reading the docs correctly it seems like we should retain them either here or in install_requires (setup.py) even though they are included transitively from psychopy.

lawhead (approved):

requirements.txt
tab-cmd: Good callout. I added back all the requirements I could find that we explicitly use. There are a few cases we might refactor to use PsychoPy methods now. When I first implemented sound feedback, PsychoPy sound wasn't working. It seems more stable now.

PR #98, opened 2020-10-13
System info

lawhead:
# Overview

Code spike to log more detailed system information.

## Ticket

https://www.pivotaltracker.com/story/show/175193913

## Contributions

- Added more information to the system_utils#get_system_info method, including platform details, operating system name, version, architecture, processor name, processor speed, CPU count, and the amount of RAM.
- Factored out screen resolution code to a separate method for efficiency; also fixed a deprecation warning.
- Added code to output the system information to the session logs.

## Test

- Ran the calibration task both with full_screen set to true and false to confirm that the letter sizes were still the same after refactoring. Inspected the session logs to confirm that the new system information was written correctly.


bcipy/helpers/system_utils.py
```diff
@@ -70,23 +72,45 @@ def bcipy_version() -> str:
...
+    """Gets the screen resolution.
+
+    Note: Use this method if only the screen resolution is needed; it is much more efficient
+    than extracting that information from the dict returned by the get_system_info method.
+
+    Returns
+    -------
+        (width, height)
+     """
+    screen = pyglet.canvas.get_display().get_default_screen()
```
tab-cmd, line 85: This will need to change slightly to work with the updated dependencies

bcipy/helpers/system_utils.py
```diff
@@ -70,23 +72,45 @@ def bcipy_version() -> str:
...
+    return (screen.width, screen.height)
 
-    # Three lines for getting screen resolution
-    platform = pyglet.window.get_platform()
-    display = platform.get_default_display()
-    screen = display.get_default_screen()
 
-    mem = psutil.virtual_memory()
+def get_system_info() -> dict:
```
tab-cmd, line 89: Awesome - this is exactly what Barry et al. asked for!

bcipy/helpers/system_utils.py
```diff
@@ -70,23 +72,45 @@ def bcipy_version() -> str:
...
-        'bcipy_version': bcipy_version()
+        'resolution': [screen_width, screen_height],
+        'memory': psutil.virtual_memory().available / 1024 / 1024,
+        'bcipy_version': bcipy_version(),
+        'platform': platform.system(),
+        'platform-release': platform.release(),
+        'platform-version': platform.version(),
+        'architecture': platform.machine(),
+        'processor': platform.processor(),
+        'cpu_count': os.cpu_count(),
```
tab-cmd, line 112: It looks like we can get the CPU name in most cases. It would be worth adding! https://github.com/workhorsy/py-cpuinfo#raw-fields

tab-cmd (approved):
Looks good. I added some suggestions

PR #100, opened 2020-10-19
System directory to support experiments

tab-cmd:
# Overview

Added a system directory that contains an experiment and field registry. Updated save method to include an experiment. 

## Ticket

https://www.pivotaltracker.com/story/show/175193520

## Contributions

- Added `.bcipy`
- Added `.bcipy/experiment/`
- Added `.bcipy/field/`
- Added `.bcipy/experiment/experiments.json`
- Added `.bcipy/field/fields.json`
- Updated `helpers/save/init_save_data_structure` to include the experiment in the save structure. Defined a DEFAULT_EXPERIMENT_ID.
- Removed copying of the helpers folder (not useful to clinical team)
- Updated `bci_main` to include the setting of the experiment id.
- Updated helper save tests

## Test

- Unit tests

## Documentaion

- No

### DIFF
--- a/bci_main.py
+++ b/bci_main.py
@@ -4,12 +4,12 @@ from bcipy.helpers.task import print_message
 from bcipy.helpers.system_utils import get_system_info, configure_logger
 from bcipy.helpers.language_model import init_language_model
 from bcipy.helpers.load import load_signal_model
-from bcipy.helpers.save import init_save_data_structure
+from bcipy.helpers.save import init_save_data_structure, DEFAULT_EXPERIMENT_ID
 from bcipy.tasks.start_task import start_task
 from bcipy.tasks.task_registry import ExperimentType
 
 
-def bci_main(parameters: dict, user: str, exp_type: int, mode: str) -> bool:
+def bci_main(parameters: dict, user: str, exp_type: int, mode: str, experiment: str = DEFAULT_EXPERIMENT_ID) -> bool:
     """BCI Main.
 
     The BCI main function will initialize a save folder, construct needed information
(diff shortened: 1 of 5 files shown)


bcipy/helpers/save.py
```diff
@@ -45,28 +49,13 @@ def init_save_data_structure(data_save_path,
...
-        for file_name in src_files:
-
-            # get the full name
-            full_file_name = os.path.join('bcipy/helpers/', file_name)
-
-            # Check that constructed file is a real file and ends in .py
-            if (os.path.isfile(full_file_name)) and '.py' in file_name:
-                # Copy over!
-                copy2(full_file_name, helper_folder_name)
-
```
lawhead, line 65: When I was looking into BIDS I meant to ask if these helpers are still needed. I'm glad we're able to clean up the data.

lawhead (approved):

bci_main.py
```diff
@@ -36,7 +38,12 @@ def bci_main(parameters: dict, user: str, exp_type: int, mode: str) -> bool:
 
     # Initialize Save Folder
     save_folder = init_save_data_structure(
-        data_save_location, user, parameter_location, mode, exp_type)
+        data_save_location,
+        user,
+        parameter_location,
+        mode=mode,
+        experiment_type=exp_type,
+        experiment=experiment)
```
lawhead, line 47: bci_main is passing an extra parameter to init_save_data_structure which causes an error (`experiment`). Sorry I didn't catch this during the review.

lawhead: Or actually, it looks like maybe this needs to be `experiment_id`, rather than `experiment`?

tab-cmd: yeah - it should be experiment_id. I'll re-open a PR

PR #102, opened 2020-10-30
Gui refactor

tab-cmd:
# Overview

Refactored base GUI to use PyQT5 instead of WxPython. This required `gui_main`, `BCInterface` and `RSVPKeyboard` to be redone.

## Ticket

https://www.pivotaltracker.com/story/show/175194654

## Contributions

- Refactored `gui_main`. I kept many of the same methods and updated them to use PyQt5 functionality. I added more default methods (ex. `build_assests`) and abstractions (`AlertResponse`, `AlertMessageType`) to help isolate the downstream GUIS from the knowledge of PyQT5. 
- Refactored `RSVPKeyboard` and `BCInterface` to use new base class and methods (`build_*`)
- Ignore non-default parameters. 

## Test

- Run `python bcipy/gui/gui_main.py`
- Run `python bcipy/gui/BCInterface.py`. Select RSVP.
- Run `python bcipy/gui/mode/RSVPKeyboard.py`. Input a user_id, run a task. Edit and load parameters.

## Documentaion

### DIFF
--- a/bcipy/gui/BCInterface.py
+++ b/bcipy/gui/BCInterface.py
@@ -37,6 +37,11 @@ class BCInterface(BCIGui):
     def build_text(self):
         self.add_static_text(text='Brain Computer Interface', position=[250, 0], size=[250, 100], background_color='black', text_color='white')
 
+    def build_assets(self):
+        self.build_buttons()
+        self.build_images()
+        self.build_text()
+
 
 def start_app():
     """Start BCIGui."""
@@ -46,11 +51,6 @@ def start_app():
         height=500,
         width=700,
         background_color='black')
-
-    ex.build_buttons()
-    ex.build_images()
-    ex.build_text()
-
 
     ex.show_gui()
 
--- a/bcipy/gui/BCInterface.py
+++ b/bcipy/gui/BCInterface.py
@@ -35,7 +35,7 @@ class BCInterface(BCIGui):
             path='bcipy/static/images/gui_images/neu.png', position=[550, 0], size=200)
 
     def build_text(self):
-        self.add_static_text(text='Brain Computer Interface', position=[250, 0], size=[250, 100], background_color='black', text_color='white')
+        self.add_static_textbox(text='Brain Computer Interface', position=[250, 0], size=[250, 100], background_color='black', text_color='white')
 
     def build_assets(self):
         self.build_buttons()
--- a/bcipy/gui/BCInterface.py
+++ b/bcipy/gui/BCInterface.py
@@ -1,7 +1,6 @@
 import subprocess
-import wx
 import sys
-from bcipy.gui.gui_main import BCIGui, app, PushButton
+from bcipy.gui.gui_main import BCIGui, app
 
 
 class BCInterface(BCIGui):
@@ -15,7 +14,7 @@ class BCInterface(BCIGui):
 
     def build_buttons(self) -> None:
         """Build buttons.
-        
+
         Construct all buttons needed for BCInterface.
         """
         self.add_button(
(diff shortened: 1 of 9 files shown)


bcipy/gui/BCInterface.py
```diff
@@ -1,57 +1,79 @@
...
-    path='bcipy/static/images/gui_images/neu.png', position=(550, 0), size=125)
-
-
-# Make the GUI Show now
-gui.show_gui()
-app.MainLoop()
+    """BCInterface. Main GUI to select different modes."""
+
+    def launch_rsvp(self) -> None:
+        print('launching')
```
tab-cmd: Nts: remove print (use logger)

bcipy/gui/gui_main.py
```diff
@@ -1,106 +1,215 @@
...
+    QFileDialog,
+    QWidget,
+    QPushButton,
+    QLabel,
+    QLineEdit,
+    QComboBox,
+    QGridLayout,
+    QMessageBox)
+from PyQt5.QtGui import QPixmap, QFont
+from PyQt5.QtCore import pyqtSlot
```
lawhead, line 19: Are we pulling in pyqt transitively? We may want to add it to our requirements.txt file.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -1,210 +1,281 @@
...
+
+        If a user id has been selected, get the id set on the task button and execute
+            the task in a new shell.
+        """
         if self.check_input():
             self.event_started = True
-            username = self.comboboxes[0].GetValue().replace(" ", "_")
-            experiment_type = event.GetEventObject().GetId()
+            username = self.user_input.currentText()
+            experiment_type = self.sender().get_id()
```
lawhead, line 80: It took a bit of digging to figure out how `self.user_input` is initialized. For clarity, we might want to declare this property in the `__init__` and add a comment that it gets set in the `show_gui` method.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -1,210 +1,281 @@
...
-        data_save_loc = parameters['data_save_loc']
+    def load_items_from_txt(self) -> List[str]:
+        """Load Items From Text.
+
+        Loads user directory names below experiments from the data path defined in parameters.json
+        and returns them as a list.
+        """
+        saved_users = []
+        data_save_loc = self.parameters['data_save_loc']
+
```
lawhead, line 132: I just realized that the parameter_location gets updated, but the parameters never get reloaded, so this will always be the data_save_loc from the default parameters. I can fix that after this is merged.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -1,210 +1,281 @@
...
-                self.parameter_location = fd.GetPath()
-
-    def edit_parameters(self, _event) -> None:
-        """Edit Parameters. Prompts for a parameters.json file to use. If the default parameters
-        are selected a copy is used.
+        self.parameters = load_json_parameters(
+            self.parameter_location, value_cast=True)
+
+        self.task_colors = itertools.cycle(
+            ['blue', 'green', 'yellow', 'red', 'limegreen', 'gray', 'pink'])
```
lawhead: These colors seem more intense in pyqt than wx. We may need a better palette.

lawhead (approved):

bcipy/gui/mode/RSVPKeyboard.py
tab-cmd: I quickly added some to test the refactor. I'll look for something a little more appealing!

bcipy/gui/mode/RSVPKeyboard.py
tab-cmd: I added a bug ticket for this! https://www.pivotaltracker.com/story/show/175541919

PR #103, opened 2020-11-02
Update params fix

lawhead:
# Overview

Modified the RSVPKeyboard GUI so that loading a new parameters file now updates GUI elements that are populated from configured values.

## Ticket

https://www.pivotaltracker.com/story/show/175541919

## Contributions

- Added more logic to the code that refreshes parameters.

## Test

- Load a new parameters file with the 'data_save_loc' parameter set to a different directory than the default value. The combo list for selecting a user id should be updated with a new set of values depending on the options available in the new directory.

### DIFF
--- a/bcipy/gui/mode/RSVPKeyboard.py
+++ b/bcipy/gui/mode/RSVPKeyboard.py
@@ -36,7 +36,7 @@ class RSVPKeyboard(BCIGui):
         # This is set in the build_inputs method that is called automatically when using show_gui
         self.user_input = None
 
-    def set_parameter_location(self, path: str):
+    def set_parameter_location(self, path: str) -> None:
         """Sets the parameter_location to the given path. Reloads the parameters
         and updates any GUI widgets that are populated based on these params."""
         self.parameter_location = path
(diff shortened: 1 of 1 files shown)


bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -206,24 +218,30 @@ def build_buttons(self) -> None:
             size=[btn_refresh_width, command_btn_height], background_color='white',
             action=self.refresh)
 
+    def update_user_list(self) -> None:
+        """Updates the user_input combo box with a list of user ids based on the
+        data directory configured in the current parameters."""
+
+        self.user_input.clear()
+        self.user_input.addItem(RSVPKeyboard.default_text)
+        for user_id in self.load_items_from_txt():
```
tab-cmd: We could use addItems here. The output of self.load_items_from_txt() should be guaranteed a `List[str]`.

bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -28,14 +28,24 @@ def __init__(self, *args, **kwargs):
...
-            self.parameter_location, value_cast=True)
+        self.parameters = load_json_parameters(self.parameter_location,
+                                               value_cast=True)
         self.task_colors = itertools.cycle(
             ['Tomato', 'orange', 'MediumTurquoise', 'MediumSeaGreen', 'MediumPurple', 'Moccasin'])
 
         # This is set in the build_inputs method that is called automatically when using show_gui
         self.user_input = None
 
+    def set_parameter_location(self, path: str):
```
tab-cmd: `-> None`

tab-cmd (approved):

PR #104, opened 2020-11-06
Convert data

lawhead:
# Overview

Added a module for data conversions that will be useful for data sharing. Implemented a conversion function to the EDF format.

## Ticket

https://www.pivotaltracker.com/story/show/175193894

## Contributions

- Added the convert.py helper module to convert bcipy raw data to EDF format.
- Added new constructor to parameters.py to make it easier to mock parameters in unit tests.
- Added functionality to the triggers.py helper for reading in a triggers.txt file.
- Demo script
- Unit tests

## Test

- Run the unit tests.
- Run the demo script against a data folder.
- In a python REPL, use the demo function plot_edf to plot the result.


bcipy/helpers/convert.py
```diff
@@ -0,0 +1,137 @@
...
+                   edf_path: str = None,
+                   overwrite=False,
+                   use_event_durations=False) -> Path:
+    """ Converts BciPy raw_data to the EDF+ filetype using pyEDFlib.
+
+    Parameters
+    ----------
+    raw_data_dir - directory which contains the data to be converted. This
+        location must also contain a parameters.json configuration file.
+    output_path - optional path to write converted data; defaults to writing
```
tab-cmd: These need to be updated

bcipy/helpers/convert.py
```diff
@@ -0,0 +1,137 @@
...
+        edf_path = Path(data_dir, 'raw.edf')
+
+    params = load_json_parameters(Path(data_dir, 'parameters.json'),
+                                  value_cast=True)
+    raw_data, _, ch_names, _, sfreq = read_data_csv(
+        Path(data_dir, params['raw_data_name']))
+    durations = trigger_durations(params) if use_event_durations else {}
+
+    with open(Path(data_dir, params['trigger_file_name']), 'r') as trg_file:
+        triggers = read_triggers(trg_file)
```
tab-cmd, line 46: We don't correct the offset here, right? I may be following wrong. We should add an argument to handle any offset correction (if present)

bcipy/helpers/convert.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Adapted from: https://github.com/holgern/pyedflib
+
+    Parameters
+    ----------
+    raw_data_dir - directory which contains the data to be converted. This
+        location must also contain a parameters.json configuration file.
+    output_path - optional path to write converted data; defaults to writing
+        a file named raw.edf in the raw_data_dir.
+    overwrite - If True, the destination file (if it exists) will be overwritten.
```
tab-cmd, line 71: A couple  of parameters to add

bcipy/helpers/tests/test_convert.py
```diff
@@ -0,0 +1,113 @@
...
+from typing import List
+
+import numpy as np
+from mne.io import read_raw_edf
+
+from bcipy.helpers.convert import convert_to_edf
+from bcipy.helpers.parameters import Parameters
+
+
+def sample_data(rows: int = 1000, ch_names: List[str] = ['c1', 'c2',
```
tab-cmd: I wonder if there's a better place for this function to live - acquisition? Otherwise, let's pull in the generate random data function to populate the channel_data

bcipy/helpers/tests/test_convert.py
```diff
@@ -0,0 +1,113 @@
...
+                                                         'c3']) -> str:
+    """Creates sample data to be written as a raw_data.csv file
+    
+    rows - number of sample rows to generate
+    ch_names - channel names
+    """
+    # Mock the raw_data file
+    sep = '\r\n'
+    meta = sep.join([f'daq_type,LSL', 'sample_rate,256.0'])
```
tab-cmd, line 28: linting

bcipy/helpers/tests/test_triggers.py
```diff
@@ -34,14 +45,16 @@ def sample_raw_data(trigger_seq: List[Tuple[str, str]] = [],
...
+    meta = sep.join([f'daq_type,{daq_type}', 'sample_rate,{sample_rate}'])
+    header = 'timestamp,' + ','.join(ch_names) + ',TRG'
 
     data = []
     for i in range(1000):
         timestamp = i + 10.0
-        channel_data = [str(random.uniform(-1000, 1000)) for _ in range(3)]
-        trg = triggers_by_time.get(timestamp, NONE_VALUE)
+        channel_data = [
+            str(random.uniform(-1000, 1000)) for _ in range(len(ch_names))
```
tab-cmd: I want to standardize how we mock our data - I think we have a random data generator function we can use / extend for your cases.

requirements.txt
```diff
@@ -16,4 +16,5 @@ pandas==1.1.3
 psutil==5.7.2
 Pillow==8.0.0
```
tab-cmd, line 18: I think you mentioned needing to upgrade mne for this - what version are you on?

tab-cmd (approved):

requirements.txt
lawhead: I did some testing and the issue I was encountering wasn't related to mne. I left the version as-is for this review so we could do more stress-testing on it.

bcipy/helpers/tests/test_triggers.py
lawhead: Yes, good point. I'll update this to use the acquisition module.

bcipy/helpers/convert.py
lawhead: Thanks, good catch.

bcipy/helpers/convert.py
lawhead: `read_triggers` is a new helper function that does correct the offset for the returned values. I can add a parameter to be able to return the values uncorrected, but I'm not sure if this would be useful without also providing the offset. Are there any instances in our code when we want these values without the offset correction?

bcipy/helpers/tests/test_convert.py
lawhead: The generation of sample data should definitely come from the acquisition generators. I'll update the code here.

As far as reading and writing a raw_data file, I think we're missing an abstraction/module. This logic is repeated a number of places throughout the code. I will create a followup ticket to clean this up since it affects a number of places.

bcipy/helpers/convert.py
tab-cmd: We may want to provide it as an argument because there are both system and static offsets. It may be useful to have them provide it as argument. As long as it's handled though we can ask the group during the demo how useful that may or not be.

PR #105, opened 2020-11-11
Upgrade parameters

lawhead:
# Overview

Give the user the ability to upgrade a parameters file with new properties when the BciPy default parameters are updated.

## Ticket

https://www.pivotaltracker.com/story/show/175542898

## Contributions

- Functionality to update a Parameters object with new values.
- GUI prompt to update the selected parameters file if it is out of date.
- Added logic to bci_main to raise an exception if the current parameters file is outdated.

## Test

- Running the bci_main code using a configuration file that is missing a property should result in an error.
- Add a property manually to the default parameters.json file. In the GUI, select an existing parameters file that does not include the property. Confirm the prompt to upgrade the configuration. The new property should exist in the file.


bcipy/gui/mode/RSVPKeyboard.py
```diff
@@ -52,15 +52,29 @@ def select_parameters(self) -> None:
         Opens a dialog to select the parameters.json configuration to use.
         """
 
-        response = self.get_filename_dialog(message='Select parameters file', file_type='JSON files (*.json)')
+        response = self.get_filename_dialog(message='Select parameters file',
+                                            file_type='JSON files (*.json)')
         if response:
             self.set_parameter_location(response)
+            # If outdated, prompt to merge with the current defaults
+            default_parameters = load_json_parameters(DEFAULT_PARAMETERS_PATH,
```
tab-cmd, line 60: Unrelated to the PR... I was thinking of making value cast true by default. It seems that is the most common case, and I don't even think we call it anywhere without casting. What do you think?

tab-cmd (approved):

PR #108, opened 2020-12-01
1.4.3 querying stopping

azizkocana:
# Overview

Includes stimulus selection and stopping decision modularization that is directly used by the decision maker during an actual typing task.

## Contributions

- Stimulus Selection:  Given the task history, stimulus selection module decides on which letters to be presented via Display. Observe that the implementation of stimuli selection disentangles letter selection and stimuli preparation (e.g. fixation cross, timing and colors for RSVP). Therefore, presented modules can be used in any other subset selection paradigm.

- Stopping Criteria: In a typing task, deciding when to stop based on confidence is upmost important. Each module, once defined in the Decision Maker, checks the history of the typing task for a specific epoch and terminates based on different rules (e.g. maximum number of iterations reached, confidence level reached, uncertainty level is achieved etc.) 

- Synthesized User: The base user simulator module which is designed to provide generated user feedback in a mock typing task.

- Copy Phrase Simulation: Mocks a typing task using synthesized user. (Serves as a smoke test for this pull request but the benefits are beyond imagination!)

## Notes

- Decision maker now expects inputs for both query and stopping modules and hence this information should be passed from parameters file.

- It is preferred to test the system with an actual user to make sure it is also working with equipment plugged in.

### DIFF
--- a/bcipy/helpers/copy_phrase_wrapper.py
+++ b/bcipy/helpers/copy_phrase_wrapper.py
@@ -8,6 +8,9 @@ from bcipy.helpers.task import BACKSPACE_CHAR, trial_reshaper
 from bcipy.signal.model.inference import inference
 from bcipy.signal.process.filter import bandpass, notch, downsample
 from bcipy.tasks.rsvp.main_frame import EvidenceFusion, DecisionMaker
+from bcipy.tasks.rsvp.query_mechanisms import NBestStimuliAgent
+from bcipy.tasks.rsvp.stopping_criteria import CriteriaEvaluator, \
+    MinIterationsCriteria, MaxIterationsCriteria, ProbThresholdCriteria
 from bcipy.helpers.language_model import norm_domain, sym_appended, \
     equally_probable, histogram
 
@@ -57,15 +60,26 @@ class CopyPhraseWrapper:
         seq_constants = []
         if backspace_always_shown and BACKSPACE_CHAR in alp:
             seq_constants.append(BACKSPACE_CHAR)
+
+        # Stimuli Selection Module
+        stopping_criteria = CriteriaEvaluator(
+            continue_criteria=[MinIterationsCriteria(min_num_seq)],
+            commit_criteria=[MaxIterationsCriteria(max_num_seq),
+                             ProbThresholdCriteria(decision_threshold)])
+
+        # TODO: Change this hard number 8!
+        stimuli_agent = NBestStimuliAgent(alphabet=alp,
+                                          len_query=8)
+
         self.decision_maker = DecisionMaker(
-            min_num_seq,
-            max_num_seq,
-            decision_threshold=decision_threshold,
+            stimuli_agent=stimuli_agent,
+            stopping_evaluator=stopping_criteria,
             state=task_list[0][1],
             alphabet=alp,
             is_txt_stim=is_txt_stim,
             stimuli_timing=stimuli_timing,
             seq_constants=seq_constants)
+
         self.alp = alp
         # non-letter target labels include the fixation cross and calibration.
         self.nonletters = ['+', 'PLUS', 'calibration_trigger']
@@ -100,7 +114,8 @@ class CopyPhraseWrapper:
 
         # Remove 60hz noise with a notch filter
         notch_filter_data = notch.notch_filter(
-            raw_data, self.sampling_rate, frequency_to_remove=self.notch_filter_frequency)
+            raw_data, self.sampling_rate,
+            frequency_to_remove=self.notch_filter_frequency)
 
         # bandpass filter from 2-45hz
         filtered_data = bandpass.butter_bandpass_filter(
@@ -113,7 +128,8 @@ class CopyPhraseWrapper:
         # downsample
         data = downsample.downsample(
             filtered_data, factor=self.downsample_rate)
-        x, _, _, _ = trial_reshaper(target_info, times, data, fs=self.sampling_rate,
+        x, _, _, _ = trial_reshaper(target_info, times, data,
+                                    fs=self.sampling_rate,
                                     k=self.downsample_rate, mode=self.mode,
                                     channel_map=self.channel_map,
                                     trial_length=window_length)
@@ -201,7 +217,8 @@ class CopyPhraseWrapper:
                          if alp_letter == prior_sym]
 
                 # display histogram of LM probabilities
-                print(f"Printed letters: '{self.decision_maker.displayed_state}'")
+                print(
+                    f"Printed letters: '{self.decision_maker.displayed_state}'")
                 print(histogram(lm_letter_prior))
 
             # Try fusing the lmodel evidence
--- a/bcipy/helpers/copy_phrase_wrapper.py
+++ b/bcipy/helpers/copy_phrase_wrapper.py
@@ -67,9 +67,9 @@ class CopyPhraseWrapper:
             commit_criteria=[MaxIterationsCriteria(max_num_seq),
                              ProbThresholdCriteria(decision_threshold)])
 
-        # TODO: Change this hard number 8!
+        # TODO: Parametrize len_query in the future releases!
         stimuli_agent = NBestStimuliAgent(alphabet=alp,
-                                          len_query=8)
+                                          len_query=10)
 
         self.decision_maker = DecisionMaker(
             stimuli_agent=stimuli_agent,
--- a/bcipy/tasks/rsvp/main_frame.py
+++ b/bcipy/tasks/rsvp/main_frame.py
@@ -92,9 +92,10 @@ class DecisionMaker:
                         arrays with prob. dist. over alp
             seq_constants(list[str]): list of letters which should appear in
                 every sequence.
-            criteria_evaluator: CriteriaEvaluator - optional parameter to
+            stopping_evaluator: CriteriaEvaluator - optional parameter to
                 provide alternative rules for committing to a decision.
-            query_agent(QueryAgent): the query selection mechanism of the system
+            stimuli_agent(StimuliAgent): the query selection mechanism of the
+                system
 
         Functions:
             decide():
@@ -124,8 +125,8 @@ class DecisionMaker:
                  is_txt_stim=True,
                  stimuli_timing=[1, .2],
                  seq_constants=None,
-                 criteria_evaluator=CriteriaEvaluator.default(),
-                 query_agent=RandomStimuliAgent(
+                 stopping_evaluator=None,
+                 stimuli_agent=RandomStimuliAgent(
                      alphabet=list(string.ascii_uppercase) +
                               ['<'] + [SPACE_CHAR])):
         self.state = state
@@ -142,11 +143,15 @@ class DecisionMaker:
         self.time = 0
         self.sequence_counter = 0
 
+        # TODO: There should be a better way to do this!
         # Stopping Criteria
-        self.min_num_seq = min_num_seq
-        self.max_num_seq = max_num_seq
-
-        self.posterior_commit_threshold = decision_threshold
+        if stopping_evaluator:
+            self.stopping_evaluator = stopping_evaluator
+        else:
+            self.stopping_evaluator = \
+                CriteriaEvaluator.default(min_num_seq=min_num_seq,
+                                          max_num_seq=max_num_seq,
+                                          threshold=decision_threshold)
 
         self.last_selection = ''
 
@@ -154,9 +159,8 @@ class DecisionMaker:
         self.seq_constants = seq_constants
 
         # TODO: change this criteria evaluator with
-        self.criteria_evaluator = criteria_evaluator
 
-        self.query_agent = query_agent
+        self.stimuli_agent = stimuli_agent
 
     def reset(self, state=''):
         """ Resets the decision maker with the initial state
@@ -170,7 +174,7 @@ class DecisionMaker:
         self.time = 0
         self.sequence_counter = 0
 
-        self.query_agent.reset()
+        self.stimuli_agent.reset()
 
     def form_display_state(self, state):
         """ Forms the state information or the user that fits to the
@@ -209,13 +213,8 @@ class DecisionMaker:
 
         self.list_epoch[-1]['list_distribution'].append(p[:])
 
-        params = dict(
-            min_num_seq=self.min_num_seq,
-            max_num_seq=self.max_num_seq,
-            threshold=self.posterior_commit_threshold)
-
         # Check stopping criteria
-        if self.criteria_evaluator.should_commit(self.list_epoch[-1], params):
+        if self.stopping_evaluator.should_commit(self.list_epoch[-1]):
             self.do_epoch()
             return True, None
         else:
@@ -236,7 +235,8 @@ class DecisionMaker:
         self.list_epoch.append({'target': None, 'time_spent': 0,
                                 'list_sti': [], 'list_distribution': []})
 
-        self.query_agent.do_epoch()
+        self.stimuli_agent.do_epoch()
+        self.stopping_evaluator.do_epoch()
 
     def schedule_sequence(self):
         """ Schedules next sequence """
@@ -265,7 +265,7 @@ class DecisionMaker:
                 """
 
         # querying agent decides on possible letters to be shown on the screen
-        query_els = self.query_agent.update_and_query(
+        query_els = self.stimuli_agent.update_and_query(
             self.list_epoch[-1]['list_distribution'][-1])
         # once querying is determined, append with timing and color info
         stimuli = rsvp_seq_generator(query=query_els,
--- a/bcipy/tasks/rsvp/main_frame.py
+++ b/bcipy/tasks/rsvp/main_frame.py
@@ -256,7 +256,9 @@ class DecisionMaker:
                  stimuli_timing=[1, .2],
                  seq_constants=None,
                  criteria_evaluator=CriteriaEvaluator.default(),
-                 query_agent=RandomAgent()):
+                 query_agent=RandomAgent(
+                     alphabet=list(string.ascii_uppercase) +
+                              ['<'] + [SPACE_CHAR])):
         self.state = state
         self.displayed_state = self.form_display_state(state)
         self.stimuli_timing = stimuli_timing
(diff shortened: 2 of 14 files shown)


bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -57,15 +60,26 @@ def __init__(self, min_num_seq, max_num_seq, signal_model=None, fs=300, k=2,
         seq_constants = []
         if backspace_always_shown and BACKSPACE_CHAR in alp:
             seq_constants.append(BACKSPACE_CHAR)
+
+        # Stimuli Selection Module
+        stopping_criteria = CriteriaEvaluator(
+            continue_criteria=[MinIterationsCriteria(min_num_seq)],
+            commit_criteria=[MaxIterationsCriteria(max_num_seq),
+                             ProbThresholdCriteria(decision_threshold)])
```
lawhead, line 68: If we want to use different stopping criteria or use a different Agent, how would we do that? It seems like we would need to repeat a lot of code that's in this constructor. It may be worth breaking this init method up into smaller methods that then get called from here.

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -57,15 +60,26 @@ def __init__(self, min_num_seq, max_num_seq, signal_model=None, fs=300, k=2,
...
+
+        # TODO: Change this hard number 8!
```
lawhead: Are you planning on making this change for the PR?

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -62,135 +62,6 @@ def save_history(self):
...
-                threshold: float - minimum likelihood required
-        """
-        if any(
-                criteria.apply(epoch, params)
-                for criteria in self.continue_criteria):
-            return False
-        return any(
-            criteria.apply(epoch, params) for criteria in self.commit_criteria)
-
-
```
lawhead, line 193: These definitely belonged in a different module. Thanks for pulling them out.

bcipy/tasks/rsvp/main_frame.py
```diff
@@ -244,15 +117,17 @@ class DecisionMaker:
...
                  is_txt_stim=True,
                  stimuli_timing=[1, .2],
                  seq_constants=None,
-                 criteria_evaluator=CriteriaEvaluator.default()):
+                 stopping_evaluator=CriteriaEvaluator.default(min_num_seq=2,
+                                                              max_num_seq=10,
+                                                              threshold=0.8),
+                 stimuli_agent=RandomStimuliAgent(
+                     alphabet=list(string.ascii_uppercase) +
+                              ['<'] + [SPACE_CHAR])):
```
lawhead, line 130: There is a potential bug here if someone provides a custom alphabet but does not include a different stimuli_agent. The stimuli_agent parameter should be given a default value of `None` and constructed in the __init__ method if it's not provided, using the alphabet parameter. Also, the default value for alphabet should use the helpers/task.py so we're not repeating code.

bcipy/tasks/rsvp/query_mechanisms.py
```diff
@@ -0,0 +1,199 @@
...
+            Return:
+                query(list[str]): queries """
+        return
+
+    def do_epoch(self):
+        """ If the system decides on a class let the agent know about it """
+        return
+
+
+class RandomStimuliAgent(StimuliAgent):
```
lawhead, line 45: We should have class documentation here.

bcipy/tasks/rsvp/query_mechanisms.py
```diff
@@ -0,0 +1,199 @@
...
+        tmp = [i for i in self.alphabet]
+        query = random.sample(tmp, self.len_query)
+
+        return query
+
+    def do_epoch(self):
+        pass
+
+
+class NBestStimuliAgent(StimuliAgent):
```
lawhead, line 65: This class needs documentation as well.

bcipy/tasks/rsvp/query_mechanisms.py
```diff
@@ -0,0 +1,199 @@
...
+        the prior information to the system. 
+        There are two competing terms;
+            I: mutual information, this utilizes the current posterior
+            M: momentum, this utilizes all likelihoods so far
+        and the queries are selected by linearly combining these two.
+        This agent, in the beginning of the epoch, explores utilizing M and 
+        eventually exploits the knowledge using I  """
+
+    def __init__(self, alphabet, len_query=4, lam=1., gam=1., dif_lam=.09,
+                 update_lam_flag=True):
```
lawhead: We are now using a version of python that supports type annotations and we are trying to make it a practice to annotate any new code. So this would be:
  ` def __init__(self, alphabet: List[str], len_query: int = 4, lam: float = 1., gam: float = 1., dif_lam: float = .09, update_lam_flag: bool = True):`

bcipy/tasks/rsvp/query_mechanisms.py
```diff
@@ -0,0 +1,199 @@
...
+        entropy_term[np.isnan(entropy_term)] = 0
+
+        # update the momentum term using likelihoods
+        momentum = self._compute_momentum(list_distribution)
+        num_passed_sequences = len(list_distribution)
+
+        # if there are no sequences shown yet, the momentum cannot be computed
+        if num_passed_sequences > 1:
+            reward = (self.lam_ - 1) * entropy_term + (
+                    self.lam_ / num_passed_sequences) * momentum
```
lawhead, line 141: The indentation is wrong here, according to pylint. I think you can just put this on a single line.

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+
+        """
+        raise NotImplementedError()
+
+
+class MinIterationsCriteria(DecisionCriteria):
+    """ Returns true if the minimum number of iterations have not yet
+        been reached. """
+
+    def __init__(self, min_num_seq):
```
lawhead: This is a good refactor. I agree that `min_num_seq` should be passed in as a constructor parameter rather than in the `decide` method. You could add a type annotation of `int` here.

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+            return True
+        return False
+
+
+class ProbThresholdCriteria(DecisionCriteria):
+    """Returns true if the commit threshold has been met."""
+
+    def __init__(self, threshold):
+        """ Args:
+                threshold(float \in [0,1]): A threshold on most likely
```
lawhead: Can this backslash be removed? It causes pylint errors.

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+        current_distribution = epoch['list_distribution'][-1]
+        if np.max(current_distribution) > self.tau:
+            log.debug("Committing to decision: posterior exceeded threshold.")
+            return True
+        return False
+
+
+class MarginCriteria(DecisionCriteria):
+    """ Stopping criteria based on the difference of two most likely candidates.
+        This condition
```
lawhead: It looks like the documentation string is incomplete here?

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+
+    def __init__(self, margin):
+        """ Args:
+                margin(float \in [0,1]): Minimum distance required between
```
lawhead: This backslash is also an issue for pylint.

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+        return d
+
+
+class MomentumCommitCriteria(DecisionCriteria):
+    """ Stopping criteria based on Renyi entropy on the simplex
+        Renyi entropy for infinity and 1 norms are special forms.
+        Attr:
+            alpha(float): a non-negative order for the entropy
+            tau(float): decision threshold
+            """
```
lawhead, line 147: The documentation does not match the parameters here.

bcipy/tasks/rsvp/stopping_criteria.py
```diff
@@ -0,0 +1,231 @@
...
+
+        tmp_p = np.ones(len(p)) * (1 - self.tau) / (len(p) - 1)
+        tmp_p[0] = self.tau
+
+        tmp = -np.sum(p * np.log2(p + eps))
+        tau_ = -np.sum(tmp_p * np.log2(tmp_p + eps))
+
+        tmp_ = np.array(prob_history)
+        mom_ = tmp_[1:] * (
+                np.log(tmp_[1:] + eps) - np.log(tmp_[:-1] + eps))
```
lawhead, line 170: The indentation for this line isn't correct. This can all be on one line.

branch_test/copy_phrase_simulation.py
```diff
@@ -0,0 +1,186 @@
+""" This script simulates a copy phrase task. """
```
lawhead, line 1: It seems like this is not the right package name for these files. Maybe @sci-tab has a suggestion here.

lawhead (approved):
It looks like a nice refactor. I'm still doing some testing but it seems like with a little cleanup this will be good to merge.

branch_test/copy_phrase_simulation.py
tab-cmd: I agree... These would not be installed with the main package as it stands. How about a `bcipy.tasks.simulate` module? 

Then,

- add `synthesized_user.py` to `bcipy.tasks.simulate` as user.py.

I'm flexible on the naming of the module, but that location seems apporatiaite. What do yall think? 

As for this file... we have a `bcipy.tasks.rsvp`. We could add it there under a simulation or tests directory.

branch_test/copy_phrase_simulation.py
```diff
@@ -0,0 +1,186 @@
...
+                    decision_maker.displayed_state.replace('_', ' '))
+                lm_prior = l_model.get_prob()
+            else:
+                lm_prior = np.ones(len_alphabet) / len_alphabet
+
+            prob = conjugator.update_and_fuse({evidence_names[0]: lm_prior})
+            prob_ = np.array([i for i in prob])
+
+            # Update the decision maker
+            d, sti = decision_maker.decide(prob_)
```
tab-cmd, line 135: unless doing math and needing x, y, theta, etc, single-letter variables are highly discouraged. `decision, stimuli` would be more descriptive here and help future readers of your code.

branch_test/copy_phrase_simulation.py
tab-cmd: Actually, this runs like a script, so maybe we can just throw it in a demo? The drawback is we don't catch regressions as easily and it won't be as reusable. Your call, but I'd recommend breaking this into methods.

bcipy/helpers/copy_phrase_wrapper.py
azizkocana: Since stopping condition right now is a module, just like the stimuli agent, it should be passed to decision maker from a parameter file. I am not sure how to do that and need your assistance. The structure of the stopping criteria was already designed including a continue criteria list and a commit criteria list. I think this also refactoring.

bcipy/helpers/copy_phrase_wrapper.py
azizkocana: This is another parameter issue. Before stimuli agent, number of stimuli was passed to decision maker. Right now this information should pass to a stimuli agent and then the stimuli agent should be inserted into the decision maker accordingly. I need your input and opinion on this.

bcipy/tasks/rsvp/stopping_criteria.py
azizkocana: I guess by incorrect you mean the coding standards.
All indentation is covered with my on the fly PEP8 checker that has a fixed column length.
Please lemme know of your column length number so I can adjust accordingly.

bcipy/tasks/rsvp/query_mechanisms.py
azizkocana: I guess by incorrect you mean the coding standards.
All indentation is covered with my on the fly PEP8 checker that has a fixed column length.
Please lemme know of your column length number so I can adjust accordingly.

bcipy/tasks/rsvp/main_frame.py
azizkocana: I need your input there. I am not sure how I can bundle the input parameter file with the classmethod.

bcipy/tasks/rsvp/query_mechanisms.py
lawhead: When I ran pylint on this file there were a couple of errors for `Wrong hanging indentation (remove 4 spaces).` So I think this has more to do with how it's being wrapped than the column length. If you run the `autotpep8` tool on this file it looks like it fixes the indentation by removing 4 preceding spaces.

bcipy/helpers/copy_phrase_wrapper.py
lawhead: I just looked back at our previous code and it looks like the stim_length parameter was never used in the Copy Phrase task. We always assume a length of 10. We should create a followup ticket to use that parameter here, but in the meantime maybe just change the value for len_query to 10 for consistency.

bcipy/tasks/rsvp/main_frame.py
lawhead: I'm proposing using the method signature:
   `..., stimuli_agent=None):`

then in the init method:
   `self.stimuli_agent = stimuli_agent or RandomStimuliAgent(alphabet=self.alphabet)`

bcipy/helpers/copy_phrase_wrapper.py
celikbasak: len_query is hard coded as 10. TODO: Parametrize len_query in the future releases.

bcipy/tasks/rsvp/query_mechanisms.py
celikbasak: Class documentation added.

bcipy/tasks/rsvp/query_mechanisms.py
celikbasak: Class documentation added.

bcipy/tasks/rsvp/query_mechanisms.py
celikbasak: I also ran pylint on this file, I did not come across `Wrong hanging indentation (remove 4 spaces)` in line 132. However I've seen many other linting errors. How should we handle the rest of the errors, for example which styling errors do we ignore?

I attached the terminal output. Please note that the line numbers may be different than yours because I added class documentation.
Screen Shot 2021-04-06 at 12 23 43

bcipy/tasks/rsvp/stopping_criteria.py
celikbasak: `int` was added by Aziz.

bcipy/tasks/rsvp/stopping_criteria.py
celikbasak: Documentation seems to be completed by Aziz.

bcipy/tasks/rsvp/stopping_criteria.py
celikbasak: This documentation was also fixed.

bcipy/tasks/rsvp/stopping_criteria.py
celikbasak: When I ran pylint, I did not come across an indentation error either. 
Side question: Is there a way to adjust the column length?

branch_test/copy_phrase_simulation.py
celikbasak: `synthesized_user.py` is renamed as `user.py` and added into `bcipy.tasks.simulate` module.

`copy_phrase_simulation.py` is added into `bcipy.tasks.rsvp.simulation` module.

After solving a few parameter load errors, they will be ready to be merged.

branch_test/copy_phrase_simulation.py
celikbasak: parameter names are updated as `decision, stimuli`

bcipy/tasks/rsvp/main_frame.py
celikbasak: The constructor is updated to `..., stimuli_agent=None):` 

And in the attributes, it is defined as:
`self.stimuli_agent = RandomStimuliAgent(alphabet=self.alphabet)`

bcipy/helpers/copy_phrase_wrapper.py
celikbasak: I will also need assistance on breaking the the init method into smaller chunks.

bcipy/tasks/rsvp/query_mechanisms.py
tab-cmd: @celikbasak Some of these can be resolved via `autopep8 --in-place --aggressive -r bcipy`. Others you'll need to address incrementally and the output suggests what is wrong (change import order, rename a variable).  

C0111 -> A docstring is needed. Something like, https://github.com/CAMBI-tech/BciPy/blob/1.5.1/bcipy/signal/model/model.py#L25.

W0613 -> that argument isn't used and linting is telling you that it likely can be removed.

W0231 -> super needs to be called. example, https://github.com/CAMBI-tech/BciPy/blob/1.5.1/bcipy/feedback/visual/visual_feedback.py#L20

Let me know if you run into issues resolving any and I can help!

bcipy/tasks/rsvp/stopping_criteria.py
tab-cmd: There is a way to change max_line_length. We set that in `setup.cfg`. I'd change it to one line in this case since it doesn't violate line length rules

bcipy/helpers/copy_phrase_wrapper.py
tab-cmd: I'm okay leaving this one and addressing it in another ticket. How can we further simplify constructing these criteria is the ask. It will take some time to think about.

PR #109, opened 2020-12-02
Params GUI Refactor

lawhead:
# Overview

This PR is a rewrite of the parameters GUI from WxPython to PyQt5. It also includes additional functionality to search for parameters.

## Ticket

https://www.pivotaltracker.com/story/show/175543725

## Contributions

- Refactor parameters form to use the Qt framework.
- Search functionality.

## Test

- Start the GUI from the command line or from the RSVPKeyboard interface. The form should display allowing customization
of parameters. Saved changes should persist.


tab-cmd:
@AlisterD I'm going to ask you to review this one with me. To do this, you'll need to checkout the `params-gui-refactor` branch, install everything from scratch, and run the parameters GUI. Acceptance Criteria (AC) should be in the pivotal ticket. We can set this up together at our meeting tomorrow morning as well :)  

@lawhead I promise to have the reviews back to you by Friday evening! Thanks for the patience.

bcipy/parameters/parameters.json
```diff
@@ -83,7 +83,7 @@
     "readableName": "Offline Analysis Alert Tone",
     "helpTip": "Specifies the path to an audio file to be played as an alert tone when offline analysis is complete. Default: bcipy/static/sounds/beep.wav",
     "recommended_values": "",
-    "type": "directorypath"
+    "type": "filepath"
```
tab-cmd, line 86: good catch!

bcipy/gui/params_form.py
```diff
@@ -1,392 +1,510 @@
...
+        return self.currentText()
 
 
-class Form(wx.Panel):
-    """The Form class is a wx.Panel that creates controls/inputs for each
-    parameter in the provided json file.
+#--- Input widgets
+
+
+class FormInput(QWidget):
```
tab-cmd: These are generic enough to be used elsewhere and would be useful to the experiment /field UI needs. What do you think about adding the classes to gui_main?

tab-cmd (approved):
A nice refactor! It works well for me locally. A few suggestions:

1. Move the generic classes to gui_main
2. Move the form to its own module (like `gui/parameters/param_form.py`). I'm not sure if we'll add more there, but it feels clean to have only the main, BCInterface, and README at GUI module root. We might add a README with use instructions there later? 

I'll still have Alister review tomorrow morning, but I don't forsee any major comments from that!

bcipy/gui/params_form.py
lawhead: I can see how these would be useful in the context of the experiment UI. We probably also want to pull out the method that selects the FormInput (subclass) based on the Parameter type. That code is currently in the ParamsForm widget but could be extracted out into a module function.

tab-cmd: I agree! If it's not too much effort in this PR let's do it. I'm okay making another ticket otherwise.

bcipy/gui/params_form.py
```diff
@@ -1,392 +1,510 @@
...
-
-        button_control_len = 1  # save button
-        rowlen = len(self.controls.keys()) + button_control_len
-        gridSizer = wx.FlexGridSizer(rows=rowlen, cols=1, vgap=10, hgap=10)
-
-        noOptions = dict()
+                form_input = TextInput(param)
+            self.controls[key] = form_input
+
+    def do_layout(self):
```
AlisterD: Add return types

AlisterD (approved):
Checked it out, tested it locally, seemed to work great! :D

bcipy/gui/params_form.py
lawhead: Thanks, good catch.

lawhead:
@sci-tab, I moved the generic classes into gui_main and put the params_form into its own module. Do you mind taking another quick look before I merge?

bcipy/gui/gui_main.py
```diff
@@ -54,6 +70,268 @@ def get_id(self):
...
+    Parameters:
+    ----------
+      parameter - the option which will be configured using this input. Used to populate the label,
+        help text, and relevant options.
+      help_font_size - font size for the help text.
+      help_color - color of the help text.
+    """
+
+    def __init__(self,
+                 parameter: Parameter,
```
tab-cmd: I don't think this should be parameter specific. Same with the functions inheriting it below. The `SearchInput` is generic and should be here for certain. 

Let's say I want to use the boolean input to query the user for something other than a parameter (ex. an experiment field). I do think it can be usefully abstracted and reused. I also don't want to extend your PR. I'd be okay leaving them in here or with the params_form and addressing with the experiment tickets. You call!

lawhead: Yeah, we should discuss in planning. Alternatively experiment fields could use the same format as a Parameter. We would just need to add a `required` boolean property.

PR #111, opened 2020-12-07
#175194315 Refactor BCInterface, add ExperimentRegistry

tab-cmd:
# Overview

Refactored BCInterface to select experiments and tasks. Refactored bci_main to accept a task and experiment id. Currently, bci_main and BCInterface do not enforce the field data collection. bci_main does not enforce a valid experiment (see ticket #176044196).

Created ExperimentRegistry.py for creating new experiments. Field creation must be added manually until ticket #175194521.

Currently, registered field names are not viewable in the UI. See ticket #176044412.

## Ticket

https://www.pivotaltracker.com/story/show/175194315

## Contributions

All changes not described here were formatting updates using autopep8. 

### Updated
- `.gitignore`: add private `.bcipy` to prevent custom experiments from being committed.
- `bci_main.py`: to use task and experiment
- `demo/bci_main_demo`: to work with updated bci_main arguments
- `README`: with new bci_main commands and updated UI. Add contributor. 
- `gui/README`: with ExperimentRegistry and other updates. Update contributor statement. 
- `tasks/task_registry`: updated values to indicate RSVP, add a `by_value` method. removed `by_mode` since it is not currently relevant. 
- `tasks/start_task`: updates exp_type to task to reflect verbiage elsewhere.
- `tests/test_save`:  updated to include task in the `init_save_data_structure` + added better dt indicator. 
- `tests/test_acquisition`: updated to include task in the `init_save_data_structure`

### Added

- `ExperimentRegistry.py`: UI for creating experiments. Field removal, editing, and creation not yet available. On load, it should require name and summary input for creation. It can add any fields in the `fields.json` after selecting register. It will prevent you from adding the same field twice. Once `Create Experiment` is selected it will save it in `.bcipy/experiment/experiments.json`.
- `helpers/load/`: added load experiments and fields methods.
- `helpers/system_utils/`: added default paths and filenames.
- `helpers/save`: added save experiments and fields methods.
- `tests/test_save`: added tests for saving and loading experiments.

### Removed
- `RSVPKeyboard.py`: removed in favor of the more generic BCInterface.

## Test

- Unit tests should pass. 
- Executing BCInterface, running all experiments in fake data mode using the default experiment. Select the add experiment button from that interface should open the ExperimentRegistry. The registry should function as described in the Contributions section.
- bci_main should function as indicated via help text with all flags. defaults are available.

## Documentaion

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes. README and in-line documentation updated accordingly.


demo/bci_main_demo.py
```diff
@@ -7,21 +7,20 @@ def main():
     from bcipy.helpers.load import load_json_parameters
 
     # Load a parameters file
-    parameters = load_json_parameters(
-        'bcipy/parameters/parameters.json', value_cast=True)
+    parameters = 'bcipy/parameters/parameters.json'
```
lawhead: We should use the `DEFAULT_PARAMETERS_PATH` from bcipy.helpers.parameters here.

bcipy/tasks/task_registry.py
```diff
@@ -31,13 +31,13 @@ class ExperimentType(Enum):
     'Calibration'
     """
```
lawhead, line 32: This doctest should be updated as well. The returned value should be 'RSVP Calibration'. Also, should this enum be renamed to TaskType?

bcipy/helpers/save.py
```diff
@@ -25,33 +24,36 @@ def save_json_data(data: dict, location: str, name: str) -> str:
...
 
 
+def save_experiment_data(data, location, name) -> str:
+    return save_json_data(data, location, name)
+
+
+def save_field_data(data, location, name) -> str:
+    return save_json_data(data, location, name)
+
+
```
lawhead, line 34: It doesn't seem like these functions add much value. Maybe we just need to make the documentation for save_json_data more generic?

bcipy/helpers/load.py
```diff
@@ -38,6 +41,44 @@ def copy_parameters(path: str = DEFAULT_PARAMETERS_PATH,
...
+
+    Returns
+    -------
+        A dictionary of experiments, with the following format:
+            { name: { fields : {name: '', required: bool}, summary: '' } }
+    
+    """
+    with open(path, 'r') as json_file:
+        return json.load(json_file)
+
```
lawhead, line 59: These functions don't guarantee that the data will have the documented fields. The structure is assumed in a number of places but not enforced. It may be worth doing a check on load or converting the JSON to a namedtuple or object.

bci_main.py
```diff
@@ -4,14 +4,14 @@
...
-from bcipy.helpers.load import load_signal_model
+from bcipy.helpers.load import load_json_parameters, load_signal_model
 from bcipy.helpers.parameters import DEFAULT_PARAMETERS_PATH
 from bcipy.helpers.save import init_save_data_structure, DEFAULT_EXPERIMENT_ID
 from bcipy.tasks.start_task import start_task
 from bcipy.tasks.task_registry import ExperimentType
 
 
-def bci_main(parameter_location: str, user: str, exp_type: int, mode: str, experiment: str = DEFAULT_EXPERIMENT_ID) -> bool:
+def bci_main(parameter_location: str, user: str, task: int, experiment: str = DEFAULT_EXPERIMENT_ID) -> bool:
```
lawhead: Looks like the type annotation for the `task` parameter is inconsistent with the documentation.

bci_main.py
```diff
@@ -52,27 +51,20 @@ def bci_main(parameter_location: str, user: str, exp_type: int, mode: str, exper
...
                      version=sys_info['bcipy_version'])
 
     logging.getLogger(__name__).info(sys_info)
 
-    return execute_task(task_type, parameters, save_folder)
+    return execute_task(task, parameters, save_folder)
 
 
-def execute_task(task_type: dict, parameters: dict, save_folder: str) -> bool:
+def execute_task(task: str, parameters: dict, save_folder: str) -> bool:
```
lawhead: The documentation for this method needs to be updated as well.

bci_main.py
```diff
@@ -166,11 +158,11 @@ def _clean_up_session(display, daq, server):
     parser.add_argument('-p', '--parameters', default='bcipy/parameters/parameters.json',
                         help='Parameter location. Must be in parameters directory. Pass as parameters/parameters.json')
     parser.add_argument('-u', '--user', default='test_user')
-    parser.add_argument('-t', '--type', default=1,
+    parser.add_argument('-t', '--task', default='RSVP Calibration',
```
lawhead, line 158: The task_options should be updated to only include the names. We were previously inputting the integer value. Also, we may want to quote these names so it's clear what should be included.

bci_main.py
```diff
@@ -166,11 +158,11 @@ def _clean_up_session(display, daq, server):
...
                         help=f'Task type. Options: ({task_options})')
-    parser.add_argument('-m', '--mode', default='RSVP',
-                        help='BCI mode. Ex. RSVP, MATRIX, SHUFFLE')
+    parser.add_argument('-e', '--experiment', default=DEFAULT_EXPERIMENT_ID, 
+                        help='Select a valid experiment to run the task for this user')
     args = parser.parse_args()
 
     # Start BCI Main
-    bci_main(args.parameters, str(args.user), int(args.type), str(args.mode))
+    bci_main(args.parameters, str(args.user), str(args.task), str(args.experiment))
```
lawhead: Now that we're passing in the task name rather than the index it seems like it would be easy to pass in a wrong value. We could attempt to get the ExperimentType here so we fail fast if it's missing. It would also simplify the type annotations in the other methods.
  `bci_main(args.parameters, str(args.user), ExperimentType.by_value(args.task), str(args.experiment))`

lawhead (approved):
This looks great!

bcipy/tasks/task_registry.py
tab-cmd: I think that makes sense!

bcipy/helpers/save.py
tab-cmd: I wanted a good abstraction, so if we changed how we loaded or validated the function wouldn't have to change.

bcipy/helpers/load.py
tab-cmd: I have a validate function in the next PR for Experiment Fields! I had to break this up after it started getting bulky.

bci_main.py
tab-cmd: I like the idea of failing faster. I'll do that. In the next PR, I validate the experiment first thing as well in order to fail before any other processes are started.

PR #114, opened 2020-12-17
Field creation UI

tab-cmd:
# Overview

Added a GUI for creation of new BciPy experiment fields. Small updates for ExperimentRegistry to allow creation of multiple experiments per session.

## Ticket

https://www.pivotaltracker.com/story/show/175194521

## Contributions

linting

### Updated

- `gui/README`: with new GUI information
- `README`: with new GUI information
- `test_generator`: assert raises a Runtime error, which encapsulates the StopIteration. Not 100% on this one...
- `ExperimentRegistry`: to use WARN instead of INFO on check_input. To call new `FieldRegistry` instead of throwing not implemented alert.
- `gui_main`: to store background_color and pass instead of logging for dropdown action.

### Added

- `FieldRegistry`: GUI for creating new Fields. It should not create duplicates and require all field be input before creation.

## Test

- Run `BCInterface.py`, create a new experiment, create a new field, close the window. Ensure new field is written and viewable in the experiment creation window. Try creating invalid fields (missing values, names already registered) and verify alerts are thrown and no data is written. Run `make test-all`

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? READMEs were updated.

### DIFF
--- a/bcipy/acquisition/tests/datastream/test_generator.py
+++ b/bcipy/acquisition/tests/datastream/test_generator.py
@@ -2,7 +2,6 @@
 """Tests for datastream generator module"""
 from builtins import next
 import unittest
-import pytest
 from past.builtins import map, range
 from mock import mock_open, patch
 from bcipy.acquisition.datastream.generator import random_data, file_data
@@ -96,7 +95,8 @@ class TestGenerator(unittest.TestCase):
             for _ in range(row_count):
                 next(gen)
 
-            with pytest.raises(StopIteration):
+            with self.assertRaises(RuntimeError):
+                # a runtime error catches the StopIteration?
                 data.append(next(gen))
 
     def test_file_with_custom_encoder(self):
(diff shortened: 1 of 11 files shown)


bcipy/acquisition/tests/datastream/test_generator.py
```diff
@@ -96,7 +95,8 @@ def test_file_generator_end(self):
             for _ in range(row_count):
                 next(gen)
 
-            with pytest.raises(StopIteration):
+            with self.assertRaises(RuntimeError):
```
lawhead: Calling `next` on a depleted Generator should raise a `StopError` exception, which is what we're explicitly testing in this situation. A RuntimeError is too general and could be the result of a different unexpected issue.

lawhead (approved):

PR #116, opened 2021-01-07
Scrollable fields (add toggle for required and remove)

tab-cmd:
# Overview

Refactor BCIGui into a Widget that can accept other widgets. Add ScrollableFrame and LineItems. Implemented a scrollable area with line items for registered fields (can toggle required and remove). 

## Ticket

https://www.pivotaltracker.com/story/show/176044412

## Contributions

- `ExperimentRegistry`: methods to construct a scrollable area and render registered fields and LineItems. Functionality to remove and toggle registered fields added. 
- `gui_main`: BCIGui refactor into a widget, allows more widgets to be added. Use QV / QH Layouts (instead of grid). Add ScrollableFrame and LineItems: two classes of QWidgets.  
- `BCInterface`: changes to support the refactored BCIGui
- `session.py`: changes to prevent issues with names containing space characters

## Test

- Run `BCInterface`, create new experiment using +, add a field and toggle requirement, attempt to remove, create a new experiment, select it in BCInterface, ensure the fields were collected as expected once start experiment is selected.
- Run `make test-all` and `make lint`

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? In-line documentation updated... I don't believe more are needed...

### DIFF
--- a/bcipy/gui/experiments/ExperimentRegistry.py
+++ b/bcipy/gui/experiments/ExperimentRegistry.py
@@ -1,7 +1,7 @@
 import sys
 import subprocess
 
-from bcipy.gui.gui_main import BCIGui, app, AlertMessageType
+from bcipy.gui.gui_main import BCIGui, app, AlertMessageType, ScrollableFrame, LineItems
 
 from bcipy.helpers.load import load_experiments, load_fields
 from bcipy.helpers.save import save_experiment_data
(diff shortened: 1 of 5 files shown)


bcipy/gui/experiments/ExperimentRegistry.py
```diff
@@ -31,14 +31,118 @@ def __init__(self, *args, **kwargs):
...
+            the required field ('true' or 'false').
+        """
+        field_name = self.window.sender().get_id()
+        for field in self.experiment_fields:
+            if field_name in field:
+                required = field[field_name]['required']
+                if required == 'false':
+                    field[field_name]['required'] = 'true'
+                else:
+                    field[field_name]['required'] = 'false'
```
lawhead, line 79: There are a couple places in this class where you are checking the 'required' field against a string value ('false' or 'true'). It may simplify your code if you convert this field to a boolean value when you load the experiment fields, and then convert it back to a string when you serialize the values. This code could then be simplified to:
    
    field[field_name]['required'] = not field[field_name]['required']

bcipy/gui/gui_main.py
```diff
@@ -620,6 +638,117 @@ def get_filename_dialog(self,
...
+        self.frame.setWidgetResizable(True)
+        self.frame.setFixedWidth(self.width)
+        self.setFixedHeight(self.height)
+
+        # if a widget is provided, add it to the scrollable frame
+        if widget:
+            self.widget = widget
+            self.frame.setWidget(widget)
+
+        # add the fram and set the layout
```
lawhead: There's a minor typo in the comment here.

lawhead (approved):

bcipy/gui/experiments/ExperimentRegistry.py
tab-cmd: I could see either way! The benefit of this approach is we don't have to loop over the experiment fields on save (it's formatted in the correct structure). This isn't trivial as it's a dict of dicts. Whereas, already have to loop over this struct during the field operations. I'll leave it here for now but I'm open to changing

PR #118, opened 2021-01-12
Validate Field inputs

tab-cmd:
# Overview

Add Integer and Float Form Inputs and apply in our form collection. Update alert messages in Experiment Field.

## Ticket

https://www.pivotaltracker.com/story/show/176210324

## Contributions

- `gui_main`: add `IntegerInput` and `FloatInput`. These use spinboxes which will prevent string or other data from being collected. It also casts the value from the form using `.value` to the appropriate type. 
- `ExperimentField`: Update to use new type inputs. Improve alert messages. Add alert message for a value error. 
- `param_form`: Update to use new type inputs

## Test

- Start BCInterface, Start Experiment, Fill in the Age information. Ensure it saves correctly. Update Age to be a float @ `.bcipy/fields/fields.json` and do the procedure above. 

## Documentaion

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Verify the updates you did here. No changes are necessary.

### DIFF
--- a/bcipy/gui/experiments/ExperimentField.py
+++ b/bcipy/gui/experiments/ExperimentField.py
@@ -163,7 +163,7 @@ class ExperimentFieldCollection(QWidget):
         """Build Save Data."""
         try:
             for field in self.field_inputs:
-                _input = field.value()
+                _input = field.cast_value()
                 name = field.label.strip(self.require_mark)
                 self.save_data[name] = _input
         except ValueError as e:
--- a/bcipy/gui/experiments/ExperimentField.py
+++ b/bcipy/gui/experiments/ExperimentField.py
@@ -22,7 +22,9 @@ from bcipy.gui.gui_main import (
     BoolInput,
     DirectoryInput,
     FileInput,
+    FloatInput,
     FormInput,
+    IntegerInput,
     TextInput
 )
 from bcipy.helpers.load import load_experiments, load_fields
@@ -38,6 +40,8 @@ class ExperimentFieldCollection(QWidget):
     field_data: List[tuple] = []
     field_inputs: List[FormInput] = []
     type_inputs = {
+        'int': IntegerInput,
+        'float': FloatInput,
         'bool': BoolInput,
         'filepath': FileInput,
         'directorypath': DirectoryInput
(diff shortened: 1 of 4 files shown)


bcipy/gui/gui_main.py
```diff
@@ -196,6 +196,58 @@ def widgets(self) -> List[QWidget]:
...
+        """Override FormInput to create a spinbox."""
+        spin_box = QDoubleSpinBox()
+        spin_box.setMaximum(100000)
+        return spin_box
+
+    def value(self) -> float:
+        """Override FormInput to return as a float."""
+        if self.control:
+            return float(self.control.text())
+        return None
```
lawhead, line 255: The parent class has the type signature `def value(self) -> str`, and there are at least two places in the code that depend on a `str` return type and will likely throw runtime errors: `FormInput#matches` method and the `ParamsForm#update_parameters` method.

Since you need the values cast to their types for validation, we could create an additional method in FormInput with a return type of `Any`, so we have (`cast_value -> Any`, `value -> str`) or (`value -> Any`, `str_value -> str`). I would still like the widget itself to know how to cast its value to a str, since there otherwise ends up being quite a bit of duplication if we have the downstream user do it.

bcipy/gui/gui_main.py
```diff
@@ -196,6 +196,58 @@ def widgets(self) -> List[QWidget]:
...
+        """Override FormInput to create a spinbox."""
+        spin_box = QSpinBox()
+        spin_box.setMaximum(100000)
+        return spin_box
+
+    def value(self) -> str:
+        """Override FormInput to return an integer value."""
+        if self.control:
+            return int(self.control.text())
+        return None
```
lawhead, line 229: See comment below regarding the return type. The signature here says it returns a `str`, but the code may return an `int`.

bcipy/gui/experiments/ExperimentField.py
```diff
@@ -157,16 +161,26 @@ def save(self) -> None:
...
+                _input = field.value()
+                name = field.label.strip(self.require_mark)
+                self.save_data[name] = _input
+        except ValueError as e:
+            self.throw_alert_message(
+                title='Error',
+                message=f'Error saving data. Invalid value provided. \n {e}',
+                message_type=AlertMessageType.WARN,
+                okay_or_cancel=True
+            )
```
lawhead, line 175: One limitation of this approach is that it only shows one error one at a time if there are multiple issues. But I'm guessing these experiment forms won't be too long so that's not a huge deal.

tab-cmd: That's true! I don't anticipate this error being thrown often or at all because of how it's being collected. It's not possible to add letters or characters to the spinbox for instance. I did want something to catch these things just in case. I'll address your other comment and see how that might change things.

bcipy/gui/gui_main.py
```diff
@@ -196,6 +203,58 @@ def widgets(self) -> List[QWidget]:
...
+        spin_box.setMaximum(100000)
+        return spin_box
+
+    def cast_value(self) -> float:
+        """Override FormInput to return as a float."""
+        if self.control:
+            return float(self.control.text())
+        return None
+
+
```
lawhead, line 257: It might be worth adding a `cast_value` to the BoolInput as well.

lawhead (approved):

bcipy/gui/gui_main.py
tab-cmd: I do think we should add it in the future, for this ticket, it would add another item to address for writing to json. If I cast to a boolean and dump it, it would store as:

`{"key": true}` instead of `{"key": "true"}`

PR #120, opened 2021-01-13
User input fixes

AlisterD:
# Overview
User ID verification with the following criteria: 
1. Maximum length of 25 letters or numbers
2. Minimum length of one letter or number
3. No special characters
4. No spaces

Added User ID Verification for the following 

#176210324

https://www.pivotaltracker.com/story/show/175563946

## Contributions

- Created function user_id_validator, imported regex to help with validation, and called function in start_experiment 

Tested as much as I could with combinations oh spaces, special charts, numbers, 

- ran make bci-gui for every case

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,3 +22,6 @@ clean:
 	find . -name "*.py[co]" -o -name __pycache__ -exec rm -rf {} +
 	find . -path "*/*.pyo"  -delete
 	find . -path "*/*.pyc"  -delete
+
+bci-gui:
+	python bcipy/gui/BCInterface.py
--- a/bcipy/gui/BCInterface.py
+++ b/bcipy/gui/BCInterface.py
@@ -4,7 +4,15 @@ import sys
 
 from typing import List
 
-from bcipy.gui.gui_main import BCIGui, app, AlertMessageType, AlertResponse
+from bcipy.gui.gui_main import (
+    AlertMessageType,
+    AlertResponse,
+    app,
+    BCIGui,
+    contains_special_characters,
+    contains_whitespaces,
+    invalid_length,
+)
 from bcipy.helpers.load import load_json_parameters, load_experiments, copy_parameters
 from bcipy.helpers.parameters import DEFAULT_PARAMETERS_PATH
 from bcipy.tasks.task_registry import TaskType
@@ -22,6 +30,8 @@ class BCInterface(BCIGui):
     default_text = '...'
     padding = 30
     btn_height = 40
+    max_length = 25
+    min_length = 1
 
     def __init__(self, *args, **kwargs):
         super(BCInterface, self).__init__(*args, **kwargs)
@@ -38,6 +48,13 @@ class BCInterface(BCIGui):
         self.user = None
         self.experiment = None
         self.task = None
+
+        self.user_id_validations = [
+            (invalid_length(min=self.min_length, max=self.max_length),
+                f'User ID must contain between {self.min_length} and {self.max_length} alphanumeric characters.'),
+            (contains_whitespaces, 'User ID cannot contain white spaces'),
+            (contains_special_characters, 'User ID cannot contain special characters')
+        ]
 
     def build_buttons(self) -> None:
         """Build Buttons.
@@ -284,7 +301,7 @@ class BCInterface(BCIGui):
     def check_input(self) -> bool:
         """Check Input.
 
-        Checks to make sure user has input all required fields. Currently, only user id is required.
+        Checks to make sure user has input all required fields.
         """
 
         # Update based on current inputs
@@ -294,12 +311,7 @@ class BCInterface(BCIGui):
 
         # Check the set values are different than defaults
         try:
-            if self.user == BCInterface.default_text:
-                self.throw_alert_message(
-                    title='BciPy Alert',
-                    message='Please input a User ID',
-                    message_type=AlertMessageType.INFO,
-                    okay_to_exit=True)
+            if not self.check_user_id():
                 return False
             if self.experiment == BCInterface.default_text:
                 self.throw_alert_message(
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -22,3 +22,6 @@ clean:
 	find . -name "*.py[co]" -o -name __pycache__ -exec rm -rf {} +
 	find . -path "*/*.pyo"  -delete
 	find . -path "*/*.pyc"  -delete
+
+bci-gui:
```
tab-cmd, line 26: Nice addition!

bcipy/gui/BCInterface.py
```diff
@@ -369,7 +418,8 @@ def start_experiment(self) -> None:
         Using the inputs gathers, check for validity using the check_input method, then launch the experiment using a
             command to bci_main.py and subprocess.
         """
-        if self.check_input():
+
+        if self.check_input() and self.user_id_validator():
```
tab-cmd: This method should go in the `check_input` method, which already extracts the self.user.

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
...
+        """Check User ID Requirements. User ID must follow the following Requirements:
+
+        1. Maximum length of 25 letters or numbers
+        2. Minimum length of at least one letter or  number
+        3. No special characters
+        4. No spaces
+        """
+
+        # Update based on current inputs
+        self.user = self.user_input.currentText()
```
tab-cmd: As mentioned below this value should already be set when you call this in `check_input()`. In the current flow, we would extract user from the input and assign self.user twice.

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
             return False
         return True
 
+    def user_id_validator(self) -> bool:
+        """Check User ID Requirements. User ID must follow the following Requirements:
+
```
tab-cmd, line 341: Great documentation!

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
...
+        4. No spaces
+        """
+
+        # Update based on current inputs
+        self.user = self.user_input.currentText()
+        allowed_characters = re.compile('[^0-9a-zA-Z]+')
+
+        # Check the length of minimum one character up to 25 characters
+        try:
+            if len(self.user) == 0 or len(self.user) > 25:
```
tab-cmd: Consider assigning a `length_user` variable since you call it twice, and you could then append it in the message of the alert using f strings,

`f'User ID must be of at least one or up to 25. Your current ID is {length_user} characters.'`

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
...
+                self.throw_alert_message(
+                    title='BciPy Alert',
+                    message='User ID must be of at least one or up to 25 non-special characters',
+                    message_type=AlertMessageType.INFO,
+                    okay_to_exit=True)
+                return False
+        # Check for white space
```
tab-cmd: These comments are off.. try running `make lint` and see if it will catch this formatting stuff for you. If not, you'll need to go in and fix!

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
             return False
         return True
 
+    def user_id_validator(self) -> bool:
+        """Check User ID Requirements. User ID must follow the following Requirements:
+
+        1. Maximum length of 25 letters or numbers
```
tab-cmd: We could add these as class-level variables at the top. For instance,

`self.max_length = 25` in the init

and then use that value instead of the hardcoded 25. Update the documentation to reflect that we set it above. This will make changing it easier in the future. I'd do this for all cases. `self.min_length`, `self.special_character_regex`

tab-cmd (changes requested):

bcipy/gui/BCInterface.py
AlisterD: deleted from function

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,54 @@ def check_input(self) -> bool:
...
+        # Check for white space
+            if re.match(r"^(?=.*[\s])", self.user):
+                self.throw_alert_message(
+                    title='BciPy Alert',
+                    message='User ID cannot contain white spaces',
+                    message_type=AlertMessageType.INFO,
+                    okay_to_exit=True)
+                return False
+        # Check for special characters
+            if allowed_characters.search(self.user):
```
lawhead, line 362: Here it looks like you are failing validation if there are _any_ allowed characters. Shouldn't this be the opposite and only raise the alert if _all_ of the characters are allowed?

lawhead: Sorry, my bad.

bcipy/gui/BCInterface.py
AlisterD: Oh that make sense! thank you for the feedback!

bcipy/gui/BCInterface.py
```diff
@@ -327,6 +328,59 @@ def check_input(self) -> bool:
...
+                return False
+        # If Any other error
+        except Exception as e:
+            self.throw_alert_message(
+                title='BciPy Alert',
+                message=f'Error, {e}',
+                message_type=AlertMessageType.CRIT,
+                okay_to_exit=True)
+            return False
+        return True
```
lawhead, line 368: There is a lot of repeated code in this method that follows the same pattern:  the user_id is checked against some condition and if it evaluates to True an alert is displayed with an error message specific to that condition. In addition to the increased code, this is also difficult to fully test and will only provide the user with 1 error message at a time. If the user_id has multiple problems this can be a frustrating experience for the end user.

One way to avoid repeating yourself is to abstract out the validations as separate functions (each of which can be independently tested), and store the validation functions along with the error message. For example:

    # These functions could be in a separate module
    def invalid_length(min=1, max=25):
        return lambda string: len(string) < min or len(string) > max

    def contains_whitespaces(string):
        return re.match(r"^(?=.*[\s])", string)

    def has_special_characters(string):
        disallowed_chars = re.compile('[^0-9a-zA-Z]+')
        return bool(disallowed_chars.search(string))
    
    # This would go in the class definition
    user_id_validations = [
      (invalid_length(min=self.min_length, max=self.max_length), f'must contain between {self.min_length} and {self.max_length} alphanumeric characters'),
      (contains_whitespaces, 'cannot contain white spaces'),
      (has_special_characters, 'cannot contain special characters') 
    ]

Then in the `user_id_validator` method you can iterate through these validations, check them against the user_id and collect the error messages for any that return True. If any problems are detected they can all be displayed to the user.

AlisterD: This is so elegant!

AlisterD (commented):
It looks great~

tab-cmd (approved):

PR #122, opened 2021-01-14
Acquisition device refactor

lawhead:
# Overview

Refactored the acquisition module to separate the concept of a device (ex. DSI-24 headset) and a connection method to that device (TCP or LSL).

## Ticket

https://www.pivotaltracker.com/story/show/175322346

## Contributions

- Created additional parameters for acq_device and acq_connection_method.
- Added DeviceSpec class and a module for loading a list of supported devices from configuration; refactored many classes to take a DeviceSpec parameter, including the fake data servers.
- Created an enumeration for supported ConnectionMethods.
- Refactored previous concept of Device -> Connector.
- Refactored fake data servers to simplify their usage. Renamed data_server module to tcp_data_server.
- Fixed all acquisition demos.
- Refactored lsl_connector in preparation for future work with multiple sources.

## Test

- The new acquisition parameters should appear in the params form and be editable.
- All acquisition module tests should run without error.
- All acquisition demos should run correctly.
- Data acquisition during an experiment should use the configured device and connection method parameters and should behave the same as before, including trigger calibration.

## Documentaion

- Updated the README in the acquisition module.


bcipy/acquisition/client.py
```diff
@@ -87,6 +90,9 @@ def __init__(self,
         self._max_wait = 0.1  # for process loop
 
         self.marker_writer = NullMarkerWriter()
+
+        # column in the acquisition data used to calculate offset.
+        self.offset_column = OFFSET_COLUMN_NAME
```
tab-cmd: I think it might be more meaningful to call it a trigger column and require that it's present. We're going to refactor the way we handle offset correction in a future ticket.

bcipy/acquisition/datastream/lsl_server.py
```diff
@@ -22,39 +29,39 @@ class LslDataServer(StoppableThread):
...
-        stream_name = params.get('name', name)
-        self.generator = generator
+        self.device_spec = device_spec
+        self.generator = generator or random_data_generator(channel_count=device_spec.channel_count)
 
-        log.debug("Starting server with params: %s", str(params))
-        info = StreamInfo(stream_name,
-                          "EEG", len(channels),
-                          self.sample_hz,
+        log.debug("Starting LSL server for device: %s", device_spec.name)
```
tab-cmd, line 54: use f strings here

bcipy/acquisition/devices.json
```diff
@@ -0,0 +1,54 @@
...
+    {
+        "name": "DSI-VR300",
+        "content_type": "EEG",
+        "channels": [ "P4", "Fz", "Pz", "F7", "PO8", "PO7", "Oz", "TRG" ],
+        "sample_rate": 300.0,
+        "connection_methods": ["TCP", "LSL"],
+        "description": "Wearable Sensing DSI-VR300"
+    },
+    {
+        "name": "g.USBamp-1",
```
tab-cmd, line 20: I like that we're registering our devices this way

bcipy/acquisition/devices.py
```diff
@@ -0,0 +1,107 @@
...
+        map(ConnectionMethod.by_name, config.get('connection_methods', [])))
+    return DeviceSpec(name=config['name'],
+                      content_type=config['content_type'],
+                      channels=config['channels'],
+                      connection_methods=connection_methods,
+                      sample_rate=config['sample_rate'],
+                      description=config['description'])
+
+
+SUPPORTED_DEVICES = {}
```
tab-cmd: let's move this to the top of the file

bcipy/acquisition/devices.py
```diff
@@ -0,0 +1,107 @@
...
+    configuration file."""
+    global SUPPORTED_DEVICES
+    with open(config_path, 'r', encoding='utf-8') as json_file:
+        specs = [make_device_spec(entry) for entry in json.load(json_file)]
+        SUPPORTED_DEVICES = {spec.name: spec for spec in specs}
+
+
+def supported_devices() -> Dict[str, DeviceSpec]:
+    """Returns the currently supported devices keyed by name."""
+    global SUPPORTED_DEVICES
```
tab-cmd: why is this call needed?

bcipy/acquisition/devices.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def register(device_spec: DeviceSpec):
+    """Register the given DeviceSpec."""
+    global SUPPORTED_DEVICES
+    SUPPORTED_DEVICES[device_spec.name] = device_spec
+
+
+# Initialize the SUPPORTED_DEVICES
+load()
```
tab-cmd: Could this be called in init_acquistion or in some other fashion?

bcipy/acquisition/protocols/lsl/lsl_connector.py
```diff
@@ -46,49 +49,87 @@ def inlet_name(inlet) -> str:
     return name.replace('-', '_')
 
 
-class LslDevice(Device):
-    """Driver for any device streaming data through the LabStreamingLayer lib.
+def channel_names(stream_info: pylsl.StreamInfo) -> List[str]:
```
tab-cmd, line 52: seems like these could be a part of the lsldevice class?

bcipy/acquisition/protocols/lsl/lsl_connector.py
```diff
@@ -231,8 +230,7 @@ def read_data(self):
             if not marker.is_empty and timestamp >= marker.timestamp:
                 trg = marker.trg
                 log.debug(("Appending %s marker %s to sample at time %s; ",
```
tab-cmd, line 232: use f strings

bcipy/acquisition/tests/test_client.py
```diff
@@ -181,6 +192,43 @@ def test_offset(self):
...
+
+        col_name = 'PHOTODIODE'
+        channels = ['ch1', 'ch2', col_name]
+        sample_hz = 100
+        trigger_at = 10
+        num_records = 500
+        n_channels = len(channels) - 1
+
+        data = [
+            mock_record(n_channels) + [0 if (i + 1) < trigger_at else 1]
```
tab-cmd, line 206: It could be useful to define a function to produce this or set it as a class value

tab-cmd (approved):
Some suggestions, but works locally for me and I like the approach!

bcipy/acquisition/datastream/lsl_server.py
lawhead: Logging is actually the one place where python is optimized to use the old string formatting style and f-strings can lead to some inefficiencies: https://docs.python.org/3/howto/logging.html#optimization. In this particular case it's probably not a big deal, but some code linters will provide a warning.

bcipy/acquisition/client.py
lawhead: Yeah, I went back and forth on this one. I'll go ahead and rename it to trigger_column for clarity.

bcipy/acquisition/devices.py
lawhead: I'll re-work this a bit to call `load` when SUPPORTED_DEVICES is first accessed. There is other code that uses this module, so I don't want to make that code always have to explicitly load devices, but I agree that this doesn't need to be done at import time.

bcipy/acquisition/devices.py
lawhead: I should prefix `SUPPORTED_DEVICES` with an underscore. My intent is that it's a private implementation detail of the module and shouldn't be accessed directly.

bcipy/acquisition/datastream/lsl_server.py
tab-cmd: I didn't know that! I just assumed f strings are more performant.

bcipy/acquisition/protocols/lsl/lsl_connector.py
lawhead: I originally had this as a class method, but it didn't reference any instance variables and I got a linter warning about it so I pulled it out into a utility function. It may be useful in other contexts when we are looking at multiple LSL sources.

bcipy/acquisition/tests/test_client.py
lawhead: Yeah, I still want to take a pass through the code and revisit mocked EEG data.

PR #126, opened 2021-03-29
EDF conversion fixes

tab-cmd:
# Overview

Updates to improve EDF data conversions. Applying trigger offsets, ability to write targetness, and handling cases where there are too many annotations. 

## Ticket

- https://www.pivotaltracker.com/story/show/176668858
- https://www.pivotaltracker.com/story/show/176650169

## Contributions

- `convert.py`: Add mode, write_targetness, and annotation_channels keyword arguments.
- `demo_convert.py`: Update to add new keywords
- `helpers/load.py`: add extract_mode method for determining the mode
- `helpers/triggers.py`: add an offset correction method
- `helpers/vizualization.py`: moved plot_edf function from demo to this module.
- `task_registry.py`: Update tasks to include the necessary mode

## Test

- Added unittest
- Converted several files and had @danielpklee confirm in BrainVision

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes. I added task/mode to the README dictionary.

### DIFF
--- a/bcipy/helpers/convert.py
+++ b/bcipy/helpers/convert.py
@@ -12,10 +12,13 @@ from bcipy.helpers.load import load_json_parameters, read_data_csv, extract_mode
 from bcipy.helpers.triggers import trigger_decoder, apply_trigger_offset, trigger_durations
 
 
+logger = logging.getLogger(__name__)
+
+
 def convert_to_edf(data_dir: str,
                    edf_path: str = None,
                    overwrite=False,
-                   write_targetness=True,
+                   write_targetness=False,
                    use_event_durations=False,
                    mode=False,
                    annotation_channels=None) -> Path:
@@ -34,7 +37,7 @@ def convert_to_edf(data_dir: str,
     overwrite - If True, the destination file (if it exists) will be overwritten.
         If False (default), an error will be raised if the file exists.
     write_targetness - If True, and targetness information is available, write
-        that instead of the stimuli markers.
+        that instead of the stimuli markers. False by default.
     mode - optional; for a given task, define the task mode. Ex. 'calibration', 'copy_phrase'.
         If not provided, it will be extracted from the data_dir.
     use_event_durations - optional; if True assigns a duration to each event.
@@ -51,7 +54,7 @@ def convert_to_edf(data_dir: str,
 
     params = load_json_parameters(Path(data_dir, 'parameters.json'),
                                   value_cast=True)
-    raw_data, _, ch_names, _, sfreq = read_data_csv(
+    raw_data, _, ch_names, _, sample_rate = read_data_csv(
         Path(data_dir, params['raw_data_name']))
     durations = trigger_durations(params) if use_event_durations else {}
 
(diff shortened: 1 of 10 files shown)


bcipy/helpers/convert.py
```diff
@@ -80,12 +124,15 @@ def write_edf(output_path: str,
...
     pmin, pmax = [raw_data.min(), raw_data.max()]
+
     n_channels = len(raw_data)
 
     try:
         writer = EdfWriter(str(output_path),
                            n_channels=n_channels,
                            file_type=FILETYPE_EDFPLUS)
+        if annotation_channels:
+            writer.set_number_of_annotation_signals(annotation_channels)
```
lawhead, line 162: I read through the documentation for this method and it seems like this is needed any time the number of triggers/annotations exceeds the number of seconds in the experiment. Wouldn't this almost always be the case since we usually show about 3 to 4 stim per second? Should we try to calculate this from the parameters (using stim_length)?

bcipy/helpers/convert.py
```diff
@@ -41,19 +55,45 @@ def convert_to_edf(data_dir: str,
...
+    triggers = compile_triggers(
+        symbol_info, trial_target_info, trigger_timing, write_targetness)
+
     events = edf_annotations(triggers, durations)
 
-    return write_edf(edf_path, raw_data, ch_names, sfreq, events, overwrite)
+    return write_edf(edf_path, raw_data, ch_names, sfreq, events, overwrite, annotation_channels)
+
+
+def compile_triggers(labels, targetness, timing, write_targetness):
```
lawhead: This method could use a docstring.

bcipy/helpers/convert.py
```diff
@@ -8,16 +8,23 @@
...
-from bcipy.helpers.triggers import read_triggers, trigger_durations
+from bcipy.helpers.load import load_json_parameters, read_data_csv, extract_mode
+from bcipy.helpers.triggers import trigger_decoder, apply_trigger_offset, trigger_durations
 
 
 def convert_to_edf(data_dir: str,
                    edf_path: str = None,
                    overwrite=False,
-                   use_event_durations=False) -> Path:
+                   write_targetness=True,
```
lawhead: Do you think it's a more common scenario that we would want to see the targetness rather than the stimuli? Should the default be False?

bcipy/helpers/tests/test_convert.py
```diff
@@ -92,25 +97,63 @@ def test_convert_defaults(self):
...
+                              mode=self.default_mode,
                               edf_path=Path(self.temp_dir, 'mydata.edf'))
 
         self.assertEqual(Path(path).name, 'mydata.edf')
+
+    def test_with_write_targetness(self):
+        """Test creating the EDF without event annotations"""
+        path = convert_to_edf(self.temp_dir,
+                              mode=self.default_mode,
+                              write_targetness=True,
```
lawhead, line 129: If this is the default we probably want a separate test where `write_targetness=False` to check that the stim name is written correctly.

lawhead (approved):
Nice work figuring this out. This one was tricky!

bcipy/helpers/convert.py
tab-cmd: In practice this wasn't clear cut and forcing the annotation channels led to some issues. I added a warning method to let the user know some triggers may be lost and to increment gradually.

PR #127, opened 2021-04-06
Session module

lawhead:
# Overview

Refactored session data from a dict to a class-based data structure. Refactored copy phrase task and session helpers to use the new data structures.

## Ticket

https://www.pivotaltracker.com/story/show/176150973

## Contributions

- Refactored session data into separate data classes
- Refactored Copy Phrase task to use new data classes
- Refactored session helper to use new classes
- Added unit tests for session helper

## Test

- Ran the Copy Phrase task and confirmed that the session output was consistent with previously generated data.
- Used the session tools on the generated output to confirm that the session helper was still working correctly and that the output was correct.
- Ran unit tests.


bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
+"""Module for functionality related to session-related data."""
+from typing import Dict, List
+from collections import Counter
+
+
+class StimSequence:
```
tab-cmd: We can call this an inquiry I think?

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+    """Represents a sequence of stimuli."""
+
+    def __init__(self,
+                 stimuli: List[str],
+                 eeg_len: int,
```
tab-cmd: We could make this a little more generic. `data_len`.

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+                 timing_sti: List[float],
```
tab-cmd: timing may be sufficient as a name

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+                 triggers: List[List],
+                 target_info: List[str],
+                 target_letter: str,
+                 current_text: str,
+                 copy_phrase: str,
```
tab-cmd: This should be optional in the case of other tasks (free_spell). Perhaps, `target_text`?

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+        self.eeg_len = eeg_len
+        self.timing_sti = timing_sti
+        self.triggers = triggers
+        self.target_info = target_info
+        self.target_letter = target_letter
+        self.current_text = current_text
+        self.copy_phrase = copy_phrase
+        self.next_display_state = next_display_state
+
+        # TODO: refactor for multimodal; List of Evidences?
```
tab-cmd, line 45: I like that idea. We can make another ticket for that if you'd like

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+class StimSequence:
+    """Represents a sequence of stimuli."""
+
+    def __init__(self,
+                 stimuli: List[str],
+                 eeg_len: int,
+                 timing_sti: List[float],
+                 triggers: List[List],
+                 target_info: List[str],
+                 target_letter: str,
```
tab-cmd: Optional as well

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+            alphabet - list of stim in the same order as the evidences.
+            n - number of most likely elements to include
+        """
+        likelihood = dict(zip(alphabet, self.likelihood))
+        return {
+            'stimuli': self.stimuli,
+            'lm_evidence': dict(zip(alphabet, self.lm_evidence)),
+            'eeg_evidence': dict(zip(alphabet, self.eeg_evidence)),
+            'likelihood': likelihood,
+            'most_likely': dict(Counter(likelihood).most_common(5))
```
tab-cmd: implement `n` here but rename to most_likely or something related

bcipy/tasks/session_data.py
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+
+class Session:
+    """Represents a data collection session. Not all tasks record session data."""
+
+    def __init__(self,
+                 save_location: str,
+                 session_type: str = 'Copy Phrase',
+                 paradigm: str = 'RSVP'):
```
tab-cmd: `mode`

bcipy/helpers/tests/resources/mock_session/session.json
```diff
@@ -0,0 +1,2431 @@
...
+            "T",
+            1.5507312049157917
+          ],
+          [
+            "<",
+            1.767507885931991
+          ],
+          [
+            "F",
+            1.984226447995752
```
tab-cmd, line 1185: Since this is a test, we could reduce this to a few inquiries? It could reduce the difficulty in looking over this data in tests

bcipy/helpers/tests/resources/mock_session/session.json
```diff
@@ -0,0 +1,2431 @@
+{
+  "session": "data/default/test_user_001/test_user_001_RSVP_Copy_Phrase_Wed_17_Mar_2021_17hr11min05sec_-0700",
+  "session_type": "Copy Phrase",
+  "paradigm": "RSVP",
```
tab-cmd: Wherever we write this it should be mode

bcipy/helpers/tests/resources/mock_session/session.json
```diff
@@ -0,0 +1,2431 @@
+{
+  "session": "data/default/test_user_001/test_user_001_RSVP_Copy_Phrase_Wed_17_Mar_2021_17hr11min05sec_-0700",
+  "session_type": "Copy Phrase",
```
tab-cmd: We might rename this to task

tab-cmd (changes requested):
Still testing locally, but this is enough to get you rolling! Great refactoring effort here!

We should add `session` to the glossary in this PR - this is the latest from my last PR: https://github.com/CAMBI-tech/BciPy/blob/bids-trigger-fix/README.md#glossary

The glossary terms themselves are negotiable (ex. mode vs paradigm) but we should document them and make tickets to update as needed!

bcipy/tasks/session_data.py
lawhead: I was looking through the usage and what this is referring to is the number of channels in the EEG data. This is going to be the same value for all Inquiries, so I'm not sure it makes sense to store this here. If anything it should be a Session variable, but I'm wondering if it even pertains to sessions at all. It seems like the concern of session-related data should be around what was presented and how decisions were made. Maybe this parameter should be removed. Thoughts?

tab-cmd: I agree, let's remove it. I think it was used for debugging at one point.

lawhead:
I like the naming suggestions. They make the code/data a lot more readable and consistent.

README.md
```diff
@@ -60,7 +60,7 @@ To use all the goodies locally (including the GUI and demo scripts)
 If wanting the latest version from PyPi:
 1. `pip install bcipy`
 
-Alternatley, if [Make](http://www.mingw.org/) is installed, you may run the follow command to install:
+Alternately, if [Make](http://www.mingw.org/) is installed, you may run the follow command to install:
```
tab-cmd, line 63: Thanks for catching the typos 🙈

tab-cmd (approved):

PR #132, opened 2021-04-19
Add abstract base class for SignalModel

nik-sm:
# Overview

Added an abstract base class `SignalModel` in preparation for refactoring the current PCA/RDA/KDE model and then integrating a new neural network model.

## Ticket

https://www.pivotaltracker.com/story/show/177691543

## Contributions

`bcipy/signal/model/model.py` - Add abstract base class
`bcipy/signal/model/__init__.py` - Make abstract base class available using `from bcipy.signal.model import SignalModel`

## Test

No tests added for this abstract class.

## Documentation

The base class has simple descriptions of the purpose of each method. Once the PCA/RDA/KDE model is refactored to this structure, more documentation can be added if useful. Alternatively, a "dummy" example class could be added if that would be useful.

### DIFF
--- a/bcipy/signal/model/model.py
+++ b/bcipy/signal/model/model.py
@@ -1,0 +1,57 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import List
+from dataclasses import dataclass
+
+import numpy as np
+
+
+class SignalModel(ABC):
+    @abstractmethod
+    def fit(self, training_data: np.array, training_labels: np.array) -> SignalModel:
+        """
+        Train the model using the provided data and labels.
+        Return self for convenience.
+        """
+        ...
+
+    @abstractmethod
+    def evaluate(self, test_data: np.array, test_labels: np.array) -> ModelEvaluationReport:
+        """Compute model performance characteristics on the provided test data and labels."""
+        ...
+
+    @abstractmethod
+    def predict(self, data: np.array, presented_symbols: List[str], all_symbols: List[str]) -> np.array:
+        """Using the provided data, compute log likelihoods over the entire symbol set."""
+        ...
+
+    @abstractmethod
+    def save(self, path: Path) -> None:
+        """Save model state to the provided checkpoint"""
+        ...
+
+    @abstractmethod
+    def load(self, path: Path) -> None:
+        """Restore model state from the provided checkpoint"""
+        ...
+
+    @abstractmethod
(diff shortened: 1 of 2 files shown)


bcipy/signal/model/model.py
```diff
@@ -0,0 +1,57 @@
...
+        """
+        ...
+
+    @abstractmethod
+    def evaluate(self, test_data: np.array, test_labels: np.array) -> ModelEvaluationReport:
+        """Compute model performance characteristics on the provided test data and labels."""
+        ...
+
+    @abstractmethod
+    def predict(self, data: np.array, presented_symbols: List[str], all_symbols: List[str]) -> np.array:
```
tab-cmd: how do you feel about `inquiry` and `symbol_set`?

bcipy/signal/model/model.py
```diff
@@ -0,0 +1,57 @@
...
+        """Save model state to the provided checkpoint"""
+        ...
+
+    @abstractmethod
+    def load(self, path: Path) -> None:
+        """Restore model state from the provided checkpoint"""
+        ...
+
+    @abstractmethod
+    def __eq__(self, other) -> bool:
```
tab-cmd: I think it would be better to add this as it became necessary unless we can enforce it

bcipy/signal/model/model.py
```diff
@@ -0,0 +1,57 @@
...
+        """Might be useful for confirming that model.load(model.save) == model"""
+        ...
+
+
+@dataclass
+class ModelEvaluationReport:
+    """
+    Describes model performance characteristics.
+
+    TODO - add more attributes as needed. Note that the model
```
tab-cmd: Cleanup docstring. This is a great idea, I like encapsulating these report metrics.

tab-cmd (approved):
Awesome to see this! Thanks, Niklas! 😄 Let's add some information about creating new models using the base class to the README.md. https://github.com/CAMBI-tech/BciPy/tree/main/bcipy/signal

lawhead (approved):

bcipy/signal/model/model.py
nik-sm: Ok, makes sense. I'll remove it for now.

bcipy/signal/model/model.py
nik-sm: I think the names on their own may not be as self-explanatory, but the docstring can make it very explicit. Happy to change to those names and update doc

PR #133, opened 2021-04-19
Signal module add tests

nik-sm:
# Overview

Added tests for the current PCA/RDA/KDE model. Restructured some folders.

## Ticket

https://www.pivotaltracker.com/story/show/177691906
https://www.pivotaltracker.com/story/show/177691916

## Contributions

- `bcipy/helpers/visualization.py` - return figure handles from `generate_offline_analysis_screen()` to allow testing

- fix broken imports and reformat demos in `bcipy/signal/model/demo/mach_learning`. NOTE - almost all the changes in the demos are due to reformatting them with `black`
- flatten out the folder structure, such as `bcipy/signal/model/demo/mach_learning/generative_mods/function_density_estimation.py` -> `bcipy/signal/model/demo/mach_learning/density_estimation.py`
- `bcipy/signal/model/mach_learning/cross_validation.py` - remove unused `grid_search` function
- Remove unnecessary `DummyDimReduction` from `bcipy/signal/model/mach_learning/dimensionality_reduction.py`
- `bcipy/signal/tests/model` now contains unit tests in `test_model.py` (whose expected outputs are contained in the folder `unit_test_expected_output`) and integration tests in `test_offline_analysis.py` (with inputs in `integration_test_input` and outputs in `integration_test_expected_output`)
- `bcipy/signal/model/mach_learning/train_model.py` - the value of `auc_init` was being calculated using the _optimized_ choices of `gam` and `lam`, but it should actually be calculated _before_ finding these optimized values. The effect of this is that it would appear that cross_validation did not help (because both the "before" and "after" are using the same values of `gam` and `lam`) whereas now it should be apparent that using values produced by cross_validation gives some improvement. This is also explained in a comment in that file.

#### Integration tests

The integration test runs offline analysis on a single real sample, saves the outputs, and checks:
- AUC number (note the model pkl file contents are NOT compared currently)
- lik_dens figure
- mean_erp figure

It may be desirable to test other outputs from the model during the integration test, but I'm not sure which ones.

#### Unit tests

- The unit tests use fake data to test each part of the model (`test_pca`, `test_rda`, `test_kde_plot`, `test_kde_values`) - these check that the result of `fit` followed by `transform` is the same as `fit_transform`, and also the same as the result currently produced by the model. 
- There is also a test of the cross validation logic (`test_cv`), but this test is less sensitive. It may be hard to design a sensitive test, because cross validation is used to optimize 2 parameters for the model, and it is not clear (to me at least) how these parameters should vary as a function of the input data or the model structure.
- There is a small test of the `inference` function (`test_inference`) to confirm that the model correctly assigns a high likelihood ratio for the target letter. 
- Added placeholder tests for notch and bandpass in `bcipy/signal/tests/process/test_frequency.py`. The analytic expected result for these filters needs to be computed, and then the tests can be finished. They currently fail with `NotImplementedError`.

### pytest-mpl
Note - this PR adds a dependency on `pytest-mpl`, which provides a decorator that makes it easy to compare figure outputs using the following structure:
```
@pytest.mark.mpl_image_compare(baseline_dir=..., filename=...)
def function_to_test():
    ...
    return fig  # function must return a figure handle or something with ".savefig" method
```

When the code is in a known working state, generate the expected results figures:
`pytest bcipy/signal/model/ [-k <NAME_OF_TEST>] --mpl-generate-path=<PATH_TO_FOLDER_OF_EXPECTED_RESULTS>`

Then to run the tests, include the flag `--mpl` (otherwise these tests will be skipped). This flag does not affect other standard tests, so it can always be safely included.

See more about `pytest-mpl` at https://pypi.org/project/pytest-mpl/.

Note that the threshold for comparison can be tuned, but the default seems sensitive enough.

### Running tests

`coverage` is giving incorrect results (it claims a line is uncovered, but adding `raise` or `breakpoint()` shows that the line is in fact covered). This might be due to an interaction with `pytest-mpl` or some other issue with my local environment.
Instead, `pip install pytest-cov` and using the `--cov=<DIR>` option for pytest works, and `coverage html` still successfully parses this output into a nice html. Happy to discuss if this needs resolution:

Run unit tests with:
```
pytest --mpl --disable-warnings --cov=bcipy/signal/model -k TestPcaRdaKdeModel
```

Run integration tests with:
```
pytest --mpl --disable-warnings --cov=bcipy/signal/model -k TestPcaRdaKdeModel
```

## Documentation

- Inline documentation was added, but more may be required - please let me know what you think.


bcipy/signal/model/mach_learning/demo/demo_cv.py
```diff
@@ -1,23 +1,10 @@
...
-from bcipy.signal.model.mach_learning.classifier.function_classifier \
-    import RegularizedDiscriminantAnalysis
-from bcipy.signal.model.mach_learning.dimensionality_reduction.function_dim_reduction \
-    import DummyDimReduction
+from bcipy.signal.model.mach_learning.classifier import RegularizedDiscriminantAnalysis
+from bcipy.signal.model.mach_learning.dimensionality_reduction import ChannelWisePrincipalComponentAnalysis
 from bcipy.signal.model.mach_learning.pipeline import Pipeline
 
 
-def test_cv(x, y):
```
nik-sm, line 10: unused code

bcipy/signal/model/mach_learning/cross_validation.py
```diff
@@ -72,57 +72,6 @@ def cost_cross_validation_auc(model, opt_el, x, y, param, k_folds=10,
     return -auc, sc_h, y_valid_h
 
 
-def grid_search(model, opt_el, x, y, grid=[5, 5], op_type='cost_auc',
```
nik-sm, line 75: unused code

bcipy/signal/model/mach_learning/dimensionality_reduction.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
```
nik-sm, line 1: File is actually just moved from `bcipy/signal/model/mach_learning/dimensionality_reduction/function_dim_reduction.py`

bcipy/signal/model/mach_learning/classifier.py
```diff
@@ -27,11 +27,12 @@ class covariance matrix multiplied by number of data samples N_i in class i
         log_det_cov(list[float]): list of negative det(cov_i)
         inv_reg_cov_i(list[ndarray]): inverse of regularized covariance matrix for class i
 
+    TODO - Aziz mentioned it is possible to parallelize to speed up -- significant gains to be had
```
tab-cmd: Let's remove this TODO and add a ticket to the backlog to explore

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
```diff
@@ -21,35 +20,30 @@ def _demo_kde():
...
-    for kernel in ['gaussian', 'tophat', 'epanechnikov']:
-        kde = KernelDensityEstimate(kernel=kernel, bandwidth=bandwidth,
-                                    num_cls=1)
+    for kernel in ["gaussian", "tophat", "epanechnikov"]:
+        kde = KernelDensityEstimate(kernel=kernel, bandwidth=bandwidth, num_cls=1)
         kde.fit(x, y)
         log_dens = kde.list_den_est[0].score_samples(x_plot)
-        ax.plot(x_plot[:, 0], np.exp(log_dens), '-',
-                label="kernel = '{0}'".format(kernel))
+        ax.plot(x_plot[:, 0], np.exp(log_dens), "-", label="kernel = '{0}'".format(kernel))
```
tab-cmd: f strings would be more performant for string concat

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
```diff
@@ -21,35 +20,30 @@ def _demo_kde():
...
-    ax.legend(loc='upper left')
-    ax.plot(x[:, 0], -0.005 - 0.01 * np.random.random(x.shape[0]), '+k')
+    ax.legend(loc="upper left")
+    ax.plot(x[:, 0], -0.005 - 0.01 * np.random.random(x.shape[0]), "+k")
 
     ax.set_xlim(-4, 9)
     ax.set_ylim(-0.02, 0.4)
     plt.show()
-    print('KDE Flows!')
+    print("KDE Flows!")
```
tab-cmd: There is no rule for single or double quotes re: PEP, but the rest of the code leans toward single unless punctation is involved.

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
```diff
@@ -21,35 +20,30 @@ def _demo_kde():
...
     return 0
```
tab-cmd: Why return 0? If no return is expected, it can be left off and None will be returned.

bcipy/signal/model/mach_learning/demo/demo_dim_reduction.py
```diff
@@ -17,11 +16,11 @@ def _demo_cw_pca():
...
     y = cw_pca.fit_transform(x, var_tol=var_tol)
-    print('CW-PCA Results')
-    print('CW-PCA flows!')
+    print("CW-PCA Results")
+    print("CW-PCA flows!")
     y_2 = cw_pca.transform(x)
-    print('X: {} |--[CW-PCA]--> Y: {}'.format(x.shape, y.shape))
-    print('MSE:{} fit_transform'.format(np.sum(np.abs(y_2 - y))))
+    print("X: {} |--[CW-PCA]--> Y: {}".format(x.shape, y.shape))
+    print("MSE:{} fit_transform".format(np.sum(np.abs(y_2 - y))))
```
tab-cmd: Let's use f-strings while we are here!

bcipy/signal/model/mach_learning/demo/demo_rda.py
```diff
@@ -2,7 +2,7 @@
 
 
 def _demo_rda():
-    from bcipy.signal.model.mach_learning.classifier.function_classifier import \
+    from bcipy.signal.model.mach_learning.classifier import \
```
tab-cmd: Let's use a top-level import, encapsulating here isn't necessary

bcipy/signal/tests/process/test_frequency.py
```diff
@@ -0,0 +1,39 @@
...
+        # time indices
+        t = np.arange(0, duration_s - 1.0 / self.sample_rate_hz, 1.0 / self.sample_rate_hz)
+        self.data = np.zeros_like(t)
+        for f in frequencies:
+            self.data += np.cos((2 * np.pi * f) * t)  # set frequency to (2*pi*f) and step along each value of t
+
+    def test_notch(self):
+        notch_freq_hz = 60
+        filtered_data = notch.notch_filter(self.data, self.sample_rate_hz, frequency_to_remove=notch_freq_hz)
+        raise NotImplementedError()
```
tab-cmd: Let's remove these tests and add tickets to address

dev_requirements.txt
```diff
@@ -6,3 +6,4 @@ mockito==1.1.0
 pylint==2.3.1
 pytest==3.1.3
 twine==3.2.0
+pytest-mpl
```
tab-cmd: Let's require at least a major version

bcipy/helpers/demo/demo_convert.py
```diff
@@ -5,7 +5,7 @@
     `python bcipy/helpers/demo/demo_convert.py -p "path://to/bcipy/data/folder"`
 """
 from bcipy.helpers.convert import convert_to_edf
-from bcipy.helpers.vizualization import plot_edf
+from bcipy.helpers.visualization import plot_edf
```
tab-cmd, line 8: Thanks for the catch!

tab-cmd (changes requested):

bcipy/helpers/visualization.py
```diff
@@ -139,9 +144,13 @@ def generate_offline_analysis_screen(
                 bbox_inches='tight',
                 format='pdf')
 
+        fig_handles.append(fig)
+
     if show_figure:
         plt.show()
 
+    return fig_handles
```
lawhead, line 154: Now that we're returning a value we should update the return type in the method signature. I believe it currently says `-> None`.

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
lawhead: I was noticing this as well. We haven't discussed yet as a team or given any guidelines here. The Python literature doesn't seem to give any recommendations beyond being consistent per project. The Ruby community prefers single quotes for the most part, and reserves the use of double quotes for strings with interpolation. This tends to work well and I believe is my preference.

bcipy/signal/model/mach_learning/demo/demo_pipeline.py
```diff
@@ -60,35 +54,35 @@ def _demo_pipeline():
...
+                sv_final = [
+                    model.pipeline[0].list_pca[i].singular_values_ for i in range(len(model.pipeline[0].list_pca))
+                ]
                 print("Initial SV:{}".format(sv_init))
                 print("-- using tolerance:{} -->".format(var_tol))
                 print("Final SV:{}".format(sv_final))
 
-                print("Init dim.:{} -> Final dim.:{}"
-                      .format(x.shape, model.line_el[1].shape))
+                print("Init dim.:{} -> Final dim.:{}".format(x.shape, model.line_el[1].shape))
```
lawhead: Here is another place for an f-string.

bcipy/signal/model/offline_analysis.py
```diff
@@ -114,7 +116,10 @@ def offline_analysis(data_folder: str = None,
         offline_analysis_tone = parameters.get('offline_analysis_tone')
         play_sound(offline_analysis_tone)
 
-    return model
+    if return_fig_handles:
+        return model, fig_handles
+    else:      
+        return model
```
lawhead: This creates an awkward API in that the type signature depends on the parameters passed. In the case where we don't want the figure handles maybe this should return `model, None`, so we can still expect to get a tuple.

bcipy/signal/tests/model/test_offline_analysis.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def get_auc(model_filename):
+        match = re.search("^model_([.0-9]+).pkl$", model_filename)
+        if not match:
+            raise ValueError()
+        return float(match[1])
+
+    def test_model_AUC(self):
+        expected_auc = self.get_auc(list(expected_output_folder.glob("model_*.pkl"))[0].name)
```
lawhead, line 51: Do these tests work cross-platform? In the past we haven't been able to successfully share pkl files across platforms. I wonder about the numpy serialized data as well.

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
nik-sm: This change comes from black - my preference is for whatever is consistent and automatable. Since I was using black in another project, I had that enabled in VS Code, but I'm happy to switch to any other autoformatter so that we can all share a spec. 

The reason I really liked black is that it operates on the AST to guarantee that it maintains program correctness, it runs super fast, and it takes a strong stance to pre-empt all discussion related to formatting (such as single vs double quote, argument indentation, docstring formatting, etc etc), such that the defaults will always adhere to PEP guidelines while preventing the need to choose in ambiguous situations. As pointed out, there's no no PEP recommendation regarding `'` vs `"` so black just picks one and sticks to it.

I guess what I'm saying is, I'd be happy to pay bribes in snacks or beer to try using black (for which the only non-default setting I use is `--line-length=120`), but I would also be just as happy to use yapf or autopep8 or whatever works, provided we can share a config file that runs in VS Code.

I'd especially like to return to this topic in the context of CI/CD. Really in this case, my failure to adhere to a pre-existing standard (whether it was a stated rule or not) could easily be caught as an error on my local machine or on the build server, thus saving time for everyone.

Sorry for the long-winded reply, and especially sorry if I have missed some linting/formatting guidelines somewhere in the project!

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
nik-sm: These demos are a bit mysterious to me - I just tried to leave them unchanged for the most part. 2 of 7 of them are actually still just broken, and I don't know how to fix them.

The behavior of the rest is of debatable value IMO. 

In this case, now that there can be automated tests that (hopefully) also demonstrate the behavior/usage of this code, I would actually argue for deleting these demos wholesale - but I will defer that to your and @lawhead 's judgement on this. 

At the very least though, I don't think it's worth spending time cleaning them up only to leave them in their current non-informative state; if it is worth having demos in addition to tests, then perhaps we should delete these and add a ticket to create some new meaningful ones?

bcipy/signal/model/offline_analysis.py
nik-sm: Alternatively it could just always return both items, and discard the fig_handles where unwanted? Happy to do whichever you prefer

bcipy/signal/tests/model/test_offline_analysis.py
nik-sm: So the `*.pkl` file here is actually kind of a lie - it's just an empty file, but from what I saw, it's the only place where the model AUC is recorded. I used this as the way to compare model AUC, because it means the two results are compared in a uniform way, but if it's too weird/misleading to store an empty pkl file, I'd be happy to store the "expected" AUC in a different way as well.

`*.npy` files are cross-platform, so these should be no problem - here's a doc describing the format and its design goals in case that's useful: https://github.com/numpy/numpy/blob/main/doc/neps/nep-0001-npy-format.rst

nik-sm: But actually I should point out - I have not run these tests except on my linux machine - so this is is a good call and definitely worth sanity-checking

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
lawhead: I've used both black and yapf; I don't really have a preference so we should definitely discuss during retro and make a decision that's stated clearly somewhere. I know these formatting discussions can provide for a frustrating experience. We currently have a Makefile with a `lint` task that runs `autopep8`, but as you mentioned there are many ways of adhering to pep8 and this is insufficient by itself.

Also, we have a Pivotal Tracker ticket for CI/CD  (https://www.pivotaltracker.com/story/show/175336031) and it's something we would like to get into place, so maybe we should prioritize this higher. Feel free to add comments and suggestions to the ticket.

bcipy/signal/model/offline_analysis.py
lawhead: That works for me!

bcipy/signal/tests/model/test_offline_analysis.py
lawhead: Okay, cool. I'll check out the branch and try running the tests.

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
nik-sm: Sounds good. For now I just tried to revert these formatting changes in the demos, and also incorporate the suggested f-string changes/removing a few odds-and-ends that were left in the demos. 

I added a comment on the CI/CD ticket and would be happy to take a crack at that sometime. (I actually would have tried it out late one night, but realized that I can't experiment with Travis without certain permissions on the repo. I might start by forking the repo and playing around, and then I could report my findings)

bcipy/signal/model/mach_learning/demo/demo_density_estimation.py
nik-sm: For now, I'm just following the approach of "preserve the demos, remove obvious crud and don't reformat" - and I'll add a ticket to discuss possibly updating or removing them later.

bcipy/signal/tests/model/test_offline_analysis.py
lawhead: The tests all pass on a Mac! As you previously mentioned, the offline_analysis test takes a really long time, so I'll go ahead and create a ticket on the backlog for separating out integration tests so we can run our suite of unit tests more quickly.

lawhead:
Thanks for addressing the PR comments and for making this compatible with Python 3.6.5. I think it's ready to merge.

PR #136, opened 2021-04-20
Add function to compute probabilities after keyinput

nik-sm:
# Overview

Added a rule to compute probability updates after a user key press

## Ticket

https://www.pivotaltracker.com/story/show/177802383

## Contributions

- `bcipy/signal/model/keyinput.py` - added the function `compute_keyinput_probs`

## Test

Note - the logic for these tests is pretty redundant with the logic for the code itself; I cannot think of an alternate way to compute these probabilities that would be suitable for making an "independent" test (but the rule itself is relatively simple).

If someone has suggestions for additional tests, I'm happy to implement them.

In the meantime, run the tests with:
```
pytest -k TestKeyInput
```

## Documentation

Function is described in docstring.


bcipy/signal/model/keyinput.py
```diff
@@ -0,0 +1,34 @@
+from typing import List
+
+
+def compute_keyinput_probs(
```
tab-cmd: I think this could be more generic than keyinput... 1compute_probabilities_given_inquiry1 or something like that. The filename (keyinput.py) could also be more generic to that point. Whether it's key input or a different kind of evidence this method could be leveraged.

bcipy/signal/model/keyinput.py
```diff
@@ -0,0 +1,34 @@
...
+    based on their response. These likelihood terms will be multiplied with the prior probabilities of
+    each symbol and then normalized, in order to update the symbol probabilities.
+
+    Args:
+        inquiry - the symbols presented in the inquiry preview
+        symbol_set - the full alphabet of symbols
+        proceed - whether or not the user wants to proceed with this inquiry
+        user_error_prob - the probability that the user selects incorrectly during preview
+    """
+    K = len(inquiry)
```
tab-cmd: no single letter varibles. This should fail linting.

bcipy/signal/model/keyinput.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if proceed:  # user liked the inquiry; presented letters are upgraded and others are downgraded
+        shown_letter_value = (1 - user_error_prob) / K
+        other_letter_value = user_error_prob / (A - K)
+
+    else:  # user disliked the inquiry; presented are downgraded and others are upgraded
+        shown_letter_value = user_error_prob / K
+        other_letter_value = (1 - user_error_prob) / (A - K)
+
+    return {"KEYINPUT": [shown_letter_value if letter in inquiry else other_letter_value for letter in symbol_set]}
```
tab-cmd: Let's return the list and ask the caller to construct the dictionary or accept an evidence label that can be used here.

bcipy/signal/model/keyinput.py
```diff
@@ -0,0 +1,34 @@
+from typing import List
+
+
+def compute_keyinput_probs(
+    inquiry: List[str],
+    symbol_set: List[str],
+    user_error_prob: float,
+    proceed: bool,
+):
```
tab-cmd: Add a return type
example a funtion that accepts a list of strings and returns a bool:

`def test(arg1: List[str]) -> bool:`

tab-cmd (changes requested):

bcipy/signal/tests/model/test_keyinput.py
```diff
@@ -0,0 +1,41 @@
...
+        self.symbol_set = self.inquiry + ["Y"]
+        self.error_prob = 0.05
+
+    def test_user_likes_inquiry(self):
+        proceed = True
+        results = compute_keyinput_probs(self.inquiry, self.symbol_set, self.error_prob, proceed)["KEYINPUT"]
+
+        p_shown = (1 - self.error_prob) / len(self.inquiry)
+        p_not_shown = (self.error_prob) / (len(self.symbol_set) - len(self.inquiry))
+        expected = [p_shown] * 5 + [p_not_shown]
```
lawhead: Given the small symbol set, rather than computing the expected values it might be more clear to just manually list these out.

bcipy/signal/model/keyinput.py
nik-sm: At some point I'll be sure to setup up autoformatting and linting to match project standards - not sure if there's a config file already available that I didn't notice

bcipy/signal/model/keyinput.py
nik-sm: Ok - returning list sounds simplest to use imo

lawhead (approved):

PR #140, opened 2021-04-30
- fix random seed in additional necessary places for tests

nik-sm:
# Overview

Fix random seed, set y-axis scale, and removed unnecessary text to make unit test figures a more useful comparison and less fragile to changes in font libraries, etc.

## Ticket

https://www.pivotaltracker.com/story/show/177983173

## Contributions

- `bcipy/signal/tests/model/test_model.py` - set target letter for inference in a deterministic, obvious way. 

## Test

Before making changes, ran the tests with `pytest --mpl -k "TestPcaRdaKde"`. They passed, except for the position of text in the KDE plot (annoying but not relevant).

Then made changes.

After making changes, regenerated expected output with `pytest --mpl-generate-path=bcipy/signal/tests/unit_test_expected_output -k "TestPcaRdaKde"`

---

Here is the comparison on `test_kde_plot` function, showing that the difference is just the text:

Expected
baseline

Observed
result

Difference
result-failed-diff

---

Here is the comparison on the integration test `test_lik_dens` - the difference here seems to just be the overall image alignment.
I'm not sure what changed, but I suspect that somehow the original result was generated with a different matplotlib version or something:

Expected
baseline

Observed
result

Difference
result-failed-diff

## Documentation

No additional documentation


bcipy/signal/tests/model/test_model.py
```diff
@@ -56,30 +56,32 @@ def setUp(self):
         baseline_dir=expected_output_folder, filename="test_inference.expected.png", remove_text=True
     )
     def test_inference(self):
+        np.random.seed(0)
```
lawhead: Would it make sense to add `np.random.seed(0)` to the end of the `setup` method?

lawhead (approved):

bcipy/signal/tests/model/test_model.py
nik-sm: Ah - yes good call. 

In the refactor, I moved some of this to a `setUpClass` and that was causing the problems (the results depend on order of test execution), but in that case I'll still just add a `setUp` that does:

```python
def setUp(self):
    np.random.seed(0)
```

PR #141, opened 2021-04-30
Refactor pca rda kde

nik-sm:
# Overview

Refactor the current signal model to inherit from the new `SignalModel` base class.

## Ticket

https://www.pivotaltracker.com/story/show/177693352

## Contributions

- `bci_main.py` - move imports to top. 

- use new model API in:
  - `bcipy/helpers/copy_phrase_wrapper.py`
  - `bcipy/helpers/demo/demo_eeg_model_related.py`
  - `bcipy/helpers/tests/test_copy_phrase_wrapper.py`
  - `bcipy/signal/model/offline_analysis.py`

- `bcipy/helpers/load.py` - move `load_signal_model` to `bcipy/signal/model` module to prevent circular import errors
- `bcipy/helpers/tests/test_load.py` - test of load model is now in `bcipy/signal/tests/model/test_pca_rda_kde.py`
- `bcipy/helpers/task.py` - use `string.ascii_uppercase` for brevity

- `bcipy/helpers/visualization.py`   - An opinionated change - removed `lik_dens.pdf` figure entirely. 
    - This figure is not well documented and likely to be misinterpreted in its current form (and thus may be of low value). 
    - It is also very hard to test, since it is constructed by "reaching inside" the model to extract partially transformed data, which is not part of the new signal model API and will not be possible for other models. 
    - The next model will be an inquiry classifier, and not a generative model, so the current figure is no longer applicable for that case. Specifically, we will not be fitting 1D Gaussians, so there will not be a useful visualization of the class boundaries, nor will there be an estimate of the data density. 

- delete the previous model inference and training in `bcipy/signal/model/inference.py` and `bcipy/signal/model/mach_learning/train_model.py`

- rename the folder `bcipy/signal/model/mach_learning` to `bcipy/signal/model/pca_rda_kde` since it contains code specific to that model
- `bcipy/signal/model/pca_rda_kde/pca_rda_kde.py` - implement new model API for the existing signal model
- `bcipy/signal/tests/model/test_offline_analysis.py` - remove test of `lik_dens.pdf` figure
- `bcipy/signal/tests/model/test_pca_rda_kde.py` - reuse shared class setup. separate tests into internals and externals
- `bcipy/tasks/tests/test_main_frame.py` - use `alphabet()` helper fn

## Test

Running tests:
`pytest --mpl -k "TestPcaRdaKde"`
`pytest --mpl -k "TestCopyPhraseWrapper"`
`pytest --mpl -k "TestOfflineAnalysis"`

## Documentation

Updated some docstrings and `bcipy/signal/model/README.md`. This readme seems to need further details, or a link to a detailed writeup of the current model.


bci_main.py
```diff
@@ -1,16 +1,20 @@
 import logging
+import argparse
```
tab-cmd: I'd be okay leaving these in context since they are only used when the file is called, not when used as a module.

bci_main.py
```diff
@@ -1,16 +1,20 @@
...
 from bcipy.helpers.system_utils import get_system_info, configure_logger, DEFAULT_EXPERIMENT_ID
 from bcipy.helpers.language_model import init_language_model
-from bcipy.helpers.load import load_json_parameters, load_signal_model, load_experiments
+from bcipy.helpers.load import load_json_parameters, load_experiments
 from bcipy.helpers.validate import validate_experiment
 from bcipy.helpers.parameters import DEFAULT_PARAMETERS_PATH
 from bcipy.helpers.save import init_save_data_structure
 from bcipy.tasks.start_task import start_task
 from bcipy.tasks.task_registry import TaskType
+from bcipy.signal.model import load_signal_model
```
tab-cmd, line 14: We keep most loading functions in helpers.

bcipy/helpers/task.py
```diff
@@ -101,9 +102,7 @@ def alphabet(parameters=None, include_path=True):
                     stimulus_array.append(img)
             return stimulus_array
 
-    return ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
-            'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
-            BACKSPACE_CHAR, SPACE_CHAR]
+    return list(ascii_uppercase) + [BACKSPACE_CHAR, SPACE_CHAR]
```
tab-cmd, line 105: 👍

bcipy/helpers/visualization.py
```diff
@@ -119,35 +117,6 @@ def generate_offline_analysis_screen(
 
     fig_handles.append(fig)
 
-    if plot_lik_dens:
```
tab-cmd, line 122: Is this no longer needed? It helped us debug at some point

bcipy/signal/model/offline_analysis.py
```diff
@@ -8,14 +8,15 @@
...
 from bcipy.signal.process.filter import bandpass, notch, downsample
-from bcipy.signal.model.mach_learning.train_model import train_pca_rda_kde_model
 from bcipy.helpers.task import trial_reshaper
 from bcipy.helpers.visualization import generate_offline_analysis_screen
 from bcipy.helpers.triggers import trigger_decoder
 from bcipy.helpers.acquisition import analysis_channels,\
     analysis_channel_names_by_pos
 from bcipy.helpers.stimuli import play_sound
 
+from .pca_rda_kde import PcaRdaKdeModel
```
tab-cmd: full import needed here

bcipy/signal/model/pca_rda_kde/demo/demo_cv.py
```diff
@@ -1,8 +1,8 @@
 import numpy as np
-from bcipy.signal.model.mach_learning.cross_validation import cross_validation
```
tab-cmd: I like this change. much cleaner contract

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+from ..base_model import ModelEvaluationReport, SignalModel
+from .classifier import RegularizedDiscriminantAnalysis
+from .cross_validation import cost_cross_validation_auc, cross_validation
+from .density_estimation import KernelDensityEstimate
+from .dimensionality_reduction import ChannelWisePrincipalComponentAnalysis
+from .pipeline import Pipeline
+from scipy.stats import iqr
+
+
+class NotFittedError(Exception):
```
tab-cmd: Let's pull out exceptions to their own module within signal.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
+import pickle
+from pathlib import Path
+from typing import List
+
+import numpy as np
+from ..base_model import ModelEvaluationReport, SignalModel
+from .classifier import RegularizedDiscriminantAnalysis
+from .cross_validation import cost_cross_validation_auc, cross_validation
```
tab-cmd: Explicit imports needed

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+        # After finding cross validation scores do one more round to learn the
+        # final RDA model
+        model.fit(train_data, train_labels)
+
+        # Insert the density estimates to the model and train using the cross validated
+        # scores to avoid over fitting. Observe that these scores are not obtained using
+        # the final model
+        bandwidth = 1.06 * min(np.std(sc_cv), iqr(sc_cv) / 1.34) * np.power(train_data.shape[0], -0.2)
```
tab-cmd: Should this be a method `calculate_bandwidth`? It may be easier to change/test later.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+        model.add(KernelDensityEstimate(bandwidth=bandwidth))
+        model.pipeline[-1].fit(sc_cv, y_cv)
+
+        self.model = model
+        self._ready_to_predict = True
+        return self
+
+    def evaluate(self, test_data: np.array, test_labels: np.array) -> ModelEvaluationReport:
+        """
+        TODO - the way AUC is (and was) calculated seems weird and needs investigation/documentation
```
tab-cmd: Add documentation and make a ticket for the TODO / remove it here.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+        tmp_model.add(self.model.pipeline[1])
+
+        lam_gam = (self.model.pipeline[1].lam, self.model.pipeline[1].gam)
+        tmp, _, _ = cost_cross_validation_auc(
+            tmp_model, 1, test_data, test_labels, lam_gam, k_folds=self.k_folds, split="uniform"
+        )
+        auc = -tmp
+        return ModelEvaluationReport(auc)
+
+    def predict(self, data: np.array, inquiry: List[str], symbol_set: List[str]) -> np.array:
```
tab-cmd, line 85: docstring

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # Evaluate likelihood probabilities for p(e|l=1) and p(e|l=0)
+        scores = np.exp(self.model.transform(data))
+
+        # Evaluate likelihood ratios (positive class divided by negative class)
+        scores = scores[:, 1] / (scores[:, 0] + 1e-10) + 1e-10
+
+        # Compute likelihoods for entire symbol set.
+        # Letters not seen receive likelihood of 1
+        # TODO - shouldn't the unseen letters have reduce
```
tab-cmd: Let's raise this outside the codebase and create tickets to explore as necessary

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+        # This maps the likelihood distribution over the symbol_set
+        #   If the letter in the symbol_set does not exist in the target string,
+        #       it takes 1
+        likelihood_ratios = np.ones(len(symbol_set))
+        for idx in range(len(scores)):
+            likelihood_ratios[symbol_set.index(inquiry[idx])] *= scores[idx]
+        return likelihood_ratios
+
+    def save(self, path: Path):
```
tab-cmd, line 118: docstring

bcipy/signal/model/utils.py
```diff
@@ -0,0 +1,23 @@
+from .pca_rda_kde import PcaRdaKdeModel
```
tab-cmd: explicit imports

bcipy/signal/model/utils.py
```diff
@@ -0,0 +1,23 @@
...
+from tkinter import Tk
+from tkinter.filedialog import askopenfilename
+
+
+def load_signal_model(filename: str = None, model_class=PcaRdaKdeModel, model_kwargs={"k_folds": 10}):
```
tab-cmd: I'd prefer this live in helpers.load to leave how we load filenames etc from this module. It's not doing anything `signal` specific and is generic enough as written. I'd be okay leaving off defaults and forcing these to be passed as well.

+ needs docstring

bcipy/signal/tests/model/test_pca_rda_kde.py
```diff
@@ -1,85 +1,79 @@
...
 
-class TestPcaRdaKdeModel(unittest.TestCase):
-    """
-    Note - When the code is in a known & working state, generate "baseline" images using:
-    `pytest --mpl-generate-path=bcipy/signal/tests/model/unit_test_expected_output -k TestPcaRdaKdeModel`
-    """
-
-    def setUp(self):
-        np.random.seed(0)  # Set seed before generating test data
+class SharedSetup(unittest.TestCase):
```
tab-cmd: Name something more explicit? `ModelSetup`

bcipy/signal/tests/model/test_pca_rda_kde.py
```diff
@@ -195,5 +189,75 @@ def test_rda(self):
...
+
+    def setUp(self):
+        np.random.seed(0)
+        self.model = PcaRdaKdeModel(k_folds=10)
+        self.model.fit(self.x, self.y)
+        np.random.seed(0)
+
+    @pytest.mark.mpl_image_compare(
+        baseline_dir=expected_output_folder, filename="test_inference.expected.png", remove_text=True
+    )
```
tab-cmd, line 197: linitng

bcipy/signal/tests/model/test_pca_rda_kde.py
```diff
@@ -195,5 +189,75 @@ def test_rda(self):
...
+        ax.set_xticks(np.arange(len(alp)))
+        ax.set_xticklabels(alp)
+        ax.set_yticks(np.arange(0, 101, 10))
+        return fig
+
+    def test_evaluate(self):
+        model_report = self.model.evaluate(self.x, self.y)
+        self.assertEqual(model_report, ModelEvaluationReport(1.0))
+
+    def test_save_load(self):
```
tab-cmd, line 224: Let's separate the load and save tests

tab-cmd (changes requested):
Great refactor! I added some comments/discussion points.

bci_main.py
nik-sm: I moved this to the same module due to an issue with circular imports. I'd be happy to move it back to `bcipy.helpers` but I think I might need some help to figure out how to prevent the circular import in that case

bcipy/signal/model/offline_analysis.py
nik-sm: I think this one also caused a circular import but let me double check

bcipy/signal/tests/model/test_pca_rda_kde.py
nik-sm: I wasn't sure how to actually test the saving/loading other than just checking "the model behavior is the same". The contents of the model itself don't seem to have a useful `__eq__`, even if you keep digging deeper until you get to the scipy library functions being used. Would be happy to change this but I think I could use some help on figuring out how to compare the models.

bcipy/helpers/visualization.py
nik-sm: I'd like to discuss this one to understand who is currently using this figure and how (and whether they would be OK with removing it). 

If you'd prefer a written description, I'd be happy to explain my thinking that way, but it might be easier to discuss in person. Just let me know which you'd prefer, and let's keep talking about this item

bcipy/signal/model/offline_analysis.py
nik-sm: (1) This works: `from bcipy.signal.model.pca_rda_kde import PcaRdaKdeModel` 
(2) This causes circular import error: `from bcipy.signal.model import PcaRdaKdeModel`.

Happy to change from `from .pca_rda_kde` to (1) if you prefer, or adjust some structure and go with (2)

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: Happy to follow if you have a style preference here. 

Just wanted to briefly explain my thinking about the module structure and why I used relative imports in a few places here.

In `bcipy/signal/model/__init__.py`, we'll expose the small subset of classes and functions that folks should interact with when importing from another module (like `bcipy/tasks/...` or something).

Within that module, the files can be agnostic about their absolute positioning, and if we use relative imports it's easy to move them to subfolders, etc.

Similar reasoning for `bcipy/signal/model/pca_rda_kde` as a module - the only exported items will basically be the `PcaRdaKdeModel`, and the other internal parts of that model, such as `ChannelWisePrincipalComponentAnalysis` should be encapsulated from the end user.

In fact, to keep following that convention, the tests for `PcaRdaKdeModel` would probably belong inside the module folder, since they _are_ inspecting these internal components - but I didn't want to make this change without discussion.

Anyway - happy to adopt whatever layout is consistent with the rest of the project and keeps these modules clean!

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: Hmm - it's true that the exact same statement occurs a number of times throughout the current code. I suppose it's a good idea to compute it in one place to prevent bugs, even if I can't really foresee changing it

nik-sm: Sigh. While refactoring this, I found that the bandwidth was actually calculated incorrectly previously. The current code preserves the way it was calculated, but it does not actually follow the "rule of thumb" that it was trying to.

https://en.wikipedia.org/wiki/Kernel_density_estimation#A_rule-of-thumb_bandwidth_estimator

The suggested rules of thumb are as follows, where:
```python
n_channels, n_items, n_samples = x.shape
```
(Note - this is a weird ordering. In pytorch/tensorflow/most numpy algorithms the batch dimension is typically kept first, and this is precisely the source of the current bug!)

Rule 1 from wikipedia:
```python
1.06 * np.std(x) * (n_items ** -0.2)
```

Rule 2 from wikipedia:
```python
0.9 * min(np.std(x), iqr(x) / 1.34) * (n_items ** -0.2)
```

Current logic (https://github.com/CAMBI-tech/BciPy/blob/main/bcipy/signal/model/mach_learning/train_model.py#L60):
```python
1.06 * min(np.std(x), iqr(x) / 1.34) * (n_channels ** -0.2)
```

The problems are:
1. It mixes the two rules of thumb (one with a multiplier of 1.06 and the other with 0.9). This is probably less important and wikipedia might have it wrong.
2. More important, it uses `n_channels` instead of `n_items`


As a takeaway, good suggestion, because it seems to have caught a bug, and DRY for the win once again.

bcipy/signal/model/offline_analysis.py
```diff
@@ -97,20 +98,20 @@ def offline_analysis(data_folder: str = None,
...
-    model.transform(x)
     fig_handles = generate_offline_analysis_screen(
         x, y, model=model, folder=data_folder,
         down_sample_rate=downsample_rate,
         fs=fs, save_figure=True, show_figure=False,
         channel_names=analysis_channel_names_by_pos(channels, channel_map))
 
     log.info('Saving the model!')
-    with open(data_folder + f'/model_{auc}.pkl', 'wb') as output:
+    with open(data_folder + f'/model_{model_performance.auc}.pkl', 'wb') as output:
```
lawhead, line 114: Should we round this to a certain number of decimal places? It seems like we only use the first few and these file names get to be pretty unwieldy. @sci-tab , thoughts?

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,113 @@
...
+        Parameters:
+            train_data: shape (Channels, Trials, Trial_length) preprocessed data
+            train_labels: shape (Trials,) binary labels
+
+        Returns:
+            trained likelihood model
+        """
+        pca = ChannelWisePrincipalComponentAnalysis(var_tol=1e-5, num_ch=train_data.shape[0])
+        rda = RegularizedDiscriminantAnalysis()
+        model = Pipeline()
```
lawhead: As a future improvement it might be nice for `Pipeline` to be able to handle *args: `model = Pipeline(pca, rda)`.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: Whoops - this was several half-completed comments from earlier. I removed and updated docstring. Let's also discuss updating the README for `bcipy/signal/model` or elsewhere to add more details about the model's behavior and how its update rule works. (It's not super complicated, and should be easy to explain pretty clearly and concisely, but at the moment I think it's not fully explained). I'll make a ticket.

bci_main.py
nik-sm: resolved after discussion

bcipy/signal/model/offline_analysis.py
nik-sm: Sounds like a good idea for a quick follow-up ticket

bcipy/helpers/visualization.py
nik-sm: We discussed and decided to:
1. remove for now, since the error cases covered by this figure are pretty well covered by tests now
2. come up with other debugging figures and metrics for future use, that work for any `SignalModel`

bcipy/signal/model/offline_analysis.py
nik-sm: going with (1)

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: fixed

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: added

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: After discussion, we decided to maintain the current behavior and investigate this further. https://www.pivotaltracker.com/story/show/178045750

bcipy/signal/model/offline_analysis.py
nik-sm: https://www.pivotaltracker.com/story/show/178045471

bcipy/signal/tests/model/test_pca_rda_kde.py
nik-sm: I'm gonna leave this for now unless you have ideas on how to test this.

Note that the previous test of loading was not testing whether a classifier is loaded correctly, but rather just whether a `parameters` dict was successfully pickled and unpickled, so the question of "How can we compare two model objects?" has not been previously answered as far as I can see.

I can't think of a simple way to compare these model objects (which are wrappers around relatively complicated Scikit-Learn objects) other than testing their behavior on some random inputs. Basically an equality comparison would have to go through each model component, identify all sklearn components inside, identify the parts of these which are modified during training and how to compare them. 

I think that may be the reason these objects like (https://github.com/scikit-learn/scikit-learn/blob/15a949460/sklearn/decomposition/_pca.py#L114 and https://github.com/scikit-learn/scikit-learn/blob/15a949460/sklearn/neighbors/_kde.py#L27) don't already have `__eq__` in sklearn.

bcipy/helpers/visualization.py
nik-sm: https://www.pivotaltracker.com/story/show/178045928

lawhead (approved):

bcipy/helpers/load.py
```diff
@@ -167,7 +166,17 @@ def load_experimental_data() -> str:
...
+def load_signal_model(model_class: SignalModel, model_kwargs: Dict[str, Any], filename: str = None):
+    """Construct the specified model and load pretrained parameters.
+
+    Args:
+        model_class (SignalModel, optional): Model class to construct.
+        model_kwargs (dict, optional): Keyword arguments for constructing model.
+        filename (str, optional): Location of pretrained model parameters.
+
+    Returns:
+        SignalModel: Model after loading pretrained parameters.
```
tab-cmd, line 182: It returns a tuple (SignalModel, str). Let's add a return type to the method signature too.

```
def load_signal_model(model_class: SignalModel, model_kwargs: Dict[str, Any], filename: str = None) -> Tuple[SignalModel, str]:
```

bcipy/signal/exceptions.py
```diff
@@ -0,0 +1,5 @@
+
+class SignalException(Exception):
```
tab-cmd, line 2: docstring

bcipy/signal/model/pca_rda_kde/density_estimation.py
```diff
@@ -1,12 +1,14 @@
...
 import numpy as np
+from scipy.stats import iqr
 
 
 class KernelDensityEstimate:
     """ Kernel density estimate implementation using scikit learn
     library. For further reference, please check scikit learn website.
     Attr:
         bandwidth(float): bandwidth of the kernel
+        scores(np.array): Shape (num_items, 2) - ratio of classification scores from RDA; used to compute bandwidth
```
tab-cmd, line 13: + num_channels

bcipy/signal/model/pca_rda_kde/density_estimation.py
```diff
@@ -35,6 +42,18 @@ def __init__(self, bandwidth=1.0, algorithm='auto', kernel='gaussian',
                                                    leaf_size=leaf_size,
                                                    metric_params=metric_params))
 
+    def _compute_bandwidth(self, scores, num_channels):
```
tab-cmd: + typing (no need to do the whole class but good to add with new stuff)

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,126 @@
...
+        Train on provided data using K-fold cross validation and return self.
+
+        Parameters:
+            train_data: shape (Channels, Trials, Trial_length) preprocessed data
+            train_labels: shape (Trials,) binary labels
+
+        Returns:
+            trained likelihood model
+        """
+        model = Pipeline([ChannelWisePrincipalComponentAnalysis(var_tol=1e-5, num_ch=train_data.shape[0]),
```
tab-cmd, line 32: We could pull up these hardcoded values to make it easier to change in the future. Not sure if it's worth it.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,126 @@
...
+                          RegularizedDiscriminantAnalysis()])
+
+        # Find the optimal gamma + lambda values
+        arg_cv = cross_validation(train_data, train_labels, model=model, k_folds=self.k_folds)
+
+        # Get the AUC using those optimized gamma + lambda
+        model.pipeline[1].lam = arg_cv[0]
+        model.pipeline[1].gam = arg_cv[1]
+        _, sc_cv, y_cv = cost_cross_validation_auc(
+            model, 1, train_data, train_labels, arg_cv, k_folds=self.k_folds, split="uniform"
```
tab-cmd: Same here.. what does one represent?

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -0,0 +1,126 @@
...
+from bcipy.signal.model.pca_rda_kde.classifier import RegularizedDiscriminantAnalysis
+from bcipy.signal.model.pca_rda_kde.cross_validation import cost_cross_validation_auc, cross_validation
+from bcipy.signal.model.pca_rda_kde.density_estimation import KernelDensityEstimate
+from bcipy.signal.model.pca_rda_kde.dimensionality_reduction import ChannelWisePrincipalComponentAnalysis
+from bcipy.signal.model.pca_rda_kde.pipeline import Pipeline
+from bcipy.signal.exceptions import SignalException
+
+
+class PcaRdaKdeModel(SignalModel):
+    def __init__(self, k_folds=10):
```
tab-cmd: We chatted briefly, but I'm in favor of removing these defaults. It can prevent a lot of headaches down the line.

bcipy/signal/tests/model/test_pca_rda_kde.py
tab-cmd: I wish we chatted in person but I was thinking to assert save and load behavior separately. You can mock the save / pkl and assert the necessary things were called and/or assert a file as expected was written (not that it loads). In another test you can mock a .pkl file and assert it did what it was supposed to on load. This will catch issues or changes in the individual methods. The test_save_load is a more integrated test of functionality where the comparison as you mention is important. I think your approach was fine. The fewer things you assert (ideally only one per test) and methods of interest you call per unittest the better as a rule of thumb. I won't block for this.

tab-cmd (approved):
A few cleanup items and suggestions. I don't need to re-review, so feel free to merge after cleanup!

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: I'd rather leave for now if that's OK

nik-sm: (just out of laziness)

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: Looks like it's the index for which element in the `Pipeline` is being optimized. In fact I don't think this parameter gets used, and it must always be a pipeline with PCA -> RDA, where the thing being optimized is RDA (index `1`). 

I'll add to a ticket to do some cleanup/minimization of this code as well, because it could probably be shortened without changing the logic.

PR #144, opened 2021-05-13
Model save fix

lawhead:
# Overview

Ensured that offline_analysis was calling the correct save method.

## Ticket

https://www.pivotaltracker.com/story/show/178148286

## Test

- Ran a calibration session through offline_analysis. Used the generated model to successfully run the Copy Phrase task.


nik-sm (approved):
Looks good

PR #145, opened 2021-05-18
Aziz PR v2

celikbasak:
# Overview

Answered comments from Aziz's previous PR https://github.com/CAMBI-tech/BciPy/pull/108. Refactored branch_test into modules.

## Ticket

- https://www.pivotaltracker.com/story/show/176444374

## Contributions

- Unanswered PR comments are resolved.
- Unused MomentumStimuliAgent class is removed.
- From previous branch_test, the scripts synthesized_user.py and copy_phrase_simulation.py are scrapped since they were unused.
- Cosmetic changes with autopep8

## Test

- Unit tests passed.

## Documentation

- In-line documentations added for RandomStimuliAgent and NBestStimuliAgent.

### DIFF
--- a/bcipy/tasks/rsvp/main_frame.py
+++ b/bcipy/tasks/rsvp/main_frame.py
@@ -160,7 +160,7 @@ class DecisionMaker:
         self.stopping_evaluator = stopping_evaluator
 
         # Stimuli Agent
-        self.stimuli_agent = RandomStimuliAgent(alphabet=self.alphabet)
+        self.stimuli_agent = stimuli_agent or RandomStimuliAgent(alphabet=self.alphabet)
 
         self.last_selection = ''
 
--- a/bcipy/tasks/rsvp/main_frame.py
+++ b/bcipy/tasks/rsvp/main_frame.py
@@ -140,9 +140,7 @@ class DecisionMaker:
                  stopping_evaluator=CriteriaEvaluator.default(min_num_inq=2,
                                                               max_num_inq=10,
                                                               threshold=0.8),
-                 stimuli_agent=RandomStimuliAgent(
-                     alphabet=list(string.ascii_uppercase) +
-                     [BACKSPACE_CHAR] + [SPACE_CHAR])):
+                 stimuli_agent=None):
         self.state = state
         self.displayed_state = self.form_display_state(state)
         self.stimuli_timing = stimuli_timing
@@ -162,7 +160,7 @@ class DecisionMaker:
         self.stopping_evaluator = stopping_evaluator
 
         # Stimuli Agent
-        self.stimuli_agent = stimuli_agent
+        self.stimuli_agent = RandomStimuliAgent(alphabet=self.alphabet)
 
         self.last_selection = ''
 
--- a/bcipy/tasks/rsvp/query_mechanisms.py
+++ b/bcipy/tasks/rsvp/query_mechanisms.py
@@ -58,7 +58,7 @@ class RandomStimuliAgent(StimuliAgent):
 
 
 class NBestStimuliAgent(StimuliAgent):
-    """ An inherited class of StimuliAgent. Updates the agent with N most likely 
+    """ An inherited class of StimuliAgent. Updates the agent with N most likely
         letters based on posteriors and selects queries.
         Attr:
             alphabet(list[str]): Query space(possible queries).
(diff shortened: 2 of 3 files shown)


bcipy/tasks/rsvp/main_frame.py
```diff
@@ -145,7 +143,7 @@ def __init__(self,
         self.stopping_evaluator = stopping_evaluator
 
         # Stimuli Agent
-        self.stimuli_agent = stimuli_agent
+        self.stimuli_agent = RandomStimuliAgent(alphabet=self.alphabet)
```
nik-sm: If the caller does `d = DecisionMaker(stimuli_agent=MyStimuliAgent())` then this will ignore it.

I think it should probably be something like:
```python
self.stimuli_agent = stimuli_agent if stimuli_agent is not None else RandomStimuliAgent(self.alphabet)
```

nik-sm: The test at https://github.com/CAMBI-tech/BciPy/blob/1.5.1/bcipy/helpers/tests/test_copy_phrase_wrapper.py#L161 fails due to this issue I believe

tab-cmd: We should probably be consistent with how we deal with these agents / evaluators. Either accept None and construct as you do for the agent, default as you do with the evaluator or remove the defaults and fail without it as an argument. I'm not a huge fan of all these defaults, but that's a discussion we can have outside this PR / in another ticket.

celikbasak: This change was a previous suggestion from @lawhead at https://github.com/CAMBI-tech/BciPy/pull/108. 
Aziz's PR version was:

`..., stimuli_agent=RandomStimuliAgent(                     alphabet=list(string.ascii_uppercase) +                              ['&lt;'] + [SPACE_CHAR])):`

Matt suggested that the `stimuli_agent` parameter should be given a default value of None and constructed in the init method if it's not provided, using the `alphabet` parameter. 

Therefore, the constructor is updated to `..., stimuli_agent=None):`
And in the attributes, it is redefined as:
`self.stimuli_agent = RandomStimuliAgent(alphabet=self.alphabet)`

nik-sm: So the original version that Aziz wrote set the default value of that argument to a `RandomStimuliAgent(...)`.
We still want the default to be this way, but if the user provides a different agent, we should use that instead.

What you said here is the key point:
> and constructed in the init method **if it's not provided**

Notice that in the current code, the stimuli agent is ALWAYS created as a `RandomStimuliAgent(...)`, even if the user tries to construct something else.
So one way or another, you'll have to inspect the user input and decide whether to create that default random agent, or to use the provided one.

There are 2 good reasons to construct this default `RandomStimuliAgent(...)` inside the body of the function, and not have it as part of the `def` line.
1. If a mutable object is used as a default in the definition of a function, then _all class instances_ will share the exact same object. They won't have different copies, they'll literally have a pointer to the same object. For example:
```python
class Trouble:
    # Notice the default empty list here, a mutable object that will unfortunately be shared between instances!
    def __init__(self, contents=[]):  
        self.contents = contents

breakfast = Trouble()
dinner = Trouble()
print(breakfast.contents)  # empty at first
dinner.contents.append("escargot")
print(breakfast.contents) # Ew! escargot for breakfast!
```
2. The object being constructed (the agent) has some parameters (alphabet) that were hard-coded, but actually should be kept in sync with the other arguments being passed. The only way to achieve this is by constructing it in the body of the function.

nik-sm: looking back at Matt's suggestion, he has an even more slick way to do this inline (https://github.com/CAMBI-tech/BciPy/pull/108#discussion_r539775439):
```python
self.stimuli_agent = stimuli_agent or RandomStimuliAgent(...)
```
This achieves the same effect.

P.S. - The current code also causes the tests to fail - have you run the unit tests with `pytest --mpl -k "not slow"`?

celikbasak: @nik-sm Thank you for your explanation! On last week's sprint planning Matt explained the OR usage:
```self.stimuli_agent = stimuli_agent or RandomStimuliAgent(alphabet=self.alphabet)```
That was my bad that I couldn't understand his suggestion. Now I implemented it.

Which errors did you get during the tests? I see couple of `ModuleNotFoundError` while running `pytest --mpl -k "not slow"`. Tests passed for the decision maker: `coverage run --branch --source=bcipy/tasks/rsvp -m pytest -k TestDecisionMaker`.

tab-cmd (approved):

PR #146, opened 2021-05-21
Copy phrase refactor

lawhead:
# Overview

Refactored Copy Phrase Task. Integrated inquiry preview evidence.

## Ticket

- https://www.pivotaltracker.com/story/show/177802363
- https://www.pivotaltracker.com/story/show/177802396

## Contributions

- Added unit tests for Copy Phrase Task.
- Refactored Copy Phrase task to increase clarity and flexibility.
- Bug fixes in session data handling when using fake data.
- Refactored session data to provide more flexibility to provide additional evidence types.
- In Copy Phrase Wrapper, factored out the evidence calculations from the decision-making; added type information and documentation.
- Added named tuples for complex return types that were used throughout the code.
- Integrated button press evidence for Inquiry Preview functionality.
 
## Test

- Run all unit tests, including the new ones added for the Copy Phrase task.
- Run the Copy Phrase task:
    - With and without inquiry preview
    - With fake data
    - With real data (mock server)


bcipy/acquisition/README.md
```diff
@@ -116,43 +116,3 @@ A `Protocol` contains device-specific behavior for how to generate mock data. It
...
-```
-
-Specify parameters. If your sampling frequency is different than predefined values, specify the filter. Returned value is another numpy array in the form:
-
-```python
-output_seq = np.array([[.3, .4, ...],
-       	                [.2, .1, ...]])
-```
-
-For other details, refer to demo file or function definition.
```
tab-cmd, line 158: Thanks!

bcipy/helpers/stimuli.py
```diff
@@ -16,13 +15,27 @@
 logging.getLogger("PIL").setLevel(logging.WARNING)
 
 
+class InquirySchedule(NamedTuple):
+    """Schedule for the next inquiries to present, where each inquiry specifies
+    the stimulus, duration, and color information.
+
+    Attributes
+    ----------
+    - stims: `List[List[str]]`
```
tab-cmd: stimuli

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -99,16 +105,47 @@ def __init__(self, min_num_inq, max_num_inq, signal_model=None, fs=300, k=2,
...
-
-        Args:
-            raw_data(ndarray[float]): C x L eeg data where C is number of
-                channels and L is the signal length
-            triggers(list[tuple(str,float)]): triggers e.g. ('A', 1)
-                as letter and flash time for the letter
-            target_info(list[str]): target information about the stimuli
-            window_length(int): The length of the time between stimuli presentation
+    def evaluate_inquiry(
+            self, raw_data, triggers, target_info, window_length
```
tab-cmd: add typing

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -1,18 +1,24 @@
...
+from bcipy.tasks.rsvp.stopping_criteria import (CriteriaEvaluator,
+                                                MaxIterationsCriteria,
+                                                MinIterationsCriteria,
+                                                ProbThresholdCriteria)
 
+_EV_NAMES = ['LM', 'ERP', 'BTN']
+EvidenceName = namedtuple('Enum', _EV_NAMES)(*_EV_NAMES)
 
 class CopyPhraseWrapper:
     """Basic copy phrase task duty cycle wrapper.
```
tab-cmd: This could use better documentation. I don't even know the parameter list is necessary, but what it does and how it operates at a high level would be helpful.

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -134,26 +171,53 @@ def evaluate_inquiry(self, raw_data, triggers, target_info, window_length):
                                     channel_map=self.channel_map,
                                     trial_length=window_length)
 
-        lik_r = self.signal_model.predict(x, letters, self.alp)
-        prob = self.conjugator.update_and_fuse({'ERP': lik_r})
-        decision, sti = self.decision_maker.decide(prob)
+        return self.signal_model.predict(x, letters, self.alp)
 
-        return decision, sti
+    def add_evidence(self, name: str, evidence: List[float]) -> np.array:
```
tab-cmd: Could this use the new `EvidenceName`?

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -1,18 +1,24 @@
...
-    MinIterationsCriteria, MaxIterationsCriteria, ProbThresholdCriteria
-from bcipy.helpers.language_model import norm_domain, sym_appended, \
-    equally_probable, histogram
+from bcipy.tasks.rsvp.stopping_criteria import (CriteriaEvaluator,
+                                                MaxIterationsCriteria,
+                                                MinIterationsCriteria,
+                                                ProbThresholdCriteria)
 
+_EV_NAMES = ['LM', 'ERP', 'BTN']
+EvidenceName = namedtuple('Enum', _EV_NAMES)(*_EV_NAMES)
```
tab-cmd: We might want to pull this out to use elsewhere. Not sure where to put it yet, but it's helpful.

bcipy/helpers/stimuli.py
```diff
@@ -174,13 +187,15 @@ def best_case_rsvp_inq_gen(alp: list,
         # append colors
         colors.append([color[i] for i in range(len(color) - 1)] +
                       [color[-1]] * stim_length)
-    return (samples, times, colors)
+    return InquirySchedule(samples, times, colors)
 
 
-def random_rsvp_calibration_inq_gen(alp, timing=[0.5, 1, 0.2],
+def random_rsvp_calibration_inq_gen(alp,
```
tab-cmd, line 193: What do you think of moving these into `task/rsvp` with the decisionmaker etc... It doesn't need to happen in this PR, but thinking of where best to put  them. I'm thinking `helpers/` module should only hold core items or connections between modules, these are very task-specific.

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
...
+        """
+        if self.parameters['show_feedback']:
+            feedback = VisualFeedback(self.window, self.parameters,
+                                      self.experiment_clock)
+            feedback.administer(
+                selection,
+                message='Selected:',
+                fill_color=self.parameters['feedback_message_color'])
+
+    def stoppage_criteria_ok(self) -> bool:
```
tab-cmd: `check_stop_criteria`?

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
...
+    def write_trigger_data(self, stim_times: List[Tuple[str, float]],
+                           trigger_file: TextIO) -> None:
+        """Save trigger data to disk.
+
+        Parameters
+        ----------
+        - stim_times : list of (stim, clock_time) tuples
+        - trigger_file : data will be appended to this file
+        """
+        # TODO: fix the helper to correctly handle the targetness of inquiry_preview stims.
```
tab-cmd, line 606: What does it write currently?

bcipy/tasks/session_data.py
```diff
@@ -19,6 +19,7 @@ class Inquiry:
         eeg_evidence - eeg evidence for each stimulus
         likelihood - combined likelihood for each stimulus
     """
+    EVIDENCE_SUFFIX = '_evidence'
```
tab-cmd: Maybe evidence names could live here too? We can use its name to get the values below.

tab-cmd (approved):
Huge effort and it looks much cleaner! Thank you 🎊

bcipy/tasks/rsvp/copy_phrase.py
lawhead: I created a ticket for this with example output: https://www.pivotaltracker.com/story/show/178253198

bcipy/helpers/stimuli.py
lawhead: I like that suggestion

bcipy/helpers/copy_phrase_wrapper.py
lawhead: It's not a true Enum, but maybe I should refactor so it is to be more explicit about types.

bcipy/helpers/copy_phrase_wrapper.py
lawhead: Agreed. I didn't look at the class-level documentation. I'll see if I can improve this.

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
     """
 
     TASK_NAME = 'RSVP Copy Phrase Task'
-
-    def __init__(
-            self, win, daq, parameters, file_save, signal_model, language_model, fake):
+    PARAMETERS_USED = [
```
nik-sm, line 74: What's this list for?

nik-sm: I'm guessing you're thinking of doing things like:
```python
t = Task(...)
...
for p in t.PARAMETERS_USED:
    add_display_widget(p)
```

But it looks like this is currently unused (or am I missing something?)

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
...
+
+    def init_copy_phrase_task(self) -> CopyPhraseWrapper:
+        """Initialize the CopyPhraseWrapper.
+
+        Returns:
+        --------
+        initialized CopyPhraseWrapper
+        """
+
+        self.copy_phrase_task = _init_copy_phrase_wrapper(
```
nik-sm, line 168: At this point, since so many parameters are extracted and passed in, is it easier to just pass around the `parameters` dict and functions can grab contents directly?

If this adds concerns about parameters being accidentally changed inside functions, it could be changed to some immutable data structure. (For example, maybe something like: https://pypi.org/project/frozendict/. Also dataclasses with `frozen=True`  would probably be suitable for python3.7+)

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
...
+
+    def next_target(self) -> str:
+        """Computes the next target letter based on the currently spelled_text.
+        """
+        if self.copy_phrase[0:len(self.spelled_text)] == self.spelled_text:
+            # if correctly spelled so far, get the next letter.
+            return self.copy_phrase[len(self.spelled_text)]
+        return BACKSPACE_CHAR
+
+    def execute(self) -> str:
```
nik-sm, line 321: This is a nice reorganization 👍 
It'll probably take me a while to step through all the logic, but the high level process is much more clear.

bcipy/tasks/rsvp/copy_phrase.py
```diff
@@ -44,368 +71,610 @@ class RSVPCopyPhraseTask(Task):
...
+        --------
+        - self.copy_phrase_task
+        """
+        evidences = [
+            self.compute_button_press_evidence(proceed),
+            self.compute_eeg_evidence(stim_times, proceed)
+        ]
+        evidence_types = []
+        for evidence in evidences:
+            if evidence:
```
nik-sm, line 424: I like this approach - all the evidences can be included and simply set to `None` if not desired. 
It would be good to have tests that check the alphabet distribution in situations like (have inquiry preview, don't have EEG), (no preview, have EEG), (have preview, have EEG) - and see that the changes to letter probabilities are as expected.
This might already be covered by some simple integration tests of copy_phrase_wrapper or of this task - but just wanted to mention it here while I think of it.

nik-sm: In particular, I mention it because the test might have to do some "under-the-hood" inspection (like checking `task.copy_phrase_task.symbol_probabilities` (or whatever the correct attribute is, rather than `symbol_probabilities`).

On the other hand, this would feel a bit strange (tests of `copy_phrase` probably should not know/care about its attributes like `self.copy_phrase_task` too much).

bcipy/tasks/session_data.py
```diff
@@ -94,13 +135,16 @@ def stim_evidence(self, alphabet: List[str], n_most_likely: int = 5) -> Dict:
...
-        return {
+        data = {
             'stimuli': self.stimuli,
-            'lm_evidence': dict(zip(alphabet, self.lm_evidence)),
-            'eeg_evidence': dict(zip(alphabet, self.eeg_evidence)),
-            'likelihood': likelihood,
-            'most_likely': dict(Counter(likelihood).most_common(n_most_likely))
         }
+        for name, evidence in self.evidences.items():
+            data[Inquiry.serialized_name(name)] = dict(zip(alphabet, evidence))
```
nik-sm: I don't understand the reason for the indirection here - why not use `data[name]`?

nik-sm (approved):
Looks good overall! Some details on task and session that I'm still not clear on, but I'll just need to keep familiarizing myself with that part of the code.

bcipy/tasks/session_data.py
lawhead: This is mainly for backwards compatibility. In the session data we have called it `eeg_evidence` and in the Copy Phrase Wrapper it's referred to as `ERP`.

bcipy/tasks/rsvp/copy_phrase.py
lawhead: Agreed. We're inconsistent in the code as to when we pass all parameters vs when we pass parameter values as arguments to other objects. For this first pass at least I wanted to be clear when the argument value was just a straight parameter value vs a computed value. I think we should take a second pass at this in another ticket where we decide whether or not to pass the whole parameter dict around and let the consuming object take what it wants, or we have another way of organizing these similar to what Tab did regarding the display code using named tuples for a subset of parameters. 

The parameters object looks like a dict but it's actually a subclass (defined in parameters.json), so we can easily enforce immutability there. We have been treating it as an immutable object but haven't enforced that.

bcipy/tasks/rsvp/copy_phrase.py
lawhead: Good call out. Because so many parameters are used for initializing the copy phrase wrapper and the display it took a while to figure out what parameters were needed for my unit tests. I intended to do a quick assertion here that all required parameters were provided but forgot to include it. I'll add that code.

bcipy/tasks/rsvp/copy_phrase.py
lawhead: Yeah, I think the tests for the Copy Phrase task should just confirm that the add_evidence method was called with the expected types, and the CopyPhraseWrapper should assert that the shape of the provided evidence is correct. Whenever possible I tried to limit the level of knowledge that the Copy Phrase task has about the implementation details of the wrapper.

nik-sm: Ok - sounds like a reasonable way to divide the unit tests

PR #147, opened 2021-05-25
Refactor trial inquiry reshaper

nik-sm:
# Overview

Add `data_reshaper` as entrypoint for `trial_reshaper` and `inquiry_reshaper`.

Refactor `trial_reshaper`.

Restructure preprocessing transforms (notch, bandpass, etc) so that they can optionally take a sampling rate and return the sampling rate of the transformed data.

Removed unused text filter.

Change model PKL file to use 4 digits of precision.

## Ticket

https://www.pivotaltracker.com/story/show/177823118
https://www.pivotaltracker.com/story/show/175322114
https://www.pivotaltracker.com/story/show/178045471

## Contributions

- `bcipy/helpers/task.py` - change `trial_reshaper` to remove unused outputs and simplify inputs and function structure. Add `inquiry_reshaper` and `data_reshaper`. Entrypoint is now `data_reshaper`, which requires `input_data_type` argument.

- Move all transforms to `bcipy/signal/process/transforms.py`. Transforms can accept optional `fs`, returning the resulting `fs` after modifying the data. `get_default_transform(...)` describes the standard Notch, Bandpass, Downsample pipeline.
- Remove helper functions from `bcipy/gui/viewer/data_source/filter.py` - these are now accomplished directly with `transforms.py`.
- Remove unused "text filter" from `bcipy/signal/process`, `bcipy/acquisition/README.md`, `bcipy/signal/tests/process/`
- `bcipy/helpers/copy_phrase_wrapper.py` - use new transforms and trial_reshaper

## Test

Ran all tests with `pytest --mpl --disable-warnings`.
Added and updated tests for `bcipy/helpers/task.py`.

## Documentation

- Added docstrings
- Removed unused readme describing text filter


bcipy/helpers/stimuli.py
```diff
@@ -30,6 +30,8 @@ class InquirySchedule(NamedTuple):
     colors: List[List[str]]
 
 # TODO: since we have a querying this should replace the other generators
```
tab-cmd: I'm not sure I understand this TODO. Do you? Is this something we could move to a ticket?

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+    """
+    # Remove the channels that we are not interested in
+    channels_to_remove = [idx for idx, value in enumerate(channel_map) if value == 0]
+    eeg_data = np.delete(eeg_data, channels_to_remove, axis=0)
+
+    # Number of samples we are interested per inquiry
+    num_samples = int(trial_length * fs * trials_per_inquiry)
+
+    # Remove unwanted elements from target info and timing info
+    targets_included = set(["target", "nontarget"])
```
tab-cmd: We should ask for the triggers we want to be included/excluded. This could be used for label assignments as well.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+
+
+def grouper(iterable, n, fillvalue=None):
+    args = [iter(iterable)] * n
+    return zip_longest(*args, fillvalue=fillvalue)
+
+
+def inquiry_reshaper(trial_target_info: list,
+                     timing_info: list,
+                     eeg_data: np.ndarray,
```
tab-cmd: It may be helpful for us to make a Data model which contains some of these repeated arguments that are intrinsically related (ex. data, fs)

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+                     trials_per_inquiry: int,
+                     offset: float = 0,
+                     channel_map: List[int] = DEFAULT_CHANNEL_MAP,
+                     trial_length: float = 0.5):
+    """Extract inquiry data and labels.
+    Args:
+        trial_target_info (list): labels each trial as "target", "non-target", "first_pres_target", etc
+        timing_info (list): Timestamp of each event in seconds
+        eeg_data (np.ndarray): shape (channels, samples) preprocessed EEG data
+        fs (int): sample rate of preprocessed EEG data
```
tab-cmd: It does not have to be preprocessed

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+        trials_per_inquiry (int): number of trials in each inquiry
+        offset (float, optional): Any calculated or hypothesized offsets in timings. Defaults to 0.
+        channel_map (tuple, optional): Describes which channels to include or discard. Defaults to DEFAULT_CHANNEL_MAP.
```
tab-cmd: Updated to list

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+                   channel_map: List[int] = DEFAULT_CHANNEL_MAP,
+                   trial_length: float = 0.5) -> Tuple[np.ndarray, np.ndarray]:
+    """Extract trial data and labels.
+
+    Args:
+        trial_target_info (list): labels each trial as "target", "non-target", "first_pres_target", etc
+        timing_info (list): Timestamp of each event in seconds
+        eeg_data (np.ndarray): shape (channels, samples) preprocessed EEG data
+        fs (int): sample rate of preprocessed EEG data
+        trials_per_inquiry (int, optional): unused, kept here for consistent interface with `inquiry_reshaper`
```
tab-cmd: probably not necessary... these aren't children of a reshaper class and so don't need the same contract.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+        offset (float, optional): Any calculated or hypothesized offsets in timings. Defaults to 0.
+        channel_map (tuple, optional): Describes which channels to include or discard. Defaults to DEFAULT_CHANNEL_MAP.
+        trial_length (float, optional): [description]. Defaults to 0.5.
+
+    Returns:
+        trial_data (np.ndarray): shape (channels, trials, samples) reshaped data
+        labels (np.ndarray): integer label for each trial
+
+    TODO ??? In all modes > calibration, we assume the timing info is given in samples not seconds.
```
tab-cmd: That assumption is not correct. We only use this on load in all data is stored in the same fashion.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,153 @@ def get_key_press(
...
+    eeg_data = np.delete(eeg_data, channels_to_remove, axis=0)
 
-    except Exception as e:
-        raise Exception(
-            f'Could not reshape trial for mode: {mode}, {fs}, {k}. Error: {e}')
+    # Number of samples we are interested per trial
+    num_samples = int(trial_length * fs)
+
+    # Remove unwanted elements from target info and timing info
+    targets_included = set(["target", "nontarget"])
```
tab-cmd: I think the same here as above in regards to accepting it as an argument. Ask for a list of things to include and potentially a targetness argument, which could be used ln 516.

bcipy/signal/model/base_model.py
```diff
@@ -1,11 +1,23 @@
...
 
 
+class InputDataType(Enum):
+    """Controls how data is reshaped before being passed to a model"""
+    TRIAL = 1
+    INQUIRY = 2
+
+
 class SignalModel(ABC):
+    @property
```
tab-cmd, line 9: 👍

bcipy/helpers/tests/test_task.py
```diff
@@ -51,12 +53,67 @@ def test_alphabet_images(self):
...
+        self.eeg = np.random.randn(self.n_channel, self.fs * 10)
+        self.channel_map = [1] * self.n_channel
+
+    def test_data_reshaper_vs_trial_reshaper(self):
+        reshaped_trials_1, reshaped_labels_1 = data_reshaper(
+            input_data_type=InputDataType.TRIAL,
+            trial_target_info=self.target_info,
+            timing_info=self.timing_info,
+            eeg_data=self.eeg,
+            fs=256,
```
tab-cmd: use self.fs?

bcipy/helpers/tests/test_task.py
```diff
@@ -51,12 +53,67 @@ def test_alphabet_images(self):
...
+    """Checks that `data_reshaper` is a transparent wrapper of `trial_reshaper` and `inquiry_reshaper`"""
+
+    def setUp(self):
+        self.target_info = ['first_pres_target', 'fixation',
+                            'target', 'nontarget', 'nontarget']
+        self.timing_info = [1.001, 1.2001, 1.4001, 1.6001, 1.8001]
+        self.n_channel = 7
+        self.fs = 256
+        self.trial_per_inquiry = 3
+        self.eeg = np.random.randn(self.n_channel, self.fs * 10)
```
tab-cmd: It looks like you create eeg data for this file a few times. It may be worth pulling it up into a test method and calling in the unittests

bcipy/helpers/tests/test_task.py
```diff
@@ -65,29 +122,55 @@ def setUp(self):
...
-        self.assertEqual(num_of_inq, 1)
-        self.assertEqual(trials_per_inquiry, 3)
 
-    def test_trial_reshaper_copy_phrase(self):
-        pass
 
+class TestInquiryReshaper(unittest.TestCase):
+    def setUp(self):
+        self.n_channel = 7
+        self.trial_length = 0.5  # seconds
```
tab-cmd: remove comment

bcipy/helpers/tests/test_task.py
```diff
@@ -65,29 +122,55 @@ def setUp(self):
...
+        reshaped_data, labels = inquiry_reshaper(
+            trial_target_info=self.target_info,
+            timing_info=self.timing_info,
+            eeg_data=self.eeg,
+            fs=self.fs,
+            trials_per_inquiry=self.trials_per_inquiry,
+            channel_map=self.channel_map,
+            trial_length=self.trial_length,
+        )
+        expected_shape = (self.n_channel, 3, int(self.trial_length * self.fs) * self.trials_per_inquiry)
```
tab-cmd: self.trials_per_inquiry?

bcipy/signal/model/offline_analysis.py
```diff
@@ -92,13 +92,16 @@ def offline_analysis(data_folder: str = None,
...
-                                channel_map=channel_map,
-                                trial_length=trial_length)
-
     model = PcaRdaKdeModel(k_folds=k_folds)
+    x, y = data_reshaper(input_data_type=model.input_data_type,
+                         trial_target_info=t_t_i,
+                         timing_info=t_i,
+                         eeg_data=data,
+                         fs=fs,
+                         trials_per_inquiry=parameters.get('stim_length', 10),
```
tab-cmd: I don't think we should use defaults in offline analysis if we can avoid it (especially for important things like stim_length). I'd rather we fail.

bcipy/signal/process/transforms.py
```diff
@@ -0,0 +1,74 @@
+from scipy.signal import butter, filtfilt, iirnotch, sosfilt
```
tab-cmd: I like the composition and get_default here. Perhaps renaming to `transform`. I don't think we should collapse the filter module. We should get those concepts separate and will add things to them independently imo.

bcipy/signal/process/transforms.py
```diff
@@ -0,0 +1,74 @@
...
+import numpy as np
+from typing import Optional
+
+
+class Composition:
+    def __init__(self, *transforms):
+        self.transforms = transforms
+
+    def __call__(self, data: np.ndarray, fs: Optional[int] = None):
+        for t in self.transforms:
```
tab-cmd: transform in self.transforms

bcipy/signal/process/transforms.py
tab-cmd: These need docstrings and return types

bcipy/signal/process/transforms.py
```diff
@@ -0,0 +1,74 @@
...
+    def __init__(self, factor: int = 2):
+        self.factor = factor
+
+    def __call__(self, data: np.ndarray, fs: Optional[int] = None):
+        if fs is None:
+            return data[:, :: self.factor]
+        return data[:, :: self.factor], fs // self.factor
+
+
+class Lambda:
```
tab-cmd: Is there a more specific label we can give f? 

In this case, should f return fs? What if it transformed it in some way that impacts it?

bcipy/tasks/rsvp/calibration/inter_inquiry_feedback_calibration.py
```diff
@@ -297,21 +298,23 @@ def _get_data_for_psd(self, inquiry_timing):
...
 
         # reshape with the filtered data with our desired window length
-        reshaped_data, _, _, _ = trial_reshaper(
-            target_info,
-            times,
-            data,
-            fs=self.fs,
-            k=self.downsample_rate, mode='calibration',
+        reshaped_data, _ = data_reshaper(
+            input_data_type=InputDataType.TRIAL,
```
tab-cmd: Let's put this in the constructor. self.input_data_type

tab-cmd (changes requested):

bcipy/helpers/task.py
nik-sm: True - I intended to convey "provide the actual sample rate, rather than the original sample rate". I'll clarify the doc

bcipy/signal/process/transforms.py
nik-sm: Will add doc. So you wanna have `bcipy/signal/process/transforms.py` -> `bcipy/signal/process/filter/transform.py` right?

bcipy/signal/process/transforms.py
nik-sm: Hmm good point - maybe just leave out this transform for now, and add it back when we have a use case in mind?

(In general though probably `f` should always operate on both `data` and `fs`, so that the user can keep these two items in sync)

bcipy/helpers/stimuli.py
nik-sm: I don't understand it. I'll remove for now, and we can make a ticket if you want

bcipy/signal/process/transforms.py
tab-cmd: I'm suggesting that transforms can stay where they are. The filters themselves (bandpass, notch, downsample) would move to `bcipy/signal/process/filter.py`

bcipy/helpers/task.py
nik-sm: Ok - I added these with the defaults being `included = set(["target", "nontarget"])` and `excluded = set([])`.

Since these sets are not mutated within the function, I think it's OK (and keeps the code simple) to just construct them in the function args.

bcipy/helpers/task.py
nik-sm: Ok - happy to include that here if you know what things should be grouped.
If it requires some thinking/design, perhaps this should be a follow-up ticket?

bcipy/helpers/task.py
nik-sm: What I imagined was that during `offline_analysis.py`, when this function gets called, it'll be annoying to construct two different sets of function arguments. Certainly doable, but it seemed simpler to accept that extra arg and discard it, than to decide what to pass in.

Would you prefer adjusting the set of arguments used, or discarding the extra arg for `inquiry` situation?

bcipy/helpers/task.py
nik-sm: Thanks for catching this. This statement was in the previous doc and I couldn't understand what it meant so I left a TODO. I'll delete it.

bcipy/signal/process/transforms.py
nik-sm: I'm gonna make the return type simpler. 
Currently it would be: `-> Union[np.ndarray, Tuple[np.ndarray, int]]`, but instead I'll just always return something for `fs` - either the updated value, or `None` if `None` was provided.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,159 @@ def get_key_press(
...
 
 
+def data_reshaper(input_data_type: InputDataType, *args, **kwargs):
+    if input_data_type == InputDataType.TRIAL:
+        return trial_reshaper(*args, **kwargs)
+    elif input_data_type == InputDataType.INQUIRY:
+        return inquiry_reshaper(*args, **kwargs)
+    else:
+        raise NotImplementedError(f"Unsupported input data type: {input_data_type}")
+
```
lawhead, line 391: Rather than dispatching to a function based on an enum value, it seems like this would be well-suited to an object-oriented design using something like a`Reshaper` parent class for the interface with `TrialReshaper` and `InquiryReshaper` subclasses. These could each override a `reshape` method or you could use the `__call__` method like you did for the filters. Using classes would also allow you to use more meaningful type hints.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,159 @@ def get_key_press(
...
-                        'B' : [39.60, -224.124, ...],
-                        'T' : [...],
-                        ...
-                    },
-                    ...
-                }
+                   channel_map: List[int] = DEFAULT_CHANNEL_MAP,
+                   trial_length: float = 0.5,
+                   targets_included: Set[str] = set(["target", "nontarget"]),
+                   targets_excluded: Set[str] = set([])) -> Tuple[np.ndarray, np.ndarray]:
```
lawhead: Not necessarily for this PR, but we should probably have an enum of target types so we don't have to worry about mis-typing strings (like 'non-target').

bcipy/helpers/task.py
nik-sm: Would be happy to restructure, but I'm not sure I quite understand - for example in `copy_phrase_wrapper`, here's what I'm imagining based on what you described.
Instead of calling a function  `data_reshaper`, it could have something like:

```python
if ...:
    reshaper = TrialReshaper(...)
else:
    reshaper = InquiryReshaper(...)

x, y = reshaper(data)
```

and in `bcipy/helpers/task.py`, something like:

```python
class Reshaper(ABC):
    @abstractmethod
    def __call__(self):
        ...

class TrialReshaper(Reshaper):
    def __call__(self, trial_target_info, timing_info, eeg_data):
        # implementation
```

The issue is this still basically seems to rely on that `if/else` statement happening somewhere.

Another approach might be to store this on the signal model itself, but I think we had decided against this 
(I think the rationale was that we don't necessarily have a signal model every time we want to reshape data, for example in `bcipy/tasks/rsvp/calibration/inter_inquiry_feedback_calibration.py`)

bcipy/helpers/task.py
nik-sm: Here's a ticket about it for now - let me know if you prefer to address it in this PR - https://www.pivotaltracker.com/story/show/178294216

bcipy/signal/process/transforms.py
nik-sm: Sorry for the confusion - I think I'm not quite picturing what you mean.
So far, I converted the filters into pytorch-style transforms - are you suggesting to take the implementation of those transforms into a separate file?

So that would be something like:

---

In `filter.py`:
```python
def downsample(data, fs, factor):
    return data[:, :: factor], fs // factor
```
In `transform.py`:
```python
class Downsample:
    ...
    def __call__(self, data, fs):
        downsample(data, fs, self.factor)
````

File: bcipy/helpers/task.py (hunk 20 above)
Comment (nik-sm) [reply bcipy/helpers/task.py]: Good idea. Looks like you can also make a printable python enum pretty easily:
```python
class TargetType(Enum):
    TARGET = "target"
    NONTARGET = "nontarget"

print(TargetType.NONTARGET.value)
# 'nontarget'
```

Here's a ticket to follow-up on this: https://www.pivotaltracker.com/story/show/178294298

bcipy/helpers/tests/test_task.py
nik-sm: I put this into a follow-up ticket: https://www.pivotaltracker.com/story/show/178294216

nik-sm: Ok - I put this into a follow-up ticket: https://www.pivotaltracker.com/story/show/178302572

bcipy/signal/process/transforms.py
tab-cmd: In filter.py:
```python
class Bandpass:    
    def __init__(self, lo, hi, sample_rate_hz, order=5):
      nyq = 0.5 * sample_rate_hz       
      lo, hi = lo / nyq, hi / nyq 
      self.sos = butter(order, [lo, hi], analog=False, btype="band", output="sos")
    def __call__(self, data: np.ndarray, fs: Optional[int] = None): 
       if fs is None:
            return sosfilt(self.sos, data)
      return sosfilt(self.sos, data), fs

class Notch:
    ...

```
In transform.py:

```python
from bcipy.signal.process.filter import Bandpass, Notch

class Composition:    
    def __init__(self, *transforms):
        self.transforms = transforms
    def __call__(self, data: np.ndarray, fs: Optional[int] = None):
      for t in self.transforms:
         data, fs = t(data, fs)
         if fs is None:
            return data       
         return data, fs

def get_default_transform(...):
```

Downsample could probably stay in transform now that I think of it more, but I could be persuaded either way.

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -177,27 +183,24 @@ def evaluate_eeg_evidence(self, raw_data: np.array,
...
+            sample_rate_hz=self.sampling_rate,
+            notch_freq_hz=self.notch_filter_frequency,
+            bandpass_low=self.filter_low,
+            bandpass_high=self.filter_high,
+            bandpass_order=self.filter_order,
+            downsample_factor=self.downsample_rate,
+        )
+        data, self.sampling_rate = default_transform(raw_data, self.sampling_rate)
+
+        x, _ = data_reshaper(input_data_type=self.signal_model.input_data_type,
```
tab-cmd: small change... let's assign a better label to this? `trials` or something equally descriptive works for me

bcipy/helpers/task.py
tab-cmd: I see what you're saying. I'd be okay leaving this.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,159 @@ def get_key_press(
...
+    Args:
+        trial_target_info (list): labels each trial as "target", "non-target", "first_pres_target", etc
+        timing_info (list): Timestamp of each event in seconds
+        eeg_data (np.ndarray): shape (channels, samples) preprocessed EEG data
+        fs (int): sample rate of preprocessed EEG data
+        trials_per_inquiry (int, optional): unused, kept here for consistent interface with `inquiry_reshaper`
+        offset (float, optional): Any calculated or hypothesized offsets in timings. Defaults to 0.
+        channel_map (tuple, optional): Describes which channels to include or discard. Defaults to DEFAULT_CHANNEL_MAP.
+        trial_length (float, optional): [description]. Defaults to 0.5.
+        targets_included (Set[str]): target labels to include
```
tab-cmd: labels would be better now that I think of it more. Not necessary to adjust here unless you agree passionately, there a few places we use target including `trial_target_info` where we mean label.

bcipy/helpers/task.py
```diff
@@ -370,186 +373,159 @@ def get_key_press(
...
+    eeg_data = np.delete(eeg_data, channels_to_remove, axis=0)
 
-    except Exception as e:
-        raise Exception(
-            f'Could not reshape trial for mode: {mode}, {fs}, {k}. Error: {e}')
+    # Number of samples we are interested per trial
+    num_samples = int(trial_length * fs)
+
+    # Remove unwanted elements from target info and timing info
+    targets_included = set(["target", "nontarget"])
```
tab-cmd: You added it as an argument, but it's not implemented in trial_reshaper yet

bcipy/helpers/task.py
```diff
@@ -370,186 +373,159 @@ def get_key_press(
...
+
+    for inquiry_idx, chunk in enumerate(grouper(zip(trial_target_info, triggers), trials_per_inquiry)):
+        # label is the index of the "target", or else the length of the inquiry
+        label = trials_per_inquiry
+        first_trigger = None
+        for trial_idx, (target, trigger) in enumerate(chunk):
+            if first_trigger is None:
+                first_trigger = trigger
+
+            if target == 'target':
```
tab-cmd: What I had in mind was defining labels to include (target, nontarget) & exclude (fixation), and then another place to define a target label as you hardcode here (optional, since this won't be available when we reshape a free spell task for instance). Then you could reshape various labels and assign targetness without hardcoding your assumption about what that label may be.

bcipy/signal/model/offline_analysis.py
```diff
@@ -92,13 +92,16 @@ def offline_analysis(data_folder: str = None,
...
     channel_map = analysis_channels(channels, type_amp)
 
-    x, y, _, _ = trial_reshaper(t_t_i, t_i, data,
-                                mode=mode, fs=fs, k=downsample_rate,
-                                offset=offset,
-                                channel_map=channel_map,
-                                trial_length=trial_length)
-
     model = PcaRdaKdeModel(k_folds=k_folds)
+    x, y = data_reshaper(input_data_type=model.input_data_type,
```
tab-cmd: while we're here let's use descriptive variable names. `trials, labels`

bcipy/helpers/task.py
lawhead: Yes, this is what I had in mind, storing the appropriate class with the signal model, rather than the enum. Rather than calling:

`x, y = data_reshaper(input_data_type=model.input_data_type, ...)`

It would be something like:

```
x, y = model.reshape(trial_target_info=t_t_i,
                     timing_info=t_i,
                     eeg_data=data,
                     fs=fs,
                     trials_per_inquiry=parameters.get('stim_length'),
                     offset=offset,
                     channel_map=channel_map,
                     trial_length=trial_length)
```
or 

`x, y = model.reshaper.reshape(...)`

If we don't have the signal model we just use the appropriate reshaper directly.

tab-cmd (approved):
Suggested changes and small cleanup items. I'm happy with the PR / tickets you've made. Please address other PR reviewers and what you can of mine. Thanks!

bcipy/signal/process/transforms.py
nik-sm: Ahh ok - sounds good. 
Yea I suppose downsample isn't really a "filter" in the signal processing sense so I'll keep it in `transform.py`

lawhead (approved):

bcipy/helpers/task.py
nik-sm: I do agree with this, and perhaps `trial_labels` would be a reasonable name to use instead of `trial_target_info` - however as I started making this change, I realized this occurs in a TON of places.

Let's take care of the renaming in a follow-up?

nik-sm: (And I'll try to use clear naming within these reshaper functions themselves for the time being)

nik-sm: https://www.pivotaltracker.com/story/show/178308208

bcipy/helpers/copy_phrase_wrapper.py
nik-sm: I went with `data` for now, since it will sometimes be trials and sometimes inquiries

PR #148, opened 2021-06-03
Fix targetness assignments in write copy phrase triggers

tab-cmd:
# Overview

Updated the write copy phrase triggers method to fix targetness assignment. Originally this was to accommodate inquiry preview, however, it also fixes an uncaught bug when writing the calibration trigger during the first inquiry.

## Ticket

https://www.pivotaltracker.com/story/show/178253198

## Contributions

- `_write_triggers_from_inquiry_copy_phrase`: updated to fix targetness assignment, adding typing, updated to f-strings
- `test_triggers`: added test for `_write_triggers_from_inquiry_copy_phrase`

## Test

- `make test-all`
- `make lint`
- `make bci-gui`: ran copy phrase in fake data mode

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Nope.

### DIFF
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -145,61 +145,66 @@ def _write_triggers_from_inquiry_calibration(array: list,
     return trigger_file
 
 
-def _write_triggers_from_inquiry_copy_phrase(array,
-                                             trigger_file,
-                                             copy_text,
-                                             typed_text,
-                                             offset=None):
-    """
-    Write triggers from copy phrase.
+def _write_triggers_from_inquiry_copy_phrase(
+        triggers: list,
+        trigger_file: TextIO,
+        copy_phrase: str,
+        typed_text: str,
+        offset: bool = False) -> TextIO:
+    """Write triggers from copy phrase.
 
-    Helper Function to write trigger data to provided trigger_file. It assigns
-        target letter based on matching the next needed letter in typed text
-        then assigns target/nontarget label to following letters.
+    Helper Function to write trigger data to provided trigger_file in a copy phrase task.
+        It assigns target letter based on matching the next needed letter in typed text
+        then assigns target/nontarget label to following letters. It's assumed if offset
+        is true, only one record will be written (the offset_correction).
 
     It writes in the following order:
         (I) presented letter, (II) targetness, (III) timestamp
     """
-
+    # write the offset and return the file
     if offset:
         # extract the letter and timing from the array
-        (letter, time) = array
-        targetness = 'offset_correction'
-        trigger_file.write('%s %s %s' % (letter, targetness, time) + "\n")
+        (symbol, time) = triggers
+        trigger_file.write(f'{symbol} offset_correction {time}\n')
+        return trigger_file
 
+    spelling_length = len(typed_text)
+    last_typed = typed_text[-1] if typed_text else None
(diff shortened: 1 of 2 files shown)


bcipy/helpers/triggers.py
```diff
@@ -145,61 +145,66 @@ def _write_triggers_from_inquiry_calibration(array: list,
...
-        targetness = 'offset_correction'
-        trigger_file.write('%s %s %s' % (letter, targetness, time) + "\n")
-
+        (symbol, time) = triggers
+        trigger_file.write(f'{symbol} offset_correction {time}\n')
+        return trigger_file
+
+    spelling_length = len(typed_text)
+    last_typed = typed_text[-1] if typed_text else None
+    target_symbol = copy_phrase[spelling_length - 1]
```
lawhead, line 173: The copy_phrase task already computes the target symbol. I wonder if that should be provided as a parameter rather than computing it here. Or maybe we should factor out that code into a helper function. 

Also, with the copy phrase task method (`next_target`) we looked at the all of the spelled text, rather than just the last letter. That way if the phrase is 'HELLO' and they have spelled 'HEEL' the target would be backspace, rather than 'O'.

bcipy/helpers/triggers.py
```diff
@@ -145,61 +145,66 @@ def _write_triggers_from_inquiry_calibration(array: list,
...
+        elif 'bcipy_key_press' in symbol:
+            targetness = 'key_press'
+
+        # assign targetness to the core symbols
+        elif symbol == '+':
+            targetness = 'fixation'
+        elif target_symbol == symbol:
+            targetness = 'target'
+        else:
+            targetness = 'nontarget'
```
lawhead, line 204: We might want to factor out the symbol targetness labeling into its own helper function. Seems like it might be useful for other code.

What do you think about using a dict to store the mappings? It wouldn't capture all possible values of key press options but do we need that flexibility?

Something like:

    def symbol_targetness(symbol: str, target: str) -> str:
        if symbol == target:
            return 'target'
        labels = {
            'calibration_trigger': 'calib',
            'inquiry_preview': 'preview',
            'bcipy_key_press_space': 'key_press',
            'bcipy_key_press_escape': 'key_press',
            '+': 'fixation'
        }
        return symbol_labels.get(symbol, 'nontarget')

lawhead (approved):

bcipy/helpers/triggers.py
tab-cmd: This is an interesting point... I think we should talk more with the clinical team on how they determine correctness or targetness in this case. If we were going to report on copy phrase performance, for instance, would HEELO be 80% or  40% correct because there was no correction? Would then HEE<LLO be ~85%?

bcipy/helpers/triggers.py
tab-cmd: I agree with both of your points. I'm not sure about the key_press though since it could be any key on the keyboard (I believe?) For the sake of release, I will make a ticket to address this further. I think we can simplify this greatly!

bcipy/helpers/triggers.py
lawhead: I think this would be an interesting discussion for the language modeling team as well. It seems like the answer could depend on whether we're using letter vs word-based probabilities.

PR #149, opened 2021-06-04
calibration trigger refactor and cleanup

tab-cmd:
# Overview

Updated rsvp display to prevent calibration trigger from coming after the inquiry preview by encapsulating into _trigger_pulse(). Refactored the calibration trigger to use text and fixed uncaught bug with TriggerCallback/sound. Cleaned up static folder. Added tests. 

## Ticket

https://www.pivotaltracker.com/story/show/178007138

## Contributions

- `rsvp/display.py`: added `_trigger_pulse` method and ensured `first_run` was checked in both inquiry preview and do_inquiry. Update `core.wait` time. It was losing a single frame. 
- `helpers/stimuli`: updated `play_sound` to use the `TriggerCallback`. 
- `helpers/triggers`: added `CalibrationType`and surfaced some default paths. Added text calibration. Deprecate sound calibration.
- `static`: updated directory names to remove redundancy.

## Test

- make test-all
- I ran timing verification tasks with text, image, text + viewer to ensure it was still accurate and static offsets did not increase (they all hovered around .1s as expected). 

**Text + Viewer:**
offset_viewer_start
**Text:**
text_offset_start
**Image:**
offset_view_image_start

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Updated in-line. I don't think any README updates are needed.


bcipy/helpers/stimuli.py
```diff
@@ -65,7 +65,7 @@ def rsvp_inq_generator(query: list,
         if is_txt:
             sample = ['+']
         else:
-            sample = ['bcipy/static/images/bci_main_images/PLUS.png']
+            sample = ['bcipy/static/images/main/PLUS.png']
```
lawhead: This if/else statement gets repeated quite a lot in this file. Seems like it could be factored out into another function. Something like `get_fixation(is_txt: bool) -> str` (or `-> List[str]`).

bcipy/helpers/triggers.py
```diff
@@ -1,14 +1,35 @@
...
+
+    Enum to define the supported calibration trigger types.
+    """
+    TEXT = 'text'
+    IMAGE = 'image'
+
+    @classmethod
+    def list(cls):
+        """Returns all enum values as a list"""
+        return list(map(lambda c: c.value, cls))
```
lawhead, line 32: 👍

bcipy/helpers/triggers.py
```diff
@@ -26,72 +47,74 @@ def reset(self):
...
+        display.callOnFlip(trigger_callback.callback, experiment_clock, trigger_name)
+        if on_trigger is not None:
+            display.callOnFlip(on_trigger, trigger_name)
+
+        calibration_box.draw()
+        display.flip()
+
+    elif trigger_type == CalibrationType.TEXT.value:
+        calibration_text = visual.TextStim(display, text='')
+
```
lawhead, line 108: Everything after this line is essentially repeated in both clauses. We might want to factor the creation of the visual stim into another function: `visual_stim(trigger_type, display) -> visual.BaseVisualStim`

or make `CalibrationType` an Abstract class with subclasses (`TextCalibrationType`, `ImageCalibrationType`) that know how to create a visual stimuli from a display.

```  
    calibration_stim = visual_stim( trigger_type, display)
    display.callOnFlip(trigger_callback.callback, experiment_clock, trigger_name)
    if on_trigger is not None:
        display.callOnFlip(on_trigger, trigger_name)
    calibration_box.draw()
     display.flip()
```

lawhead (approved):

bcipy/gui/viewer/data_viewer.py
```diff
@@ -64,6 +59,11 @@ def __init__(self,
         # Default filter
         self.filter = Downsample(self.downsample_factor)
 
+        self.refresh_rate = refresh
```
tab-cmd, line 62: @lawhead I addressed all comments I intended in this PR. I added this after testing everything and running into an issue with downsample in the data viewer. Can you look this part over to make sure it looks ok?

Changed:

- Output of Downsample when fs is not defined
- samples per second uses the downsample rate 
- changed default (not sure how this impacts everything though)

bcipy/helpers/triggers.py
tab-cmd: I do think we'll need to do something like this. We will likely need more offset corrections throughout the sessions and will need to refactor to support that. I'll hold off for those tickets to implement something like this.

bcipy/gui/viewer/data_viewer.py
```diff
@@ -501,7 +501,7 @@ def main(data_file: str,
     parser.add_argument('-r',
                         '--refresh',
                         help='refresh rate in ms',
-                        default=500,
+                        default=300,
```
lawhead: This is the refresh rate of the viewer, not the EEG device. It was set to refresh the display every 500 ms (half second). Was this choppy? The downside to a more frequent refresh is more performance overhead.

bcipy/gui/viewer/data_viewer.py
lawhead: It seems like the samples_per_second is not correct. This value is used to get the raw data before the filters are applied so it doesn't seem like it would retrieve enough data.

bcipy/gui/viewer/data_viewer.py
tab-cmd: It wasn't buggy at 300 but I will revert since it is configurable!

PR #153, opened 2021-06-16
Stimuli Ordering

tab-cmd:
# Overview

Added stimuli ordering parameter and implemented in RSVP tasks. Various logging and typing improvements while I was in the files.

## Ticket

https://www.pivotaltracker.com/story/show/175769809

## Contributions

- `copy_phrase_wrapper`: update logging and exception handling. add stim order.
- `main_frame`: update `DecisionMaker` to use stim_order. Remove constants (unused).. consider writing a ticket to address? Use BACKSPACE_CHAR.
- `demo_stimuli_generation.py`: update imports and add a case showing the new ordering functionality.
- `bcipy/helpers/stimuli.py ` add `StimuliOrder`. Add method `alphabetize`. Implement `alphabetize` where stimuli are generated. Remove unused or abandoned generators.  update logging, typing and exception handling.
- `parameters.json`: add `stim_order`.
- Updated tasks to pull in and use the stim_order parameters in their generators.

## Test

- Added tests to `test_stimuli` and updated the copy phrase test with new parameter.
- `make test-all`
- `bcipy`: try running with random and alphabetical in both calibration and copy phrase. 

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? I don't think so.. I updated inline as I noticed.

### DIFF
--- a/bcipy/helpers/stimuli.py
+++ b/bcipy/helpers/stimuli.py
@@ -1,19 +1,40 @@
 import glob
 import itertools
+import re
 import logging
 import random
 from os import path, sep
-from typing import Iterator, List, NamedTuple
+from typing import Iterator, List, Tuple, NamedTuple
 
 import numpy as np
+from enum import Enum
+
 import sounddevice as sd
 import soundfile as sf
 from PIL import Image
+
+from bcipy.helpers.exceptions import BciPyCoreException
+
 from psychopy import core
 
 # Prevents pillow from filling the console with debug info
 logging.getLogger('PIL').setLevel(logging.WARNING)
+log = logging.getLogger(__name__)
 DEFAULT_FIXATION_PATH = 'bcipy/static/images/main/PLUS.png'
+
+
+class StimuliOrder(Enum):
+    """Stimuli Order.
+
+    Enum to define the ordering of stimuli for inquiry.
+    """
+    RANDOM = 'random'
+    ALPHABETICAL = 'alphabetical'
+
+    @classmethod
+    def list(cls):
+        """Returns all enum values as a list"""
+        return list(map(lambda c: c.value, cls))
 
 
 class InquirySchedule(NamedTuple):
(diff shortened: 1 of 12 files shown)


bcipy/tasks/rsvp/main_frame.py
```diff
@@ -229,6 +231,7 @@ def decide(self, p) -> Tuple[bool, InquirySchedule]:
             return True, None
         else:
             stimuli = self.schedule_inquiry()
+            print(stimuli)
```
lawhead: Looks like there is a debug print statement here.

bcipy/helpers/stimuli.py
```diff
@@ -31,30 +52,38 @@ class InquirySchedule(NamedTuple):
...
+    Given a list of string stimuli, return a list of sorted stimuli by alphabet.
+    """
+    return sorted(stimuli, key=lambda x: re.sub(r'[^a-zA-Z0-9 \n\.]', 'ZZ', x).lower())
+
+
+def rsvp_inq_generator(query: List[str],
+                       timing: List[float] = [1, 0.2],
+                       color: List[str] = ['red', 'white'],
+                       stim_number: int = 1,
+                       stim_order: StimuliOrder = StimuliOrder.RANDOM.value,
```
lawhead: Based on the default value and your usage below, it looks like `stim_order` should have type `str`. Otherwise we need to change the default value should to `StimuliOrder.RANDOM` and adjust the comparison. I'm okay with this either way. It's coming out of the parameters file as a string so that might be easier to deal with (although maybe we should discuss supporting enums and doing a cast). However, it might be easier to communicate expected values if it has the type `StimuliOrder`.

lawhead: Looks like this comment applies to the other functions in this file as well.

bcipy/helpers/tests/test_stimuli.py
```diff
@@ -314,5 +359,30 @@ def test_image_fixation_uses_default(self):
...
+    def test_alphabetize_image_name(self):
+        list_of_images = ['testing.png', '_ddtt.jpeg', 'bci_image.png']
+        expected = ['bci_image.png', 'testing.png', '_ddtt.jpeg']
+        response = alphabetize(list_of_images)
+        self.assertEqual(expected, response)
+
+    def test_alphabetize_special_characters_at_end(self):
+        character = '<'
+        expected = ['A', 'G', 'Q', 'Z', character]
+        self.list_to_alphabetize.append(character)
```
lawhead: It might be more instructive to put the backspace character in the middle or near the front, since we expect it to be sorted near the end. I would also add the space character here. I tested locally and these work as expected. 👍

lawhead (approved):

PR #154, opened 2021-06-18
PyQt Signal viewer

lawhead:
# Overview

Ported the signal viewer to use PyQt5. Note that the wxpython version is still available in the code for testing purposes.

## Ticket

- https://www.pivotaltracker.com/story/show/175566482
- https://www.pivotaltracker.com/story/show/175543734

## Contributions

- Ported Signal Data Viewer to PyQt5
- Fixed issues with filters in wxPython version of the viewer and refactored for clarity.
- Updated the data streamer with an increased timeout, which fixed an issue with the viewer initialization causing an offset cursor.
- Added the ability to set the fixed scale from within the viewer
- Added caching for each axes to maintain the overall min and max values which provide a more consistent scale over time.

## Test

- Run the data viewer against a sample stream or file. Things should work identically to the wxPython versions. The existing wxPython version has been maintained for testing comparisons.


bcipy/gui/viewer/qt_data_viewer.py
```diff
@@ -0,0 +1,585 @@
+"""EEG Data Viewer"""
```
lawhead, line 1: Most of the functionality related to managing data and rendering using Matplotlib is identical to the wxpython version thanks to the multiple backends supported by Matplotlib. The main differences are in the widget creation (see `initUI`), events, and timer.

tab-cmd:
Very cool to see this PR! 

I changed data_viewer to qt_data_viewer in `start_viewer` and tried to run a calibration. I get the following error (although it does launch and I can calibrate): 
```
bcipy/gui/viewer/qt_data_viewer.py:544: UserWarning: Tight layout not applied. tight_layout cannot make axes height small enough to accommodate all axes decorations
  sys.exit(app.exec_()
```

Comes through like this regardless of autoscale or filtering:
Screen Shot 2021-06-18 at 12 30 09 PM

The old viewer (just checked to make sure it worked on my machine):
Screen Shot 2021-06-18 at 12 30 47 PM

lawhead:
Hm, I'll look into that. I was mostly running this stand-alone while connecting to an LSL server. Seems like there is an initialization issue.

lawhead:
@sci-tab, I merged the latest from 1.5.1 and added code to handle multiple monitors. When I tested using the qt viewer with calibration I was not able to replicate your error. Can you update your code and see if it's still and issue on your machine?

tab-cmd:
> @sci-tab, I merged the latest from 1.5.1 and added code to handle multiple monitors. When I tested using the qt viewer with calibration I was not able to replicate your error. Can you update your code and see if it's still an issue on your machine?

I pulled the code and re-installed it. It still happens. I tried with a second monitor and without. I'm on macOS High Sierra. Interestingly, if I make it bigger by dragging with a mouse, it works okay (see image) but still looks like the labels are overlapping.

Screen Shot 2021-06-21 at 12 33 58 PM

I wonder if some padding or slight change to default sizes might help? I'll test on Windows later today.

lawhead:
> > @sci-tab, I merged the latest from 1.5.1 and added code to handle multiple monitors. When I tested using the qt viewer with calibration I was not able to replicate your error. Can you update your code and see if it's still an issue on your machine?
> 
> I pulled the code and re-installed it. It still happens. I tried with a second monitor and without. I'm on macOS High Sierra. Interestingly, if I make it bigger by dragging with a mouse, it works okay (see image) but still looks like the labels are overlapping.
> 
> Screen Shot 2021-06-21 at 12 33 58 PM
> 
> I wonder if some padding or slight change to default sizes might help? I'll test on Windows later today.

I was able to replicate this locally by adjusting the `tight_layout` padding in the `init_canvas` method. I'll play around with some settings and see if I can get this to work.

bcipy/gui/viewer/qt_data_viewer.py
```diff
@@ -0,0 +1,727 @@
...
+
+        # figure size is in inches.
+        self.figure = Figure(figsize=(12, 9),
+                             dpi=72,
+                             tight_layout={'pad': 0.0})
+        # space between axis label and tick labels
+        self.yaxis_label_space = 60
+        self.yaxis_label_fontsize = 12
+        # fixed width font so we can adjust spacing predictably
+        self.yaxis_tick_font = 'DejaVu Sans Mono'
```
tab-cmd: Pull up to class variable

bcipy/gui/viewer/qt_data_viewer.py
```diff
@@ -0,0 +1,727 @@
...
+        vbox = QVBoxLayout()
+
+        self.init_canvas()
+        vbox.addWidget(self.canvas)
+
+        # Toolbar
+        self.toolbar = QVBoxLayout()
+
+        controls = FixedHeightHBox()
+        font_size = 11
```
tab-cmd: Pull up to class value

bcipy/gui/viewer/qt_data_viewer.py
```diff
@@ -0,0 +1,727 @@
...
+        for i, _channel in enumerate(self.active_channel_indices):
+            # Initial data plotted will be a list 0.0; this will be updated
+            # with real data every tick using the `update_plots` method.
+            data = channel_data[i].tolist()
+            self.axes[i].set_ybound(lower=self.y_min, upper=self.y_max)
+            self.axes[i].plot(data, linewidth=0.8)
+            # plot cursor
+            self.axes[i].axvline(cursor_x, color='r')
+
+    def update_buffer(self, fast_forward=False):
```
tab-cmd: typing

bcipy/gui/viewer/qt_data_viewer.py
```diff
@@ -0,0 +1,727 @@
...
+        # assume at least 2 digits to start.
+        baseline_chars = 2
+        # Value determined by trial and error; this may change if the tick font
+        # or font size is adjusted.
+        ytick_digit_width = 7
+        return self.yaxis_label_space - \
+            ((chars - baseline_chars) * ytick_digit_width)
+
+
+def lsl_data():
```
tab-cmd: return types

tab-cmd (approved):
Works well for me locally! I think we can safely deprecate the old viewer, replacing this, and remove WxPython from our requirements file! 

The only difference I noted was in changing aspects of the viewer (filter, channels to display) it now pauses and requires a user to hit start instead of refreshing. I don't dislike the behavior, but we should call it out to the team and/or perhaps make the start button more pronounced after pausing occurs (change color to green etc).

lawhead:
> The only difference I noted was in changing aspects of the viewer (filter, channels to display) it now pauses and requires a user to hit start instead of refreshing. I don't dislike the behavior, but we should call it out to the team and/or perhaps make the start button more pronounced after pausing occurs (change color to green etc).

Yes, this is something I was going to call out. The UI felt very laggy when I tried to resume immediately and I often found that more than 1 thing needed be changed before resuming so I went ahead and just paused on every change. If we don't like this we can resume to the previous behavior. I liked your suggestion of turning the Start button green so I just added that refinement.

PR #157, opened 2021-06-24
Replace data viewer

lawhead:
# Overview

Replaced the original WxPython version of the signal data viewer with the new PyQt version.

## Ticket

https://www.pivotaltracker.com/story/show/175543734

## Contributions

- Replaced wxPython signal viewer with PyQt version
- Removed wxPython dependency from the project

## Test

- Run the data viewer as before, either standalone with a mock server or in conjunction with a Calibration experiment. The viewer used should be the PyQt5 version. In this version the Start button turns green when paused.


lawhead (commented):
The diff makes this seem more complicated than it is. I deleted the original version and renamed `qt_data_viewer.py` to `data_viewer.py`.

tab-cmd (approved):

PR #160, opened 2021-07-23
#175674845 ; module for raw data format

lawhead:
# Overview

Many BciPy modules depend on our raw_data.csv format to be structured in a specific way, and the code that writes and reads this data format is repeated in several places throughout the codebase. This PR creates a raw_data module that provides an interface for reading and writing the raw data format without knowing the details of its implementation.

## Ticket

https://www.pivotaltracker.com/story/show/175674845

## Contributions

- Created a single module responsible for reading and writing data into our raw data format.
- Added unit tests for all module functionality.
- Refactored existing code to use the new module.
- Provided methods to generate sample data for tests.
- Refactored existing tests to use data generated from the module (see test_convert.py and test_triggers.py).

## Test

- Ran all unit and integration tests.
- Ran a copy phrase task and performed offline analysis to confirm that this functionality still worked.
- Started the data viewer with a setting to read from an existing raw_data.csv file to confirm that this functionality still works.

### DIFF
--- a/bcipy/helpers/load.py
+++ b/bcipy/helpers/load.py
@@ -8,10 +8,9 @@ from tkinter import Tk
 from tkinter.filedialog import askdirectory, askopenfilename
 from typing import Any, Dict, List, Tuple
 
-import numpy as np
-import pandas as pd
 from bcipy.helpers.exceptions import BciPyCoreException, InvalidExperimentException
 from bcipy.helpers.parameters import DEFAULT_PARAMETERS_PATH, Parameters
+from bcipy.helpers.raw_data import RawData
 from bcipy.helpers.system_utils import (
     DEFAULT_EXPERIMENT_PATH,
     DEFAULT_FIELD_PATH,
@@ -199,7 +198,17 @@ def load_signal_model(model_class: SignalModel,
     return signal_model, filename
 
 
-def load_csv_data(filename: str = None) -> str:
+def choose_csv_file(filename: str = None) -> str:
+    """GUI prompt to select a csv file from the file system.
+
+    Parameters
+    ----------
+    - filename : optional filename to use; if provided the GUI is not shown.
+
+    Returns
+    -------
+    file name of selected file; throws an exception if the file is not a csv.
+    """
     if not filename:
         try:
             Tk().withdraw()  # we don't want a full GUI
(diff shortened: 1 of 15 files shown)


bcipy/helpers/load.py
```diff
@@ -199,7 +198,17 @@ def load_signal_model(model_class: SignalModel,
     return signal_model, filename
 
 
-def load_csv_data(filename: str = None) -> str:
+def choose_csv_file(filename: str = None) -> str:
```
tab-cmd, line 201: 👍

bcipy/helpers/raw_data.py
```diff
@@ -0,0 +1,303 @@
...
+
+import numpy as np
+import pandas as pd
+
+from bcipy.signal.generator.generator import gen_random_data
+
+TIMESTAMP_COLUMN = 'timestamp'
+
+
+class RawData:
```
tab-cmd, line 13: We spoke about this in person briefly! I think this could reside in acquisition? I don't foresee using this structure outside that module. If you can think of one, we should leave it here.

bcipy/helpers/raw_data.py
```diff
@@ -0,0 +1,303 @@
...
+    def __exit__(self, *args, **kwargs):
+        """Exit the context manager. Close resources"""
+        self._file_obj.close()
+
+    def writerow(self, row: List) -> None:
+        assert self._csv_writer, "Writer must be initialized"
+        self._csv_writer.writerow(row)
+
+
+def load(filename: str) -> RawData:
```
tab-cmd, line 208: I like that we can support both full read and iterative. It makes a lot of sense

bcipy/helpers/tests/test_raw_data.py
```diff
@@ -0,0 +1,272 @@
...
+        columns = ['timestamp', 'ch1', 'ch2', 'ch3', 'TRG']
+        data = RawData(daq_type='DSI-24', sample_rate=300.0, columns=columns)
+        self.assertEqual(data.daq_type, 'DSI-24')
+        self.assertEqual(data.sample_rate, 300.0)
+        self.assertEqual(data.columns, columns)
+
+    def test_write_with_no_data(self):
+        """Test that raw data can be persisted to disk."""
+        path = Path(self.temp_dir, 'test_raw_data.csv')
+        data = RawData(daq_type='DSI-24',
```
tab-cmd: There are some repeated steps in many of these unittests. It may be more change resilient to bring path/data up to `self` and reconstruct them when required (using the LSL daq type, etc)

bcipy/helpers/tests/test_triggers.py
```diff
@@ -13,67 +16,56 @@
...
-def sample_raw_data(trigger_seq: List[Tuple[str, str]] = [],
-                    first_trg_time: int = 100,
-                    trigger_interval: int = 10,
-                    daq_type: str = 'TestStream',
-                    sample_rate: int = 300,
-                    ch_names: List[str] = ['c1', 'c2',
-                                           'c3']) -> Tuple[str, List[float]]:
-    """Helper function for creating mock data that looks like the raw_data.csv
-    output. Adds trigger data to the TRG column at the specified interval.
+def write_sample_raw_data(raw_data_path: str,
```
tab-cmd: write_sample_trigger_data? We can reserve RawData as a concept for acquisition data output.

tab-cmd (approved):
Some suggestions, but I like the approach!

bcipy/helpers/raw_data.py
lawhead: The RawData object is now used in a number of different places including offline_analysis. I think it makes sense to have it at a higher level than acquisition since it's a cross cutting concern. I do think it would make sense to introduce a `core` module for things like this, though. It's not really a helper.

PR #161, opened 2021-07-30
#179057693 ; Analysis Channels

lawhead:
# Overview

We need the ability to configure which channels are used for analysis (or excluded). Prior to refactoring to use DeviceSpecs, we maintained a list of analysis channels for each device. However, this information was lost in the refactor and the code currently uses all channels (except 'TRG'). This pull request modifies the DeviceSpec class with an additional attribute that allows us to configure channels which should be ignored for modeling purposes.

## Ticket

https://www.pivotaltracker.com/story/show/179057693

## Contributions

- Created additional DeviceSpec parameter
- Updated devices.json configuration with channels to exclude
- Updated the acquisition helper
- Updated unit tests

## Test

- Ran the test suite to confirm that all unit tests pass

### DIFF
--- a/bcipy/acquisition/devices.json
+++ b/bcipy/acquisition/devices.json
@@ -6,7 +6,8 @@
             "T5", "O1", "O2", "X3", "X2", "F7", "F8", "X1", "A2", "T6", "T4", "TRG" ],
         "connection_methods": [ "TCP", "LSL" ],
         "sample_rate": 300.0,
-        "description": "Wearable Sensing DSI-24"
+        "description": "Wearable Sensing DSI-24",
+        "excluded_from_analysis": ["Pz", "TRG", "X1", "X2", "X3"]
     },
     {
         "name": "DSI-VR300",
@@ -14,7 +15,8 @@
         "channels": [ "P4", "Fz", "Pz", "F7", "PO8", "PO7", "Oz", "TRG" ],
         "sample_rate": 300.0,
         "connection_methods": ["TCP", "LSL"],
-        "description": "Wearable Sensing DSI-VR300"
+        "description": "Wearable Sensing DSI-VR300",
+        "excluded_from_analysis": ["TRG"]
     },
     {
         "name": "g.USBamp-1",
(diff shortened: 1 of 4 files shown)


bcipy/acquisition/devices.json
```diff
@@ -6,15 +6,17 @@
             "T5", "O1", "O2", "X3", "X2", "F7", "F8", "X1", "A2", "T6", "T4", "TRG" ],
         "connection_methods": [ "TCP", "LSL" ],
         "sample_rate": 300.0,
-        "description": "Wearable Sensing DSI-24"
+        "description": "Wearable Sensing DSI-24",
+        "excluded_from_analysis": ["Pz", "TRG", "X1", "X2", "X3"]
```
tab-cmd: Let's keep PZ, remove A1/A2

bcipy/acquisition/devices.json
```diff
@@ -49,6 +52,7 @@
         "channels": [ "P4", "Fz", "Pz", "F7", "PO8", "PO7", "Oz", "TRG" ],
         "sample_rate": 300.0,
         "connection_methods": ["TCP", "LSL"],
-        "description": "Alias to DSI-VR300 for backwards compatibility."
+        "description": "Alias to DSI-VR300 for backwards compatibility.",
```
tab-cmd, line 55: If we don't use them in practice anymore (the backward compatibility devices), let's remove them for clarity! I don't think any other users rely on it being here.

bcipy/acquisition/devices.json
```diff
@@ -28,11 +30,12 @@
...
         "content_type": "EEG",
-        "channels": [ "P3", "C3", "F3", "Fz", "F4", "C4", "P4",  "Cz", "A1", "Fp1", "Fp2", "T3",
-            "T5", "O1", "O2", "F7", "F8", "A2", "T6", "T4", "TRG" ],
+        "channels": [ "P3", "C3", "F3", "Fz", "F4", "C4", "P4", "Cz", "Pz", "Fp1", "Fp2", "T3",
+            "T5", "O1", "O2", "X3", "X2", "F7", "F8", "X1", "A2", "T6", "T4", "TRG" ],
         "connection_methods": [ "TCP", "LSL" ],
         "sample_rate": 300.0,
-        "description": "Alias to DSI-24 for backwards compatibility and demos."
+        "description": "Alias to DSI-24 for backwards compatibility and demos.",
+        "excluded_from_analysis": ["Pz", "TRG", "X1", "X2", "X3"]
```
tab-cmd: Should we error or warn on invalid channel names? I'm not sure we want the app to fail, but it may avoid some silent failing (does capitalization matter?). 

"excluded_from_analysis": ["Pz", "TG", "X1", "X2", "X3"]. ---->

Warning "Excluded channel TG not found in self.channels={[]}, Skiping."

tab-cmd (approved):

tab-cmd:
I was also thinking it would be good to report this where applicable. Do you see an easy way to do that in `offline_analysis`? We can make a ticket for it if not!

bcipy/acquisition/devices.json
lawhead: They are still used in demo code and servers. I'll try to phase them out, though.

PR #162, opened 2021-08-02
Refactor task

tab-cmd:
# Overview

Restructures and renames submodules within `task`. The changes are to help better support the addition of new paradigms and tasks in the near future (Matrix). 

## Ticket

https://www.pivotaltracker.com/story/show/179031815

## Contributions

- Moved all decision, evidence, and data code from `task` into a submodule `control` (naming very negotiable!). Reduced the name of all submodules. `stopping_criteria` --> `criteria`. `session_data` --> `data`. etc. All logic changed will be described below.
- `bcipy/task/main.py`: renamed task.py to main.py to avoid repetition. Updated to use ABC.
- `start_task`: remove auc_filename. It was unused.
- `task`: updated init to reduce import overhead / improve clarity.
- `control.handler.save_history`: log a warning and return None in favor of 0. 

## Test

- `make test-all`
- `bcipy`

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)?  Yes! I updated the task and main README. I didn't see any other locations, but please point any I might have missed.


lawhead (approved):
I generally like this re-organization and I think it makes a lot of sense when we add more paradigms. The name 'control' seems alright to me. I can't think of anything better at the moment. My one hesitation is including the data classes under 'control'. I wonder if data should be included at the top level (/task/data.py) or in its own space (/task/data/...). I don't feel strongly about this, though.

PR #164, opened 2021-08-04
Language model base class

tab-cmd:
# Overview

Added the language model base class. This will serve as the building block for all new language models in BciPy V2. The `language_model` module will be deprecated. 

## Ticket

https://www.pivotaltracker.com/story/show/178695658

## Contributions

- Add `language` module and `main.py`. Contains ABC for `LanguageModel` in line with our team redesign. 
- Add README describing module briefly. 

Langauge Model Redesign

## Test

- `make test-all`

## Documentation

### DIFF
--- a/bcipy/language/main.py
+++ b/bcipy/language/main.py
@@ -35,3 +35,11 @@ class LanguageModel(ABC):
     def load(self, path: Path) -> None:
         """Restore model state from the provided checkpoint"""
         ...
+
+    def reset(self) -> None:
+        """Reset language model state"""
+        ...
+
+    def state_update(self, evidence: List[Tuple]) -> None:
+        """Update state by predicting and updating"""
+        ...
--- a/bcipy/language/main.py
+++ b/bcipy/language/main.py
@@ -1,0 +1,37 @@
+from abc import ABC, abstractmethod
+from pathlib import Path
+from typing import Tuple, List, Union
+from enum import Enum
+
+
+class ResponseType(Enum):
+    SYMBOL = 'Symbol'
+    WORD = 'Word'
+
+
+class LanguageModel(ABC):
+
+    response_type: ResponseType
+    symbol_set: List[str]
+
+    @abstractmethod
+    def predict(self, evidence: List[Tuple]) -> List[tuple]:
+        """
+        Using the provided data, compute log likelihoods over the entire symbol set.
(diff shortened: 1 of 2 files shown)


bcipy/language/main.py
```diff
@@ -0,0 +1,37 @@
...
+        """
+        Using the provided data, compute log likelihoods over the entire symbol set.
+        Args:
+            evidence - [('A', .08), ('B', .1)]
+
+        Response:
+            probability - dependant on response type, a list of words or symbols with probability 
+        """
+        ...
+
```
lawhead, line 28: In looking back over the 'prelm' and 'oclm' models I don't see how they would use `update` and `load`. Maybe we should remove the `@abstractmethod` annotation, which would force them to implement these methods, and just provide a default method with `pass`. Also, update doesn't have any parameters. How would the class update its state? The existing models do some state management in the `state_update` method, which sort of combines this notion of `predict` and `update`.

bcipy/language/main.py
```diff
@@ -0,0 +1,37 @@
...
+    @abstractmethod
+    def update(self) -> None:
+        """Update the model state"""
+        ...
+
+    @abstractmethod
+    def load(self, path: Path) -> None:
+        """Restore model state from the provided checkpoint"""
+        ...
```
lawhead: We may want an optional method to `reset` any state.

tab-cmd: Good idea! I'll add it

lawhead (approved):
This looks like a good start and we can evolve it as we get feedback from the LM team. I'm looking forward to the RandomLangageModel being implemented.

bcipy/language/main.py
tab-cmd: They definitely wouldn't use those methods. The architecture will change substantially here to more closely match the signal model and other LM library patterns, as opposed to interacting with an API we assume is serving locally. In the new approach, after some offline training (outside of BciPy) an artifact will be available to load (not sure of the file type yet that's a SPIKE ticket we have created). The first model will almost certainly come from [huggingface](https://huggingface.co/).

In terms of input into the methods... I'm not sure what will be required as this is a bit outside of my field. I think it makes sense to have a `state_update` that combines them explicitly (still requiring them separately though). I'll add some args for update.

PR #168, opened 2021-08-30
PyQt file dialog

lawhead:
# Overview

Replaced Tk dialog instances with an alternative GUI that uses PyQt Widgets.

## Ticket

https://www.pivotaltracker.com/story/show/176100240

## Contributions

- Created a new stand-alone widget to encapsulate the PyQt file dialog widgets while providing the simplicity of the interface provided by Tk.
- Replaced the existing calls to Tk dialogs with the new PyQt version.

## Test

- Tested existing uses of file dialogs and directory prompts, including offline_analysis and Copy Phrase task to ensure that the user is prompted for the correct file type and the computation continues after a file or directory is chosen.
- Ensured that all unit tests ran successfully.

### DIFF
--- a/bcipy/acquisition/datastream/producer.py
+++ b/bcipy/acquisition/datastream/producer.py
@@ -30,7 +30,7 @@ class Producer(threading.Thread):
     def __init__(self,
                  queue,
                  freq=1 / 100,
-                 generator=random_data_generator(),
+                 generator=None,
                  maxiters=None):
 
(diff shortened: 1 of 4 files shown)


tab-cmd (approved):
This is so nice! 😭  The failing ubuntu builds can be fixed by pulling or rebasing from the release branch. I don't see any issue merging as is though, I tested it locally. Thanks!

PR #173, opened 2021-09-24
Adds methods to convert.py to support data compression/decompression

theJokerEvoker:
# Overview

Adds methods to convert.py to support data compression/decompression.

## Ticket

https://www.pivotaltracker.com/story/show/179306557

## Contributions

- Modified convert.py, added compress, decompress, archive_list, tar_name_checker
- compress - Adds files to a tar archive and compresses them using the gzip compression format.
- decompress - Takes .tar.gz archive, decompresses, and extracts its contents. Should only be used with files created by compress() method.
- archive_list - Takes .tar.gz archive and returns a full list of its contents, including folders, subfolders, and files, with their full relative paths and names.
- tar_name_checker - Helper method that takes a tar file name and checks it for the appropriate file extension (".tar.gz" for now), returning it if it already does, and adding the extension and then returning it if it does not.

## Test

- Run test_convert.py unit test suite
- Tested successfully on BciPy data recording, compressed 686 MB folder to 572 MB archive (17% compression)

## Documentation

- Added docstrings where applicable

### DIFF
--- a/bcipy/helpers/convert.py
+++ b/bcipy/helpers/convert.py
@@ -217,10 +217,25 @@ def edf_annotations(triggers: List[Tuple[str, str, float]],
 
 
 def compress(tar_file_name: str, members: List[str]) -> None:
-    """Compress.
-       Adds files (members) to a tar_file and compresses them using gzip
+    """
+    File compression and archiving.
 
-    tar_file_name (str): the name of resulting compressed file
+    Adds files to a tar archive and compresses them using the gzip compression
+        format.
+
+    Parameters
+    ----------
+    tar_file_name (str): name of resulting compressed tar archive. Input string
+        can either include or not include file extension (.tar.gz), as this will
+        be checked.
+    members (List[str]): list of files and folders to be compressed into archive.
+        Each file or folder requires the relative file path and full file extension.
+        Individual file paths and names that are very long may throw an error
+        upon extraction, proceed with caution.
+
+    Returns
+    -------
+        None
     """
 
     for member in members:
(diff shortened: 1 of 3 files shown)


bcipy/helpers/convert.py
```diff
@@ -2,18 +2,22 @@
...
 from pyedflib import FILETYPE_EDFPLUS, EdfWriter
+from tqdm import tqdm
 
 from bcipy.helpers.load import load_json_parameters, load_raw_data, extract_mode
 from bcipy.helpers.triggers import trigger_decoder, apply_trigger_offset, trigger_durations
 
 
 logger = logging.getLogger(__name__)
 
+FILE_LENGTH_LIMIT = 150
```
tab-cmd, line 19: This could be an input argument with a default, but this seems okay too since it's a warning

bcipy/helpers/convert.py
```diff
@@ -210,3 +214,79 @@ def edf_annotations(triggers: List[Tuple[str, str, float]],
     """
     return [(timestamp, durations.get(targetness, 0.0), label)
             for (label, targetness, timestamp) in triggers]
+
+
+def compress(tar_file_name: str, members: List[str]) -> None:
+    """Compress.
```
tab-cmd: Update docstrings to match other methods

bcipy/helpers/convert.py
```diff
@@ -210,3 +214,79 @@ def edf_annotations(triggers: List[Tuple[str, str, float]],
...
+        progress = tqdm(members)
+        for member in progress:
+            # Adds file/folder to the tar file and compresses it
+            tar.add(member)
+            # Sets progress description of progress bar
+            progress.set_description(f"Compressing {member}")
+
+
+def decompress(tar_file: str, path: str) -> None:
+    """Extracts tar_file and places specified "members" in "path"
```
tab-cmd: Same as above

tab-cmd (approved):
Consider suggestions and update docstings

bcipy/helpers/convert.py
```diff
@@ -210,3 +214,79 @@ def edf_annotations(triggers: List[Tuple[str, str, float]],
...
+       If members remains None, all members will be extracted
+    """
+
+    full_tar_name = f'{tar_file}.tar.gz'
```
lawhead: It's not clear from the method documentation that you only want the base name of the file. You should either update the documentation or make the implementation more lenient by checking for the suffix (or stripping it) before adding it here. The same comment applies to the other methods. Given the repetition, that code could be factored out into another function.

lawhead (approved):

PR #174, opened 2021-09-24
InsufficientDataException fix

lawhead:
# Overview

During execution of the Copy Phrase task, the `get_data_for_decision` function in the task helper would sometimes throw an `InsufficientDataException` after querying the acquisition client for data. The client had enough data in the desired range but the timestamp alignment was slightly off. The query was modified to explicitly ask for a given number of samples.

## Ticket

https://www.pivotaltracker.com/story/show/179644506

## Contributions

- Added utility methods to raw_data module to assist in debugging.
- Modified Copy Phrase task to query the acquisition module for a given number of samples.

## Test

- The Copy Phrase task should run without throwing an InsufficientDataException

### DIFF
--- a/bcipy/acquisition/protocols/lsl/lsl_client.py
+++ b/bcipy/acquisition/protocols/lsl/lsl_client.py
@@ -13,6 +13,7 @@ from bcipy.acquisition.protocols.lsl.lsl_connector import (channel_names,
 from bcipy.acquisition.protocols.lsl.lsl_recorder import LslRecordingThread
 from bcipy.acquisition.record import Record
 from bcipy.helpers.clock import Clock
+from bcipy.gui.viewer.ring_buffer import RingBuffer
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -13,6 +13,7 @@
 from bcipy.acquisition.protocols.lsl.lsl_recorder import LslRecordingThread
 from bcipy.acquisition.record import Record
 from bcipy.helpers.clock import Clock
+from bcipy.gui.viewer.ring_buffer import RingBuffer
```
tab-cmd, line 16: We should move this outside of gui... not sure the best place

tab-cmd (approved):
Thanks for the fix. I was able to calibrate and run several copy phrase sessions with the change!

bcipy/acquisition/protocols/lsl/lsl_client.py
lawhead: Agreed. I'll create a followup ticket for this.

PR #177, opened 2021-10-13
Add preview only mode

tab-cmd:
# Overview

Added the ability to preview an inquiry without a progress method. This is configured by setting preview_inquiry to True and preview_inquiry_progress_method to 0 (the new default).

## Ticket

https://www.pivotaltracker.com/story/show/179844270

## Contributions

- `parameters.json`: Added an option to preview_inquiry_progress_method (0 == preview only)
- `rsvp.display.py`: Updated display to factor in a preview only option. The get_key_press is still used and the response ignored. Initial testing with that removed led to odd behavior alongside our get_key_press call in copy_phrase (no triggers were written). Added `preview_only` to PreviewInquiryProperties.
- `task.paradigm.rsvp.copy_phrase.py`: added a check for preview only mode to determine whether to add button evidence 

## Test

- `make test-all`
- ran copy phrase in fake data mode with inquiry preview on (all preview_inquiry_progress_method conditions) and off.

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Verify the updates you did here. Updated inline documentation as needed.

### DIFF
--- a/bcipy/display/rsvp/display.py
+++ b/bcipy/display/rsvp/display.py
@@ -182,14 +182,15 @@ class PreviewInquiryProperties:
         """Initialize Inquiry Preview Parameters.
 
         preview_inquiry_length(float): Length of time in seconds to present the inquiry preview
-        preview_inquiry_progress_method(int): Method of progression for inquiry preview. 1 == press to accept
-            inquiry 2 == press to skip inquiry
+        preview_inquiry_progress_method(int): Method of progression for inquiry preview.
+            0 == preview only; 1 == press to accept inquiry; 2 == press to skip inquiry.
         preview_inquiry_key_input(str): Defines which key should be listened to for progressing
         preview_inquiry_isi(float): Length of time after displaying the inquiry preview to display a blank screen
         """
         self.preview_inquiry_length = preview_inquiry_length
         self.preview_inquiry_key_input = preview_inquiry_key_input
         self.press_to_accept = True if preview_inquiry_progress_method == 1 else False
+        self.preview_only = True if preview_inquiry_progress_method == 0 else False
         self.preview_inquiry_isi = preview_inquiry_isi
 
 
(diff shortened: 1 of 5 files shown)


bcipy/display/rsvp/display.py
```diff
@@ -424,22 +425,26 @@ def preview_inquiry(self) -> Tuple[List[float], bool]:
...
         while timer.getTime() > 0:
             # wait for a key press event
             response = get_key_press(
                 key_list=[self._preview_inquiry.preview_inquiry_key_input],
                 clock=self.experiment_clock,
             )
-            if response:
+
+            # break if a response given unless this is preview only and wait the timer out
+            if response and not self._preview_inquiry.preview_only:
```
lawhead, line 442: Given that the default value of `self._preview_inquiry` is None, we probably want an assertion at the top of this method that it exists to prevent an exception here (and elsewhere in the method). This would fix a lot of mypy errors since it infers the type of self_preview_inquiry to be `Optional[PreviewInquiryProperties]`.

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -451,8 +451,9 @@ def compute_button_press_evidence(
             tuple of (evidence type, evidence) or None if inquiry preview is
             not enabled.
         """
-        if not self.parameters[
-                'show_preview_inquiry'] or not self.current_inquiry:
+        if not self.parameters['show_preview_inquiry'] \
+                or not self.current_inquiry \
+                or self.rsvp._preview_inquiry.preview_only:
```
lawhead: Since `_preview_inquiry` is accessed outside of the class maybe we should remove the `_` prefix. Otherwise, would it make sense to have a `preview_only` method in the display class?

lawhead (approved):
Looks like there is some linting that needs fixing before merge.

bcipy/task/paradigm/rsvp/copy_phrase.py
tab-cmd: I ended up adding it to our `Parameters` to avoid duplicating where we look for and set that property. Let me know if that works of if you have a better idea!

lawhead: Hm, interesting approach. It's essentially a computed parameter. I like the idea of formalizing this idea a bit more so that maybe it's specified in the parameters.json file somehow. Seems like it could be useful in other contexts.

PR #178, opened 2021-10-13
Backspace always shown

tab-cmd:
# Overview

Added constants to stimuli agent and cleaned up best_selection methods. We should revisit this as the number of constants grow - it could result in unexpected behaviors with some agents. We only allow for setting of backspace as a constant at this time. 

## Ticket

https://www.pivotaltracker.com/story/show/179128130

## Contributions

- `helpers.stimuli.py`: clean up language to match glossary definitions. Remove unnecessary list comprehension.
- `control.handler.py`: added the ability to set constants in defined stimuli agent 
- `task/control/query.py` remove redundant best_selection in favor of one with constants. Implemented constants in return_stimuli methods.

## Test

- `make test-all`
- ran copy phrase with backspace always shown

## Documentation

### DIFF
--- a/bcipy/helpers/stimuli.py
+++ b/bcipy/helpers/stimuli.py
@@ -63,7 +63,7 @@ def alphabetize(stimuli: List[str]) -> List[str]:
 def rsvp_inq_generator(query: List[str],
                        timing: List[float] = [1, 0.2],
                        color: List[str] = ['red', 'white'],
-                       stim_number: int = 1,
+                       inquiry_count: int = 1,
                        stim_order: StimuliOrder = StimuliOrder.RANDOM,
                        is_txt: bool = True) -> InquirySchedule:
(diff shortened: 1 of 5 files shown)


bcipy/helpers/stimuli.py
```diff
@@ -182,12 +181,9 @@ def best_case_rsvp_inq_gen(alp: list,
     if inq_constants and not set(inq_constants).issubset(alp):
         raise BciPyCoreException('Inquiry constants must be alphabet items.')
 
-    # create a list of alphabet letters
-    alphabet = [i for i in alp]
-
```
lawhead, line 187: Sometimes in our codebase we do this to copy a list since it will be mutated by the method it's passed into, but it doesn't look like best_selection mutates its arguments. Maybe in a previous iteration? Anyway, nice cleanup.

lawhead (approved):

PR #184, opened 2021-11-11
Session data refinements

lawhead:
# Overview

Added some refinements to the session data (for Copy Phrase): limited the evidence precision for better readability; added the selected letter for each inquiry.

## Ticket

- https://www.pivotaltracker.com/story/show/180155037
- https://www.pivotaltracker.com/story/show/180154885

## Contributions

- Limited precision for evidence values in the session data to 5 decimal places.
- Added the selected character to each inquiry in the session data.

## Test

- Ran all unit tests.
- Ran the Copy Phrase task and examined the resulting session.json data to ensure that all evidence values are rounded appropriately and the selected value is correct. Looked at scenarios where a letter was spelled and also where backspace was used to delete a letter.

### DIFF
--- a/bcipy/task/data.py
+++ b/bcipy/task/data.py
@@ -130,13 +130,10 @@ class Inquiry:
             for name, value in data.items() if name.endswith(EVIDENCE_SUFFIX)
         }
 
-        computed = [
-            attr for attr in dir(cls)
-            if isinstance(getattr(cls, attr), property)
-        ]
         non_evidence_data = {
             name: value
-            for name, value in data.items() if name not in computed
+            for name, value in data.items()
+            if not name.endswith(EVIDENCE_SUFFIX)
         }
         inquiry = cls(**non_evidence_data)
         if len(data['stimuli']) == 1 and isinstance(data['stimuli'][0], list):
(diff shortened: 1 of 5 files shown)


bcipy/task/data.py
```diff
@@ -6,6 +6,21 @@
 EVIDENCE_SUFFIX = "_evidence"
 
 
+def rounded(values: List[float], precision: int) -> List[float]:
```
tab-cmd, line 9: The type signature for this is off if precision can be `None`. I think this may be more straightforward if we enforce int for precision.. either forcing the upstream callers to check whether it should be rounded at all or have a sufficient default value `precision: int == 10`.

bcipy/task/data.py
```diff
@@ -249,7 +270,7 @@ def as_dict(self,
             'task': self.task,
             'mode': self.mode,
             'series': series_dict,
-            'total_time_spent': self.total_time_spent,
+            'total_time_spent': round(self.total_time_spent, 2),
```
tab-cmd: I'd pull this up as a instance variable. `self.time_spent_precision = 2` or use the same as `self.precision`?

tab-cmd (approved):

PR #185, opened 2021-11-29
Trigger handling

theJokerEvoker:
# Overview

Created a new implementation for Trigger handling

## Ticket

[TriggerType](https://www.pivotaltracker.com/story/show/178294298)
[Trigger class](https://www.pivotaltracker.com/story/show/178875697)
[TriggerHandler](https://www.pivotaltracker.com/story/show/178875740)

## Contributions

triggers.py- Added Trigger, TriggerType, FlushSensitivity, TriggerHandler

## Test

- Added unit tests for main functionalities of Trigger and TriggerHandler
- Using TriggerHandler, input list of Triggers and check that it properly writes them to a text file on close()
- Using TriggerHandler, input a path to a txt file of Triggers and check that it properly loads as a list of Triggers, or throws exceptions if something goes wrong

## Documentation

- Added doc strings and in-line for Trigger, TriggerType, FlushSensitivity, TriggerHandler

### DIFF
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -785,17 +785,19 @@ class TriggerHandler:
 
         if exclusion:
             for type in exclusion:
-                for item in triggers_list:  # possible location of refactor
-                    triggers_list[:] = [item for item in triggers_list if not type.value == item[1]]
+                triggers_list[:] = [item for item in triggers_list if not type.value == item[1]]
 
         new_trigger_list = []
         for trigger in triggers_list:
-            new_trigger_list.append(
-                Trigger(trigger[0],
-                        TriggerType(trigger[1]),
-                        str(trigger[2])
-                        )
-            )
+            try:
+                new_trigger_list.append(
+                    Trigger(trigger[0],
+                            TriggerType(trigger[1]),
+                            str(trigger[2])
+                            )
+                )
+            except Exception as e:
+                raise BciPyCoreException(f'{trigger} read from {path} is not able to be cast')
 
         return new_trigger_list
 
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from dataclasses import dataclass
+# from dataclasses import dataclass
 from enum import Enum
 from typing import Dict, List, Optional, TextIO, Tuple
 
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -709,12 +709,6 @@ class TriggerHandler:
         self.write()
         self.file.close()
 
-    def __enter__(self):
-        return self
-
-    def __exit__(self, type, value, traceback):
-        self.close()
-
     def write(self):
         """
         Writes current Triggers in self.triggers[] to .txt file in self.file_name.
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -1,6 +1,6 @@
 import logging
 import os
-# from dataclasses import dataclass
+from dataclasses import dataclass
 from enum import Enum
 from typing import Dict, List, Optional, TextIO, Tuple
 
@@ -661,19 +661,15 @@ class TriggerCategory(Enum):
         return f'{self.value}'
 
 
-# @dataclass(frozen=True)
+@dataclass(frozen=True)
 class Trigger:
     """
     Object that encompasses data for a single trigger instance.
     """
 
-    def __init__(self,
-                 label: str,
-                 category: TriggerCategory,
-                 time: str):
-        self.label = label
-        self.category = category
-        self.time = time
+    label: str
+    category: TriggerCategory
+    time: float
 
     def __repr__(self):
         return f'Trigger: label=[{self.label}] category=[{self.category}] time=[{self.time}]'
@@ -799,7 +795,7 @@ class TriggerHandler:
                 new_trigger_list.append(
                     Trigger(trigger[0],
                             TriggerCategory(trigger[1]),
-                            str(trigger[2])
+                            float(trigger[2])
                             )
                 )
             except Exception as e:
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -641,7 +641,6 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
     return corrected
 
 
-
 class TriggerType(Enum):
     """
     Enum for the primary types of Triggers.
@@ -727,10 +726,23 @@ class TriggerHandler:
 
         self.triggers = []
 
-    def load(self,
-             path: str,
-             offset: Optional[float]=None,
-             exclusion: Optional[List[TriggerType]]=None) -> List[Trigger]:
+    @staticmethod
+    def read_text_file(path: str) -> List[str]:
+        if not path.endswith('.txt') or not os.path.exists(path):
+            raise FileNotFoundError(f"Valid triggers .txt file not found at [{path}]."
+                                    "\nPlease rerun program.")
+
+        triggers_list = []
+        with open(path) as raw_txt:
+            for line in raw_txt:
+                line_split = line.split()
+                triggers_list.append(line_split)
+        return triggers_list
+
+    @staticmethod
+    def load(path: str,
+             offset: Optional[float] = None,
+             exclusion: Optional[List[TriggerType]] = None) -> List[Trigger]:
         """
         Loads a list of triggers from a .txt of triggers.
 
@@ -758,15 +770,7 @@ class TriggerHandler:
         """
 
         # Checking for file with given path, with or without .txt
-        if not path.endswith('.txt') or not os.path.exists(path):
-            raise FileNotFoundError(f"Valid triggers .txt file not found at [{path}]."
-                                    "\nPlease rerun program.")
-
-        txt_list = []
-        with open(path) as raw_txt:
-            for line in raw_txt:
-                line_split = line.split()
-                txt_list.append(line_split)
+        triggers_list = TriggerHandler.read_text_file(path)
 
         if offset:
             # If there is exclusion but no offset,
@@ -775,23 +779,25 @@ class TriggerHandler:
             if isinstance(offset, list) and exclusion is None:
                 exclusion = offset
             else:
-                for item in txt_list:
+                for item in triggers_list:
                     time_float = float(item[2]) + offset
                     item[2] = str(time_float)
 
         if exclusion:
             for type in exclusion:
-                for item in txt_list:
-                    txt_list[:] = [item for item in txt_list if not type.value == item[1]]
+                for item in triggers_list:  # possible location of refactor
+                    triggers_list[:] = [item for item in triggers_list if not type.value == item[1]]
 
         new_trigger_list = []
-        for e in txt_list:
-            new_trigger_list.append(Trigger(e[0],
-                                    TriggerType(e[1]),
-                                    str(e[2])))
+        for trigger in triggers_list:
+            new_trigger_list.append(
+                Trigger(trigger[0],
+                        TriggerType(trigger[1]),
+                        str(trigger[2])
+                        )
+            )
 
         return new_trigger_list
-
 
     def add_triggers(self, triggers: List[Trigger]) -> List[Trigger]:
         """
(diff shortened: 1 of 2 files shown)


bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+
+class TriggerType(Enum):
+    """
+    Enum for the primary types of Triggers.
+    """
+
+    NONTARGET = "nontarget"
+    TARGET = "target"
+    FIXATION = "fixation"
+    PROMPT = "first_pres_target"
```
tab-cmd: This may be a good time to update this to prompt outright. It will make older data less compatible but it's more clear.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+    @classmethod
+    def list(cls):
+        """Returns all enum values as a list"""
+        return list(map(lambda c: c.value, cls))
+
+    def __str__(self):
+        return f'{self.value}'
+
+
+class Trigger:
```
lawhead, line 666: We should make `Trigger` immutable to retain the current behavior. It seems like a data class would work well here:

    from dataclasses import dataclass
    @dataclass(frozen=True)
    class Trigger:
        label: str
        type: TriggerType
        time: float

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+    """
+    Object that encompasses data for a single trigger instance.
+    """
+
+    def __init__(self,
+                 label: str,
+                 type: TriggerType,
+                 time: str):
```
lawhead: Shouldn't `time` be a `float`? Also, I don't know if it would be confusing to override the Python global for `type`, since `trigger.type` doesn't mean the same thing as `type(trigger)`.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+        self.label = label
+        self.type = type
+        self.time = time
+
+    def __repr__(self):
+        return f'Trigger: label=[{self.label}] type=[{self.type}] time=[{self.time}]'
+
+
+class FlushSensitivity(Enum):
```
lawhead: It seems like sensitivity refers to a relative level. Maybe a better term would be `FlushPeriod`, which indicates the interval of time between occurrences.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+            None
+        """
+
+        for trigger in self.triggers:
+            self.file.write(f'{trigger.label} {trigger.type.value} {trigger.time}\n')
+
+        self.triggers = []
+
+    @staticmethod
+    def read_text_file(path: str) -> List[str]:
```
lawhead: I think the return type should be `List[List[str]]` here.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+    """
+    Enum that defines how often list of Triggers will be written and dumped.
+    """
+
+    EVERY = "flush after every trigger addition"
+    END = "flush at end of session"
+
+
+class TriggerHandler:
+
```
lawhead, line 693: This class could use a docstring. Also, if the static methods are converted to module functions this class has a single responsibility and could be renamed `TriggerWriter`.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+            # This moves "offset" into its proper parameter of exclusion
+            if isinstance(offset, list) and exclusion is None:
+                exclusion = offset
+            else:
+                for item in triggers_list:
+                    time_float = float(item[2]) + offset
+                    item[2] = str(time_float)
+
+        if exclusion:
+            for type in exclusion:
```
lawhead: This variable could be renamed `trigger_type` so we're not unintentionally redefining the built-in `type` function.

bcipy/helpers/triggers.py
```diff
@@ -638,3 +639,185 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+            to be removed from the loaded trigger list.
+
+        Returns
+        -------
+            List of Triggers from loaded .txt file with desired modifications
+        """
+
+        # Checking for file with given path, with or without .txt
+        triggers_list = TriggerHandler.read_text_file(path)
+
```
lawhead: There are some aspects of the implementation here that I'm finding confusing. In the type signature we state that offset is either a `float` or `None`, so it's not clear why we are checking to see if it's a `list`. Maybe the intention is that if an offset is provided then we want to exclude the last entry which has a `TriggerType` of 'offset_correction', in which case we will need an additional value added to the `TriggerType` enum.

I think the more common scenario is that rather than applying a manually provided offset we want the option of using the offset recorded in the trigger file. The offset parameter would then be a `bool`, where a True value applies this offset and a False value leaves the trigger times unchanged.

bcipy/helpers/triggers.py
theJokerEvoker: I'm guessing this is an area we'll definitely have to go back and refactor carefully?

bcipy/helpers/triggers.py
theJokerEvoker: Is it enough to simply add the `@dataclass(frozen=True)` before the class definition as such (after importing dataclass, of course):
```
@dataclass(frozen=True)
class Trigger:
    """
    Object that encompasses data for a single trigger instance.
    """

    def __init__(self,
                 label: str,
                 type: TriggerType,
                 time: str):
        self.label = label
        self.type = type
        self.time = time
```
Or do I have to add the class attributes individually, as you wrote them? I haven't worked with this functionality before, so I'm unfamiliar with how it works.

bcipy/helpers/triggers.py
theJokerEvoker: Perhaps a rename to `FlushFrequency` might also work?

bcipy/helpers/triggers.py
theJokerEvoker: I don't know if it's too unwieldy or makes enough sense, but maybe something like `category` would also work? I'm not sure  if, in this context of possible confusion with built-in `type`,  `trigger.category` would be distinct enough from `trigger.type`. I think both possibilities would require a bit of refactoring with the `Enum` and such as well, though.

bcipy/helpers/triggers.py
theJokerEvoker: Added a short docstring, will discuss the renaming as well

bcipy/helpers/triggers.py
theJokerEvoker: I think I initially decided to make `time` a `str` because I wasn't sure if it was going to cause problems down the line with casting time back and forth between `str` and `float`, looking back I guess it does make more sense as just a regular `float`. Do you think this will be much of an issue, for usability or otherwise?

bcipy/helpers/triggers.py
theJokerEvoker: If I'm understanding correctly, are you referring to these few lines in the `load` method?
```
if offset:
            if isinstance(offset, list) and exclusion is None:
                exclusion = offset
            else:
                for item in triggers_list:
                    time_float = float(item[2]) + offset
                    item[2] = str(time_float)
```
If you are, the reason I ended up having to do it like this was because if the `load` method was given an `exclusion` but no `offset`, the method would try to read in the `exclusion` as an `offset` due to the ordering of the parameters, which would fail and throw an error, since it was expecting a `float` but receiving a `list`. I couldn't figure out how to ignore the `offset` if it was `None` while an `exclusion` was present, so I ended up having to just do it like this.

In regards to the latter scenario you mentioned, I was unaware that this is how the offset is usually handled-- I can definitely change the handler and add a `TriggerCategory Enum` to fit this if needed.

bcipy/helpers/triggers.py
theJokerEvoker: Also, I ran the `test_triggers.py` file after making these changes, and it looks like freezing the class is causing issues with the unit tests. Given my unfamiliarity, I'm going to remove the immutability feature for the time being, if that's alright.

lawhead: I'm okay leaving this to later, but if you're interested the Python documentation (https://docs.python.org/3/library/dataclasses.html) has some good information on data classes. I think the errors you were getting in the tests were caused by having an `init` method. If you define the class like I have in my comment above things should work, with the exception of the `test_print_trigger`, since dataclasses define their own `__repr__`, which is similar but not exactly like the definition you provided.

bcipy/helpers/triggers.py
lawhead: I think `float` is preferred for usability.

bcipy/helpers/triggers.py
lawhead: I don't have a strong preference on the variable name. Maybe @sci-tab has an opinion here?

bcipy/helpers/triggers.py
lawhead: It seems like if we are going to use an offset it will always be the one provided in the triggers.txt file that we're reading, so I'm not sure it makes sense to have that as a parameter. You can check out the `read_triggers` function to see how this is handled (basically, we read all records in the file and `pop()` off the last result).  @sci-tab, do you have thoughts on this?

bcipy/helpers/triggers.py
theJokerEvoker: Looks like it worked, thanks! The `__repr__` looks like it still works, based on the documentation it looks like the current `__repr__` just overrides the `__repr__()` the `dataclass` would create.

bcipy/helpers/triggers.py
tab-cmd: I think this PR will require enough refactoring and thought we can take some liberty.. let's make it "prompt"

bcipy/helpers/triggers.py
tab-cmd: We may have an offset or offsets provided in the triggers.txt... We should support loading / applying those optionally here. We also apply static offsets which are factored in outside of the triggers.txt file. @theJokerEvoker next time we meet let's discuss how to address this.

bcipy/helpers/triggers.py
tab-cmd: No strong preference. I think `trigger_type` is descriptive enough here

tab-cmd (approved):

bcipy/helpers/triggers.py
theJokerEvoker: @lawhead Tab and I have talked about it and made some adjustments to this issue-- we've made it so that the `read_text_file()` method now expects the first line of the `.txt` to be a Trigger that contains a "system offset" value, which will then be eventually returned to `load()` to be used. This assumption about the "system offset" will be implemented in [this ticket](https://www.pivotaltracker.com/story/show/179956988). We still maintain the ability to include a manual offset with the `load()`, however, though it is by default 0. Would you mind taking another look at the changes?

bcipy/helpers/triggers.py
```diff
@@ -638,3 +640,195 @@ def read_triggers(triggers_file: TextIO) -> List[Tuple[str, str, float]]:
...
+        triggers_list, system_offset = TriggerHandler.read_text_file(path)
+
+        try:
+            if exclusion:
+                for trigger_type in exclusion:
+                    triggers_list[:] = [item for item in triggers_list if not trigger_type.value == item[1]]
+
+            # apply system and provided offsets
+            for item in triggers_list:
+                item[2] = float(item[2]) + offset + system_offset
```
lawhead, line 796: One thing to note is that we may not always want to apply the system_offset. In the recent refactor of the acquisition code the timestamps given in the triggers.txt file are using the same clock as those in the raw data, so if we want to relate the two we don't want to apply the offset. But maybe that distinction should be handled elsewhere.

bcipy/helpers/triggers.py
lawhead: @theJokerEvoker, thanks for your work on this. This version is a lot clearer for me.

lawhead (approved):

PR #186, opened 2021-12-08
Session Metrics

lawhead:
# Overview

This PR includes additional metrics to the session data output by the system to allow investigators to more accurately assess user performance.

## Tickets

* [Consistent Session Metrics](https://www.pivotaltracker.com/story/show/179649835)
* [Custom session metrics for Copy Phrase and Inquiry Preview](https://www.pivotaltracker.com/story/show/179806994)

## Contributions

- Added calculated values to Session data that compute some overall metrics.
- Added optional field in the Session data for `task_summary` data.
- Added custom metrics for the Copy Phrase / Inquiry preview tasks.

## Test

- Created additional unit tests and ran them.
- Ran the Copy Phrase task and inspected the resulting session.json data to ensure that values are being correctly calculated.

### DIFF
--- a/bcipy/task/paradigm/rsvp/copy_phrase.py
+++ b/bcipy/task/paradigm/rsvp/copy_phrase.py
@@ -2,9 +2,8 @@ from typing import List, NamedTuple, Optional, TextIO, Tuple
 
 from psychopy import core
 
-from bcipy.display import (InformationProperties,
-                           PreviewInquiryProperties, StimuliProperties,
-                           TaskDisplayProperties)
+from bcipy.display import (InformationProperties, PreviewInquiryProperties,
+                           StimuliProperties, TaskDisplayProperties)
 from bcipy.display.rsvp.mode.copy_phrase import CopyPhraseDisplay
 from bcipy.feedback.visual.visual_feedback import VisualFeedback
 from bcipy.helpers.clock import Clock
@@ -77,8 +76,9 @@ class RSVPCopyPhraseTask(Task):
         'feedback_pos_y', 'feedback_stim_height', 'feedback_stim_width',
         'filter_high', 'filter_low', 'filter_order', 'fixation_color',
         'info_color', 'info_font', 'info_height', 'info_text', 'is_txt_stim',
-        'lm_backspace_prob', 'max_inq_len', 'max_inq_per_series', 'max_minutes',
-        'max_selections', 'min_inq_len', 'notch_filter_frequency', 'preview_inquiry_isi',
+        'lm_backspace_prob', 'max_inq_len', 'max_inq_per_series',
+        'max_minutes', 'max_selections', 'min_inq_len',
+        'notch_filter_frequency', 'preview_inquiry_isi',
         'preview_inquiry_key_input', 'preview_inquiry_length',
         'preview_inquiry_progress_method', 'session_file_name',
         'show_feedback', 'show_preview_inquiry', 'spelled_letters_count',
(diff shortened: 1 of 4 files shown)


bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -640,8 +643,77 @@ def name(self) -> str:
...
+    def as_dict(self) -> dict:
+        """Computes the task summary data to append to the session."""
+
+        selections = [
+            inq for inq in self.session.all_inquiries if inq.selection
+        ]
+        correct = [inq for inq in selections if inq.is_correct_decision]
+        incorrect = [inq for inq in selections if not inq.is_correct_decision]
+
+        # TODO: what about SPACE?
```
tab-cmd: We should bring this up to the group.. I'm not sure we'd want to separate backspace in our task summary.

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -640,8 +643,77 @@ def name(self) -> str:
...
+        correct_letters = [
```
tab-cmd: replace letters with `symbols`

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -640,8 +643,77 @@ def name(self) -> str:
...
+        switch_per_selection = (btn_presses /
+                                sel_count) if sel_count > 0 else 0
+        accuracy = (len(correct) / sel_count) if sel_count > 0 else 0
+
+        # Note that minutes includes startup time and any breaks.
+        minutes = self.session.total_time_spent / 60
+        return {
+            'selections_correct': len(correct),
+            'selections_incorrect': len(incorrect),
+            'selections_correct_letters': len(correct_letters),
```
tab-cmd: `selections_correct_symbols`? ... Is there value in distinguishing backspace? I feel like we could derive that with the rest of the metrics and not hold us to that hardcoded symbol

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -640,8 +643,77 @@ def name(self) -> str:
...
+            'selections_correct_backspace': len(correct_backspace),
+            'switch_response_time': None,  # TODO: parameter?
```
tab-cmd: we can derive this from the triggers we get back from the display.. it would be inquiry_preview - key_press. Would this be average_response_time?

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -640,8 +643,77 @@ def name(self) -> str:
...
+            'correct_rate': len(correct) / minutes if minutes else 0,
+            'copy_rate': len(correct_letters) / minutes if minutes else 0
+        }
+
+    def btn_press_count(self) -> int:
+        """Compute the number of times the switch was activated. Returns 0 if
+        inquiry preview mode was off or mode was preview-only."""
+
+        show_preview = self.parameters['show_preview_inquiry']
+        preview_mode = self.parameters['preview_inquiry_progress_method']
```
tab-cmd: If these are the only two parameters we need, we could ask they be provided explicitly?

tab-cmd (approved):

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: Ah, okay. That's why that ticket was listed as a blocker. I'll address this parameter in a followup PR.

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: Actually, this also makes me realize that I should use the backspace character configured in the parameters.

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: I agree. I think we should leave out that one.

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: I pulled this out in a separate ticket: https://www.pivotaltracker.com/story/show/180605653

PR #187, opened 2021-12-20
Switch response metric

lawhead:
# Overview

Added an Inquiry Preview session metric for switch response time.

## Ticket

https://www.pivotaltracker.com/story/show/180605653

## Contributions

- Added trigger types for triggers used in the Inquiry Preview / Copy Phrase task.
- Added new metric that computes the average switch response time in seconds.
- Some refactoring of the trigger helpers to correctly load existing triggers.txt files. 
- Unit tests for new functionality

## Test

- Ran all unit tests.
- Ran copy phrase task to ensure that all session data was correctly written.
- Tested the switch_response_time calculations against previously generated triggers.txt files using different `preview_inquiry_progress_method` values.


bcipy/helpers/triggers.py
```diff
@@ -651,7 +651,10 @@ class TriggerType(Enum):
     TARGET = "target"
     FIXATION = "fixation"
     PROMPT = "prompt"
-    SYSTEM = "offset"
+    SYSTEM = "offset_correction"
```
tab-cmd: I have a PR to change how we write and read system offsets. It will be as implemented (with a single offset written). I think it'd be best to wait for that PR. I should be done this week.

bcipy/helpers/triggers.py
```diff
@@ -651,7 +651,10 @@ class TriggerType(Enum):
...
+    CALIBRATION = "calib"
+    KEY_PRESS = "key_press"
```
tab-cmd: I'd be in favor of a more generic `EVENT` or similar and having the label describe what it is... key press, mouse click, etc. What do you think? It would constitute a change in the task as well. I'm happy to address that in another PR if you're okay with this approach.

bcipy/helpers/triggers.py
```diff
@@ -784,33 +793,13 @@ def load(path: str,
...
-                            float(trigger[2])
-                            )
-                )
-            except Exception as e:
-                raise BciPyCoreException(f'{trigger} read from {path} is not able to be cast')
-
-        return new_trigger_list
+        triggers, system_offset = TriggerHandler.read_text_file(path)
+        excluded_types = exclusion or []
+        total_offset = offset + system_offset
```
tab-cmd, line 314: I like how succinct this is, however the custom exceptions were helpful and intentional. I'd be okay changing the logic flow but we need some error handling with better messaging.

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -715,6 +722,55 @@ def btn_press_count(self) -> int:
...
+        """Computes the average switch response in seconds."""
+
+        # Remove consecutive items with the same type; we are only interested
+        # in PREVIEW followed by a KEY_PRESS.
+        triggers = destutter(self.switch_triggers(), key=lambda trg: trg.type)
+        pairs = list(zip(triggers[::2], triggers[1::2]))
+
+        # Confirm that the data is structured as expected.
+        for preview, keypress in pairs:
+            valid = preview.type == TriggerType.PREVIEW and keypress.type == TriggerType.KEY_PRESS
```
tab-cmd: This is hard to follow visually.. maybe

```
valid = (preview.type == TriggerType.PREVIEW) and (keypress.type == TriggerType.KEY_PRESS)
```

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -715,6 +722,55 @@ def btn_press_count(self) -> int:
...
+        if not self.trigger_path:
+            return []
+        triggers, _offset = TriggerHandler.read_text_file(self.trigger_path)
+        return [
+            trg for trg in triggers
+            if trg.type in [TriggerType.PREVIEW, TriggerType.KEY_PRESS]
+        ]
+
+
+def destutter(items: List, key: Callable = lambda x: x) -> List:
```
tab-cmd: This should live somewhere else... helpers? I can suggest something if one doesn't come to mind.

tab-cmd (changes requested):

bcipy/helpers/triggers.py
lawhead: Okay, sounds good. If possible, I would like to support any previously used trigger types.

bcipy/helpers/triggers.py
lawhead: The custom exceptions that we previously had here were swallowing some of the errors I was encountering which made it hard to debug. I went ahead and added them back (see the `read_text_file` method) with some additional details to hopefully make debugging easier.

bcipy/helpers/triggers.py
lawhead: I added the `EVENT` prefix to the keypress for now, but I think this is not quite what you had in mind? Let's discuss in our next meeting and create a followup task for this.

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: This function seemed a bit too generic for any of our existing helpers so I went ahead and created a new one. If you would rather see it elsewhere let me know.

bcipy/helpers/triggers.py
tab-cmd: We discussed this a bit in the Trigger PR but the implications are pretty far reaching. As of 2.0, we discussed dropping support for these older offset correction methods. If requested, we could provide a helper to convert older data to this newer format. This would require reading in the triggers.txt and writing a single offset value to the triggers file.

bcipy/helpers/list_processing.py
```diff
@@ -0,0 +1,20 @@
+"""Utility functions for list processing."""
```
tab-cmd: Would `helpers/list.py` work? It might make the module generic enough to pull other operations on lists here.

tab-cmd (approved):

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -744,6 +751,37 @@ def btn_press_count(self) -> int:
...
+        return sum(response_times) / count if count > 0 else None
+
+    def switch_triggers(self) -> List[Trigger]:
+        """Returns a list of switch-related triggers"""
+        if not self.trigger_path:
+            return []
+        triggers, _offset = TriggerHandler.read_text_file(self.trigger_path)
+        return [
+            trg for trg in triggers
+            if trg.type in [TriggerType.PREVIEW, TriggerType.EVENT]
```
lawhead, line 782: This is currently working and gives us the appropriate triggers to use for the switch response annotation, but if we add any other triggers that use `TriggerType.EVENT` we need to remember to make the changes here to look for the symbol of interest.

PR #188, opened 2022-01-08
Uniform Language Model

lawhead:
# Overview

Created a new UniformLanguageModel that uses the new LanguageModel base class. Refactored the existing code base to use the new model. 

## Ticket

https://www.pivotaltracker.com/story/show/178695693

## Contributions

- Created a new UniformLanguageModel class that provides equal probabilities for each symbol.
- Integrated the new LanguageModel API:
    - Refactored bci_main to ensure that a language model is always created for spelling tasks, using the uniform model by default.
    - Refactored language_model helper to simplify the creation of language models.
    - Added a `predict` model to the `PRELM` language model that can take the spelled text as a parameter. This method resets the server after every call.
    - Refactored CopyPhraseWrapper to call the `predict` method; removed the extraneous code for when no LM is present.

## Test

- Ensured that all unit tests pass.
- Ran Copy Phrase Task with no model selected to ensure the UniformModel is used by default and that parameters were correctly passed.
- Ran Copy Phrase Task with the `PRELM` and confirmed that this model is correctly configured.


bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -286,60 +290,47 @@ def initialize_series(self) -> Tuple[bool, InquirySchedule]:
...
+
+            # update the lmodel and get back the priors
+            lm_letter_prior = self.lmodel.predict(update)
+
+            # normalize to probability domain if needed
+            normalized = getattr(self.lmodel, 'normalized', False)
+            if not normalized:
+                lm_letter_prior = norm_domain(lm_letter_prior)
+
+            if BACKSPACE_CHAR in self.alp:
```
nik-sm, line 308: should it be `if BACKSPACE_CHAR not in self.alp`?

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -286,60 +290,47 @@ def initialize_series(self) -> Tuple[bool, InquirySchedule]:
...
+                # Append backspace if missing.
+                sym = (BACKSPACE_CHAR, self.backspace_prob)
+                lm_letter_prior = sym_appended(lm_letter_prior, sym)
+
+            # convert to format needed for evidence fusion;
+            # probability value only in alphabet order.
+            prior = [
```
nik-sm, line 315: Is it necessary to sort/subset here? Seems like the language model should provide the correct set of letters in the correct order, and that could be covered by a unit test.

On the other hand, doesn't cost much to double check...

bcipy/helpers/language_model.py
```diff
@@ -21,15 +17,19 @@ def init_language_model(parameters):
...
+
+    # introspect the model arguments to determine what parameters to pass.
+    args = inspect.signature(model).parameters.keys()
 
-    port = int(parameters['lang_model_server_port'])
-    lmodel = LangModel(lmtype, logfile="lmwrap.log", port=port)
-    lmodel.init()
-    return lmodel
+    # select the relevant parameters into a dict.
+    params = {key: parameters[key] for key in args & parameters.keys()}
```
nik-sm, line 31: Is is possible for there to be a missing arg? (Something expected by `args` that is not in `parameters.keys()`)

nik-sm: Taking the intersection made me think that they are non-overlapping sets like this:
```
{--args--}
      {--parameters--}
```

If it's guaranteed that all of `args` will find their values in `parameters`, you could just do: `{key: parameters[key] for key in args}` to be more explicit

bcipy/language/tests/test_uniform.py
```diff
@@ -0,0 +1,55 @@
...
+    def test_predict_with_fixed_backspace(self):
+        """Test predict with a fixed backspace probability."""
+        lmodel = UniformLanguageModel(lm_backspace_prob=0.05)
+        symbol_probs = lmodel.predict(evidence=[])
+        self.assertTrue((BACKSPACE_CHAR, 0.05) in symbol_probs)
+
+        probs = [prob for sym, prob in symbol_probs]
+        self.assertEqual(len(set(probs)), 2,
+                         "Backspace probability should be different")
+        self.assertTrue(0.05 in probs)
```
nik-sm, line 54: Is `BACKSPACE_CHAR` at a known position in the list? You could check that `0.05` is at that expected position.

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -97,6 +97,10 @@ def __init__(self,
                  stim_length: int = 10,
                  stim_order: StimuliOrder = StimuliOrder.RANDOM):
 
+        if not lmodel:
+            assert is_txt_stim, "Language model is required."
```
nik-sm: Here it seems like `UniformLanguageModel` will only work with text, but inside that class it seems like it can handle either text or img. (Maybe I'm misreading...)

bcipy/language/uniform.py
```diff
@@ -0,0 +1,92 @@
...
+        list of probabilities (floats)
+    """
+    n_letters = len(alphabet)
+    if not specified:
+        return np.full(n_letters, 1 / n_letters)
+
+    # copy specified dict ignoring non-alphabet items
+    overrides = {k: specified[k] for k in alphabet if k in specified}
+
+    prob = (1 - sum(overrides.values())) / (n_letters - len(overrides))
```
nik-sm, line 91: Maybe should include `assert sum(overrides.values()) < 1`

nik-sm (changes requested):
Few minor comments - otherwise looks good!

bcipy/helpers/copy_phrase_wrapper.py
lawhead: We don't currently have any assertions in place that the alphabet supported by the language model is the same one, symbol for symbol, used in the task. Now that our team structure is more integrated and we have more control over the language models in use maybe we can be stricter about the expected inputs and not have so many checks in place here.

In this case we saying that if the the task alphabet has a backspace character, ensure that the resulting probabilities from the language model also have a value and insert one if not.

bcipy/helpers/copy_phrase_wrapper.py
lawhead: See my comment above. Ideally the language model does send us exactly what we need. In the past I believe there may have been some extra characters. But now that we're starting with a clean slate for the language model we should do another refactor to make some stronger guarantees.

bcipy/helpers/language_model.py
lawhead: Some of the language models have extra parameters not provided by the parameters.json file. In particular I think there was one for log file name.

bcipy/helpers/copy_phrase_wrapper.py
lawhead: This is probably a bit confusing. The `UniformLanguageModel` can support both text or images. If the symbol set is text we know how to construct the alphabet, but if it's images we need more information that's not available at this point (the path to the images, etc.). It might be simpler to just always require a language model.

bcipy/helpers/copy_phrase_wrapper.py
nik-sm: I see - so `sym_appended()` will append backspace if needed or else do nothing?

nik-sm (approved):

bcipy/helpers/language_model.py
tab-cmd: I think we can deprecate those other parameters (ports) and support for its own logfile in a follow up ticket. I don't think it makes sense for there to be a separate log file anymore.

lawhead: Agreed

bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -97,6 +97,13 @@ def __init__(self,
                  stim_length: int = 10,
                  stim_order: StimuliOrder = StimuliOrder.RANDOM):
 
+        if not lmodel:
```
tab-cmd, line 100: I think we can safely assert this is defined upstream and avoid erroring on image stimuli here? We could then update the type signature to `LanguageModel` and require it as input. What do you think?

tab-cmd (approved):

bcipy/language/tests/test_uniform.py
sliu126: The position in symbol_probs should not matter.

bcipy/language/uniform.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns:
+    --------
+        list of probabilities (floats)
+    """
+    n_letters = len(alphabet)
+    if not specified:
+        return np.full(n_letters, 1 / n_letters)
+
+    # copy specified dict ignoring non-alphabet items
+    overrides = {k: specified[k] for k in alphabet if k in specified}
```
sliu126, line 88: I think here overrides is the same thing as specified?

bcipy/language/uniform.py
```diff
@@ -0,0 +1,93 @@
...
+            overrides = {BACKSPACE_CHAR: self.backspace_prob}
+        probs = equally_probable(self.symbol_set, overrides)
+        return list(zip(self.symbol_set, probs))
+
+    def update(self) -> None:
+        """Update the model state"""
+
+    def load(self, path: Path) -> None:
+        """Restore model state from the provided checkpoint"""
+
```
sliu126, line 66: We may not need to add these empty methods here...

sliu126 (approved):

bcipy/language/uniform.py
tab-cmd: These are required via the base class `LanguageModel` definition!

bcipy/language/uniform.py
lawhead: This protects against injecting symbols that are not in the alphabet. So it's close, but not exactly the same.

bcipy/language/uniform.py
lawhead: If these are consistently unused we may want to revisit the API at some point.

PR #189, opened 2022-01-10
TriggerHandler Implementation

tab-cmd:
# Overview

Implemented `TriggerHandler` in BciPy. This required refactoring all instances of reading and writing trigger data in the codebase. Several tasks and scripts were removed to avoid refactoring unused code.
 
## Ticket

- https://www.pivotaltracker.com/story/show/179956988

### Updated

- `helpers/convert.py`: to use refactored `trigger_decoder`
- `helpers/triggers.py`: refactored `trigger_decoder` to use `TriggerHandler` . Added several TriggerTypes to support tasks. Remove old trigger writing methods. 
- `offline_analysis.py`: to use refactored `trigger_decoder`
- `paradigm/rsvp/calibration/calibration.py`: to use `TriggerHandler` and determine targetness
- `paradigm/rsvp/copy_phrase.py`: to use `TriggerHandler` and determine targetness

### Removed

- `paradigm/rsvp/calibration/inter_inquiry_feedback_calibration.py`
- `paradigm/rsvp/calibration/alert_tone_calibration.py` 
- `scripts/python/feedback_logs.py`
- `scripts/python/offset_viewer.py`
- `scripts/python/timing_tools.py`

## Test

- `make test-all`, `bcipy` --> ensure all triggers wrote correctly to triggers.txt, `bcipy --task "RSVP Copy Phrase"` --> ensure all triggers wrote correctly to triggers.txt.

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Added docstrings to new methods, updated READMEs and existing docstrings where applicable.


bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -93,6 +95,36 @@ def generate_stimuli(self):
...
+    def convert_timing_triggers(self, timing: List[tuple]) -> List[Trigger]:
+        """Convert Stimuli Times to Triggers.
+
+        Using the stimuli presentation times provided by the display, convert them into BciPy Triggers.
+        """
+        triggers = []
+        x = 0
+
+        # loop over timing and assign the correct TriggerType
+        for letter, time in timing:
```
lawhead: How about:
    `for i, (letter, time) in enumerate(timing):`

bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -93,6 +95,36 @@ def generate_stimuli(self):
                                              is_txt=self.rsvp.is_txt_stim,
                                              color=self.color)
 
+    def convert_timing_triggers(self, timing: List[tuple]) -> List[Trigger]:
```
lawhead: This method could be a function, which would make it easier to add some tests for it.

bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -93,6 +95,36 @@ def generate_stimuli(self):
...
+        x = 0
+
+        # loop over timing and assign the correct TriggerType
+        for letter, time in timing:
+            if x == 0:
+                targetness = TriggerType.PROMPT
+                target_letter = letter
+            elif x == 1:
+                targetness = TriggerType.FIXATION
+            elif x > 1 and target_letter == letter:
```
lawhead: What do you think about initializing the `target_letter` outside of the loop?
    `target_letter = timing[0][1] if timing else None`

I found it a bit confusing that it's initialized within a branch and used in the `else` condition, but that's probably my bias showing up from other languages where scoping rules would prevent this.

bcipy/helpers/triggers.py
```diff
@@ -121,197 +133,7 @@ def _calibration_trigger(experiment_clock: Clock,
...
-        elif icon == target:
-            targetness = 'target'
-        else:
-            targetness = 'nontarget'
-        trigger_file.write('%s %s %s' % (icon, targetness, presentation_time) +
-                           "\n")
-
-
-def trigger_decoder(mode: str, trigger_path: str = None, remove_pre_fixation=True) -> tuple:
+def trigger_decoder(trigger_path: str, remove_pre_fixation=True, offset=0, exclusion=[]) -> tuple:
```
lawhead: The diff is hard to read, but I think the documentation no longer matches the parameters.

bcipy/helpers/triggers.py
```diff
@@ -694,17 +229,20 @@ class TriggerHandler:
...
         self.triggers = []
 
         if os.path.exists(self.file_name):
             raise Exception(f"[{self.file_name}] already exists, any writing "
                             "will overwrite data in the existing file.")
 
-        self.file = open(self.file_name, 'w+')
+        self.file = open(f'{self.path}/{self.file_name}', 'w+')
+
+    def __exit__(self):
```
lawhead: The `__exit__` method needs some additional parameters:
   `def __exit__(self, exc_type, exc_value, traceback)`

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -129,9 +129,9 @@ def __init__(self, win, daq, parameters, file_save, signal_model,
                                               self.static_clock, self.experiment_clock)
         self.file_save = file_save
 
-        self.trigger_save_location = f"{self.file_save}/{parameters['trigger_file_name']}"
+        self.trigger_handler = TriggerHandler(self.file_save, parameters['trigger_file_name'], FlushFrequency.EVERY)
         self.session_save_location = f"{self.file_save}/{parameters['session_file_name']}"
-
+        self.first_run = True
```
lawhead: `first_run` can be a computed property from `inq_counter`.
```
@property
def first_run(self) -> bool:
   return self.inq_counter == 0
```

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -612,33 +611,75 @@ def write_session_data(self) -> None:
...
+            # write offset first
+            self.trigger_handler.add_triggers(
+                [Trigger(
+                    'starting_offset',
+                    TriggerType.OFFSET,
+                    # offset will factor in true offset and time relative from beginning
+                    (self.daq.offset(self.rsvp.first_stim_time) - self.rsvp.first_stim_time)
+                )]
+            )
+            self.first_run = False
```
lawhead: With `first_run` as a computed property, this update should be removed.

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -612,33 +611,75 @@ def write_session_data(self) -> None:
...
+
+        triggers = self.convert_timing_triggers(stim_times, target_stimuli)
+        self.trigger_handler.add_triggers(triggers)
+
+    def convert_timing_triggers(self, timing: List[tuple], target_stimuli: str) -> List[Trigger]:
```
lawhead: This could also be made a function.

lawhead: What do you think about splitting this into two functions:

```
def trigger_type(symbol: str, target_stimuli: str) -> TriggerType:
    if symbol == 'inquiry_preview':
        return TriggerType.PREVIEW
    if 'bcipy_key_press' in symbol:
        return TriggerType.EVENT
    if symbol == '+':
        return TriggerType.FIXATION
    if target_stimuli == symbol:
        return TriggerType.TARGET
    return TriggerType.NONTARGET

def convert_timing_triggers(timing: List[tuple], target_stimuli: str) -> List[Trigger]:
    """Convert Stimuli Times to Triggers.

    Using the stimuli presentation times provided by the display, convert them into BciPy Triggers.
    """
    return [
        Trigger(symbol, trigger_type(symbol, target_stimuli), time)
        for symbol, time in timing
    ]
```

scripts/python/offset_viewer.py
```diff
@@ -1,191 +0,0 @@
-"""GUI tool to visualize offset analysis when analyzing system latency."""
```
lawhead, line 1: Did these scripts get relocated somewhere, or are they no longer used?

bcipy/task/paradigm/rsvp/calibration/calibration.py
tab-cmd: I'm not opposed to doing this. I just copied the same logic we used before to avoid any unexpected changes.

bcipy/task/paradigm/rsvp/copy_phrase.py
tab-cmd: I think this is a nice suggestion

scripts/python/offset_viewer.py
tab-cmd: The functionality will be moved into the repository itself - in the short run we'll be without it. There are tickets to support offsets and photodiodes in the backlog.

bcipy/task/paradigm/rsvp/calibration/calibration.py
tab-cmd: This seems like something all rsvp tasks may need and with slightly different trigger typing logic... I can add tests to it here too! How were you thinking of pulling it out?

lawhead: I was thinking it could be a function in this same file since it doesn't need anything specific from the class (property or other method call) for doing the conversion.

In looking at the difference between this function and the one used by CopyPhrase, the unique thing is how we determine a trigger type (given a symbol, target, and position), so maybe that's the task specific code.

bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -166,31 +192,46 @@ def execute(self):
...
+                    TriggerType.OFFSET,
+                    # offset will factor in true offset and time relative from beginning
+                    (self.daq.offset(self.rsvp.first_stim_time) - self.rsvp.first_stim_time)
+                )]
+            )
+
+        # make sure triggers are written for the inquiry
+        self.trigger_handler.add_triggers(self.convert_timing_triggers(timing))
+
+    def write_offset_trigger(self) -> None:
```
lawhead, line 209: Is this method used? It looks like it was replaced with `write_trigger_data`.

bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -166,31 +192,46 @@ def execute(self):
...
 
+    def write_trigger_data(self, timing: List[Tuple[str, float]], first_run) -> None:
+        # write offsets. currently, we only check for offsets at the beginning.
+        if self.daq.is_calibrated and first_run:
+            self.trigger_handler.add_triggers(
+                [Trigger(
+                    'starting_offset',
+                    TriggerType.OFFSET,
+                    # offset will factor in true offset and time relative from beginning
+                    (self.daq.offset(self.rsvp.first_stim_time) - self.rsvp.first_stim_time)
```
lawhead, line 202: I'm still trying to determine if this is correct. The data acquisition offset is the number of seconds elapsed from when recording started to the first stim and that's what was previously recorded as our offset in the triggers.txt file. This seems like a different approach and is maybe used differently? It seems like we are lacking documentation about how we're using offsets and when we use trigger times vs sample numbers.

tab-cmd: I think it's the same when applied... I did train a model with the values exported like this with no change in AUC. Previously, we loaded two values for this correction; we subtracted the calibration trigger from the offset and used that to transform all other timestamps. Now, we write the full offset value which would give us time since start for all values and do not write the calibration trigger. I'm not sure where the best place to document this is. I can add more detail to the TriggerHandler or start a section in the main README for timing / loading of data? Or both if helpful!

lawhead: Thanks for the explanation. I am more familiar with how we are using Copy Phrase triggers than Calibration. I think we can work together on the documentation. Not a blocker for this ticket.

bcipy/task/paradigm/rsvp/calibration/calibration.py
tab-cmd: They are both used! I still write the offset only value in case someone did want the LSL timestamps... not sure if this is helpful or not. It's filtered out as a TriggerType.SYSTEM when it loads by default.

lawhead (approved):

celikbasak (approved):

nik-sm (approved):

PR #192, opened 2022-01-14
Matrix calibration

jcgangemi1:
# Overview

Updates to matrix calibration to generate unique stimuli for each inquiry.  Refactored rsvp_inq_generator to be used for general inq_generator.  

## Ticket

https://www.pivotaltracker.com/story/show/179031773
https://www.pivotaltracker.com/story/show/179031810

## Contributions

- bcipy/helpers/stimuli.py: changed rsvp_inq_generator to inq_generator so that it may be used for matrix as well
- bcipy/task/paradigm/matrix/calibration.py: update to generate_stimuli
- bcipy/display/matrix/display.py: update prompt_target function to cut off extra target and fixation
- bcipy/task/control/handler.py: updated to the general inq_generator

## Test

- bcipy --task "Matrix Calibration" --> ensure all targets and stimuli were unique and the inquiry was displayed correctly

## Documentation

- Added comments where aplicable.  No major updates.


bcipy/helpers/stimuli.py
```diff
@@ -249,7 +249,7 @@ def calibration_inquiry_generator(
         idx = np.random.permutation(np.array(list(range(len_alp))))
         rand_smp = (idx[0:stim_length])
 
-        # define the target and fixation that come before an inquiry
+        # # define the target and fixation that come before an inquiry
```
lawhead: Looks like there is an extra symbol here.

bcipy/display/paradigm/rsvp/display.py
```diff
@@ -366,7 +366,7 @@ def wait_screen(self, message: str, message_color: str) -> None:
 
         Args:
             message(string): message to be displayed while waiting
-            message_color(string): color of the message to be displayed
+            message_`1  `color(string): color of the message to be displayed
```
lawhead: Is this a typo?

bcipy/task/paradigm/matrix/calibration.py
```diff
@@ -0,0 +1,269 @@
+from typing import List, Tuple
```
lawhead, line 1: This class has a LOT of duplication with the RSVP calibration. Now that we have a working task, it would be worth trying to factor out the commonality and determine what is actually unique to matrix. Is it just the display? Maybe we have a parent Calibration class that both RSVP and Matrix inherit from and they override the necessary method. Or maybe the Calibration task takes a display object as a parameter.

bcipy/task/paradigm/matrix/calibration.py
```diff
@@ -0,0 +1,269 @@
...
+        self.color = [parameters['stim_color']]
+
+        self.task_info_color = parameters['task_color']
+
+        self.stimuli_height = parameters['stim_height']
+
+        self.is_txt_stim = parameters['is_txt_stim']
+        self.eeg_buffer = parameters['eeg_buffer_len']
+
+        self.enable_breaks = parameters['enable_breaks']
```
lawhead, line 81: We are already setting `parameters` as an instance variable. Rather than having a bunch of additional instance variables that just wrap these values, my preference is just to use `self.parameter[...]`. If the value is updated by other methods or is a computed value it should definitely stay an instance variable, though.

lawhead (approved):

celikbasak (approved):
Looks good overall! I agree with Matt that we could clean up the duplications from RSVP and maybe inherit from a shared class.

bcipy/task/paradigm/matrix/calibration.py
tab-cmd: A ticket was created for this. We'll need to think about how we'd implement row/column vs scp. Many of these methods would change then, but it may make more sense for those two to be separate classes and inherit a common Calibration class as you say. https://www.pivotaltracker.com/story/show/181077145

bcipy/task/paradigm/matrix/calibration.py
tab-cmd: since there is no standard, I think this should go through as is. Let's surface this as a style discussion next we meet as a group. I can see drawbacks to both, but don't dislike how we've changed copy phrase.

tab-cmd (approved):

PR #193, opened 2022-01-21
Language model

sliu126:
# Overview

Added GPT2 language model using Huggingface implementation. Integrated with the existing language model base class. The language model makes character-level predictions given left context (evidence strings).

## Ticket

[https://www.pivotaltracker.com/story/show/178695479](url)

## Contributions

- Implemented a new GPT2LanguageModel class
- Addressed issues raised during meetings with Tab, Matt and Niklas
- Made changes based on discussions during meetings
- The module is able to make character-level predictions based on left context, using marginalized character-level probability distribution returned by the GPT2 language model

## Test

- test_gpt2.py in the test folder is an example test script that can be used to test the module.


bcipy/language/tests/test_gpt2.py
```diff
@@ -0,0 +1,22 @@
...
+    correct_char_rank = [c[0] for c in next_char_pred].index("S") + 1
+    print(correct_char_rank)
+    next_char_pred = lm.state_update(list("does_it_make_sens"))
+    print(next_char_pred)
+    correct_char_rank = [c[0] for c in next_char_pred].index("E") + 1
+    print(correct_char_rank)
+    next_char_pred = lm.state_update(list("does_it_make_sense"))
+    print(next_char_pred)
+    correct_char_rank = [c[0] for c in next_char_pred].index("_") + 1
+    print(correct_char_rank)
```
nik-sm: Can you convert this into a test that will be discovered by pytest? This will allow the test to be run automatically (locally and on github during pull requests)

You can see examples of test structure in the other tests folders `bcipy/*/tests/`.
Here's an example for the EEG signal model: https://github.com/CAMBI-tech/BciPy/blob/1.5.1/bcipy/signal/tests/model/test_pca_rda_kde.py

You could make a class that inherits from `unittest.TestCase` - here's also an example structure to copy/paste from:
```python
import unittest

class TestNewThing(unittest.TestCase):
    def setUpClass(self):
        # if you want some heavyweight work to happen only once
        self.my_costly_attr = ...

    def setUp(self):
        # if you want some lightweight work to happen fresh before each test
        self.my_quick_attr = ...
        
    def test_first_feature(self):
        expected = ...
        observed = ...
        self.assertEqual(expected, observed)  

    def test_error_on_bad_action(self):
         with self.assertRaises(ValueError):
             self.model.do_something_invalid()
```

nik-sm: If loading the model is too slow for a unit test (more than like ~1-10s), you can use the `@pytest.mark.slow` decorator on a function, or on the whole class, like this:
```python
@pytest.mark.slow
class MyTestCase(unittest.TestCase)
    @pytest.mark.slow
    def test_something_slow(self):
        ...
```

Then, people can run only the fast tests with `pytest -k "not slow"`, only the slow tests with `pytest -k "slow"`, etc

Slow tests are not (currently) run automatically during pull requests.
However, if the slow part is only downloading, we could still run this test during pull requests, and cache the downloaded file: https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows.
Setting that up would certainly take a bit of trial-and-error - happy to help with this.

nik-sm (changes requested):
Looks good overall! I left a comment about adding tests. The more tests the better (e.g. ~1 per behavior that the class has) - but even just some sanity check is still useful.

If everyone prefers, we could address this in a follow-up though.

bcipy/language/README.md
```diff
@@ -12,3 +12,5 @@ The core methods of any `LanguageModel` include:
 
 You may of course define other methods, however all integrated BciPy experiments using your model will require those to be defined!
 
+The pretrained GPT2 language model is saved in [this folder on Google Drive](https://drive.google.com/drive/folders/1pkvwHA8SR7awxf7fj7Ds4FhY6SGxeGtX?usp=sharing). Download the files in the folder and put them in a local directory. Then use the path to the local directory to load the model. (Alternatively, just pass in the model name like "gpt2" as the language model path and the pretrained language model will be downloaded and stoed in local cache)
+
```
lawhead, line 16: Since BciPy is publicly available and not just used by our group we may not want to include the save model location in the README. We would need to commit to continuing to provide it at that location.

bcipy/language/tests/test_gpt2.py
```diff
@@ -0,0 +1,22 @@
+from bcipy.language.model.gpt2 import GPT2LanguageModel
+from bcipy.helpers.task import alphabet
+from bcipy.language.main import ResponseType
+
+
+if __name__ == "__main__":
```
lawhead: I think this file should be referred to as a demo, rather than a test.

bcipy/language/tests/test_gpt2.py
lawhead: I agree that it would be great to have some unit tests. My preference would be if the underlying gpt2 model was mocked. We primarily want to test that your class is using it correctly. For instance, there are are a number of things you are managing in the `predict` method  (behavior changes depending on whether you are at the beginning of the word, etc). It would be good to demonstrate these conditions in a test. You may also want to have checks for your probability distributions.

bcipy/language/model/gpt2.py
```diff
@@ -0,0 +1,188 @@
...
+                if next_char in self.symbol_set and next_char != BACKSPACE_CHAR:
+                    char_to_prob[next_char] += self.curr_word_predicted_prob[idx]
+
+        # normalize char_to_prob
+        sum_char_prob = sum(char_to_prob.values())
+        for char in char_to_prob:
+            char_to_prob[char] /= sum_char_prob
+
+        # build a list of tuples (char, prob)
+        char_prob_tuples = [(k, v) for k, v in sorted(char_to_prob.items(), key=lambda item: item[1], reverse=True)]
```
lawhead: You don't need the list comprehension here:

    list(sorted(char_to_prob.items(), key=lambda item: item[1], reverse=True))

bcipy/language/model/gpt2.py
```diff
@@ -0,0 +1,188 @@
...
+            text = "<|endoftext|>"
+
+        # tokenize the text
+        indexed_tokens = self.tokenizer.encode(text)
+        tokens_tensor = torch.tensor([indexed_tokens])
+
+        # predict the next word
+        with torch.no_grad():
+            model = self.model
+            outputs = model(tokens_tensor)
```
lawhead: I don't think the `model` variable here is needed.

    self.model(tokens_tensor)

lawhead (approved):

bcipy/language/tests/test_gpt2.py
sliu126: I agree that unit tests would be helpful. I can look into it but I prefer it in a follow-up (submit another PR for unit tests). Hope it's OK.

bcipy/language/model/gpt2.py
sliu126: Yes, I will change this.

bcipy/language/model/gpt2.py
sliu126: Yes, I will change this.

bcipy/language/tests/test_gpt2.py
sliu126: I agree. I will move the file to a demo folder and change the file name.

nik-sm (approved):

bcipy/language/tests/test_gpt2.py
tab-cmd: Yes, please create a pivotal ticket for us to point during a refinement! https://www.pivotaltracker.com/n/projects/2460065

bcipy/language/README.md
tab-cmd: I'd be okay removing the link and having them download as it describes in the alternate. For our own internal use cases, we can probably use the drive one. Not sure which would download faster.

tab-cmd (approved):

PR #196, opened 2022-01-31
Unicode trigger labels

tab-cmd:
# Overview

Update `TriggerHandler` to support reading and writing of unicode trigger labels. This is required for operation of the `RSVP Time Test Calibration`.

## Ticket

[Pivotal Ticket Link](https://www.pivotaltracker.com/story/show/180972841)

## Contributions

- `helpers/triggers.py`: add encoding of `utf-8` to TriggerHandler
- `helpers/tests/test_triggers.py`: update to reflect encoding as a kwarg
- `helpers/README.md`: added a readme with high level information about the submodules. Added detailed trigger information.
- `task/task_registry.py`: encountered odd behavior testing this unicode fix with a mistyped task name. This change fails bci_main sooner with a more helpful message detailing the available registered tasks. 

## Test

- `make test-all`: all tests pass
- `offline_analysis.py`: ensured triggers loaded correctly

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Added a README detailing the trigger module.


bcipy/helpers/README.md
```diff
@@ -0,0 +1,140 @@
+# BciPy Helpers
```
nik-sm, line 1: Awesome documentation!🔥

bcipy/helpers/README.md
```diff
@@ -0,0 +1,140 @@
...
+# apply a 2 second offset to all timestamps loaded 
+triggers = TriggerHandler.load(path_to_trigger_save_location, offset=2.0)
+```
+
+### Trigger File Format
+
+Triggers as written from BciPy tasks are assumed to have the following structure,
+
+1. A single trigger is written per line with three columns (label type timestamp).
+2. If a clock offset is present, it is written with the type offset and will be added to any timestamp values written in the file. When using various clocks that don't
```
nik-sm: Is it possible to include multiple offsets? For example

```
starting_offset offset -5.0
another_offset offset 5.0
...
```

bcipy/helpers/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+
+### Loading
+
+To load a BciPy triggers.txt file, the TriggerHandler load method can be used. Because it is a staticmethod, the class does not need to be initialized before the load method is used. 
+
+```python
+from bcipy.helpers.triggers import TriggerHandler
+
+path_to_trigger_save_location = './triggers.txt'
```
nik-sm, line 93: Is there a reason why `.txt` suffix is included during load, but not during save?

nik-sm (approved):
Looks good!

bcipy/helpers/README.md
tab-cmd: No reason. It should be changed though. I have a ticket in the backlog to remove extensions from our other file types (session, raw data) and will add this to it! https://www.pivotaltracker.com/story/show/181115524

bcipy/helpers/README.md
tab-cmd: That's a good callout, we only use the first offset written. The other will be loaded as a trigger and can be applied using other functions to the other loaded Triggers if desired. https://github.com/CAMBI-tech/BciPy/blob/1.5.1/bcipy/helpers/triggers.py#L274

bcipy/helpers/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Triggers
+
+Triggers consist of a label, type and a timestamp. These are what is used to align events both in real-time and after a session for further processing or model training. During operation of a BciPy task, Triggers are generated when display or other events of interest occur. These are timestamped using a monotonic clock (See helpers/clock.py).
+
+A new Trigger may defined as follows:
+```python
+from bcipy.helpers.triggers import Trigger, TriggerType
+
+# label can be any utf-8 complaint string
```
lawhead: This should be 'compliant'.

bcipy/helpers/triggers.py
```diff
@@ -212,6 +212,8 @@ class TriggerHandler:
     writing triggers and loading triggers from a txt file.
     """
 
+    encoding = 'utf-8'
```
lawhead: We are getting warnings on all of our `open` statements throughout BciPy. Should we have a constant in `system_utils` for this?

lawhead (approved):

bcipy/helpers/triggers.py
tab-cmd: I'm running python 3.8.10. I get warnings for other things when I run the tests or suite, but none for open. Can you paste the warning here?

lawhead: I'm seeing these as pylint warnings:

   `Using open without explicitly specifying an encoding`

lawhead: Thanks for fixing these!

PR #197, opened 2022-02-05
Integrate GPT2 Language Model

lawhead:
# Overview

Integrated the new GPT2LanguageModel so BciPy users can select it and it will be used to support the Copy Phrase Task.

## Ticket

[https://www.pivotaltracker.com/story/show/181018867](https://www.pivotaltracker.com/story/show/181018867)

## Contributions

- Refactored parent LanguageModel to use stricter checking for supported response type and to include common functionality
- Refactored other existing language models to use the parent class
- Updated the language model helper code for selecting a language model and passing it the correct parameters.
- Cleanup

## Test

- Ran all unit tests
- Ran Copy Phrase task with each of the available language models (PRELM, GPT2, UNIFORM) and analyzed the log to confirm that the probability histogram changed appropriately with each model.

### DIFF
--- a/bcipy/helpers/copy_phrase_wrapper.py
+++ b/bcipy/helpers/copy_phrase_wrapper.py
@@ -1,8 +1,8 @@
 """Defines the CopyPhraseWrapper."""
-from typing import Any, List, Tuple
+from typing import List, Tuple
 
-import numpy as np
 import logging
+import numpy as np
 from bcipy.helpers.acquisition import analysis_channels
 from bcipy.helpers.exceptions import BciPyCoreException
 from bcipy.helpers.language_model import (
@@ -23,6 +23,7 @@ from bcipy.task.control.criteria import (
     ProbThresholdCriteria,
 )
 from bcipy.task.data import EvidenceType
+from bcipy.language.main import LanguageModel, ResponseType
 from bcipy.language.uniform import UniformLanguageModel
 
 
@@ -82,7 +83,7 @@ class CopyPhraseWrapper:
                  ],
                  task_list: List[Tuple[str, str]] = [('I_LOVE_COOKIES',
                                                       'I_LOVE_')],
-                 lmodel: Any = None,
+                 lmodel: LanguageModel = None,
                  is_txt_stim: bool = True,
                  device_name: str = 'LSL',
                  device_channels: List[str] = None,
@@ -96,13 +97,6 @@ class CopyPhraseWrapper:
                  notch_filter_frequency: int = 60,
                  stim_length: int = 10,
                  stim_order: StimuliOrder = StimuliOrder.RANDOM):
-
-        if not lmodel:
-            # There is enough information provided to construct a text-based
-            # UniformLanguageModel but additional parameters are needed for
-            # an image-based model.
-            assert is_txt_stim, "A language model is required when using non-text stimulus."
-            lmodel = UniformLanguageModel(lm_backspace_prob=backspace_prob)
 
         self.conjugator = EvidenceFusion(evidence_names, len_dist=len(alp))
 
@@ -146,9 +140,13 @@ class CopyPhraseWrapper:
 
         self.mode = 'copy_phrase'
         self.task_list = task_list
-        self.lmodel = lmodel
         self.channel_map = analysis_channels(device_channels, device_name)
         self.backspace_prob = backspace_prob
+
+        self.lmodel = lmodel if lmodel else UniformLanguageModel(
+            response_type=ResponseType.SYMBOL,
+            symbol_set=self.alp,
+            lm_backspace_prob=backspace_prob)
 
     def evaluate_inquiry(
             self, raw_data: np.array, triggers: List[Tuple[str, float]],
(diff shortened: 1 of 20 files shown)


bcipy/helpers/copy_phrase_wrapper.py
```diff
@@ -82,7 +83,7 @@ def __init__(self,
                  ],
                  task_list: List[Tuple[str, str]] = [('I_LOVE_COOKIES',
                                                       'I_LOVE_')],
-                 lmodel: Any = None,
+                 lmodel: LanguageModel = None,
```
tab-cmd: I'd vote to require language model as input.

bcipy/parameters/parameters.json
```diff
@@ -744,16 +744,25 @@
     "type": "bool"
```
tab-cmd: I think we could safely remove `lang_model_enabled` and require the type to be set with uniform as our default. It would be more accurate to what is happening.

bcipy/parameters/parameters.json
```diff
@@ -744,16 +744,25 @@
...
-    "helpTip": "Specifies which language model to use. Default: PRELM",
+    "helpTip": "Specifies which language model to use. Default: GPT2",
     "recommended_values": [
+      "GPT2",
       "PRELM",
       "UNIFORM"
     ],
     "type": "str"
   },
+  "lm_path": {
```
tab-cmd, line 750: If the name would work, we could get it from the type or it could be set on the model itself. The path could be a pretrained or null for download.  Does the path to the pretrained model include the file itself or only the directory?

bcipy/language/model/gpt2.py
```diff
@@ -1,26 +1,27 @@
...
+
+from bcipy.helpers.task import BACKSPACE_CHAR, SPACE_CHAR
+from bcipy.language.main import LanguageModel, ResponseType
 
 
 class GPT2LanguageModel(LanguageModel):
+    """Character language model based on GPT2."""
 
-    def __init__(self, response_type, symbol_set, lm_path):
+    def __init__(self, response_type, symbol_set, lm_path="gpt2"):
```
tab-cmd: It would be helpful to have a log message in the terminal when downloading for the first time.

tab-cmd (changes requested):
@sliu126 I couldn't seem to download from huggingface and then use this offline again. I found some files saved in .cache/huggingface and could use it while connected to the internet, however once I shut off wifi and tried specifying a path to that cache it fails and wants to connect to the huggingface API. I wasn't able (via search) able to find another config.json or the other files it was interested in loading while offline.. Is there a way to specify where the pretrained is saved? Is it possible to use it that way?

I had no trouble using the downloaded model you linked in the README or when connected to the internet.

bcipy/helpers/copy_phrase_wrapper.py
sliu126: Yes the language model should be passed in.

bcipy/language/model/gpt2.py
sliu126: Huggingface prints a message indicating downloading progress when downloading for the first time.

bcipy/parameters/parameters.json
sliu126: The path only includes the directory.

bcipy/language/model/gpt2.py
```diff
@@ -1,26 +1,27 @@
...
         Args:
             response_type - SYMBOL only
             symbol_set - list of symbol strings
             lm_path - path to language model files
         """
-        self.response_type = response_type
-        self.symbol_set = symbol_set
-        self.__validate_response_type()
+        super().__init__(response_type=response_type, symbol_set=symbol_set)
+        self.normalized = True
```
sliu126: I'm not sure why we need this normalized variable, since the output distribution should always be normalized.

tab-cmd:
We should describe the Uniform Language Model in the README. I'm okay linking to that pretrained google drive, but we should also put an email link to reach out should it go down or we move it... We can use cambi_support@googlegroups.com

bcipy/language/model/gpt2.py
tab-cmd: I saw the download bar and it's helpful. It doesn't say what it's downloading or why.. when running a task it's unclear what is happening

sliu126:
Yes, when offline we need to load from model files saved locally. It seems that we need internet connection when loading cached model.

bcipy/language/model/gpt2.py
lawhead: Yes, this is for downstream users. See CopyPhraseWrapper. The results of PRELM are in log likelihood and must be converted. This is just a flag. See also the LanguageModel base class for the documentation comment.

sliu126 (approved):

bcipy/parameters/parameters.json
lawhead: This parameter has type "str" because the model allows 'gpt2' as the lm_path, otherwise it would be type of "directorypath". Maybe we should change the model so the default value is `None`, and sets it to "gpt2" if missing.

bcipy/language/model/gpt2.py
lawhead: It probably makes sense to require language models to provide probabilities and just update PRELM accordingly.

tab-cmd:
> 

https://huggingface.co/docs/transformers/installation#offline-mode There is some support for offline use! I made a ticket in pivotal to support this in future releases

tab-cmd (approved):

PR #200, opened 2022-02-23
Fix inquiry reshaper

nik-sm:
# Overview

Fix `InquiryReshaper` to use the correct window of time for each inquiry.

Sidenote - this now returns data with shape `(inquiries, channels, samples)`.
The point of using this is to get data to feed into neural net for classifying entire inquiries, and that will expect to have batch dimension in front.
If it is preferred to be more consistent with `TrialReshaper` (which returns shape `(channels, trials, samples)`), we could adjust this PR (but future code will have to do
`data.transpose([1,0,2])`).

## Ticket

https://www.pivotaltracker.com/story/show/181349027

## Contributions

- Change how the time for an inquiry is calculated

## Test

- Updated the unit test for InquiryReshaper

## Documentation

- Docstring updated

### DIFF
--- a/bcipy/helpers/task.py
+++ b/bcipy/helpers/task.py
@@ -392,7 +392,7 @@ class InquiryReshaper(Reshaper):
             labels_excluded (Set[str]): labels to exclude. Defaults to empty set.
 
         Returns:
-            reshaped_data (np.ndarray): inquiry data of shape (Channels, Inquiries, Samples)
+            reshaped_data (np.ndarray): inquiry data of shape (Inquiries, Channels, Samples)
             labels (np.ndarray): integer label for each inquiry. With `trials_per_inquiry=K`,
                 a label of [0, K-1] indicates the position of `target_label`, or label of K indicates
                 `target_label` was not present.
--- a/bcipy/helpers/task.py
+++ b/bcipy/helpers/task.py
@@ -392,7 +392,7 @@ class InquiryReshaper(Reshaper):
             labels_excluded (Set[str]): labels to exclude. Defaults to empty set.
 
         Returns:
-            reshaped_data (np.ndarray): inquiry data of shape (Inquiries, Channels, Samples)
+            reshaped_data (np.ndarray): inquiry data of shape (Channels, Inquiries, Samples)
             labels (np.ndarray): integer label for each inquiry. With `trials_per_inquiry=K`,
                 a label of [0, K-1] indicates the position of `target_label`, or label of K indicates
                 `target_label` was not present.
@@ -433,7 +433,7 @@ class InquiryReshaper(Reshaper):
 
             reshaped_data.append(eeg_data[:, first_trigger: last_trigger + trial_duration_samples])
 
-        return np.stack(reshaped_data), labels
+        return np.stack(reshaped_data, 1), labels
 
 
 class TrialReshaper(Reshaper):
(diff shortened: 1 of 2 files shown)


tab-cmd:
> If it is preferred to be more consistent with TrialReshaper (which returns shape (channels, trials, samples)), we could adjust this PR (but future code will have to do
data.transpose([1,0,2])).

I'd be open to changing `TrialReshaper` to (trial, channel, samples) as well to keep them consistent.

bcipy/helpers/task.py
```diff
@@ -414,32 +411,29 @@ def __call__(self,
...
         # triggers in seconds are mapped to triggers in number of samples.
         triggers = list(map(lambda x: int((x + offset) * fs), timing_info))
 
-        # shape (Channels, Inquiries, Samples)
-        reshaped_data = np.zeros((len(eeg_data), n_inquiry, num_samples))
-
         # Label for every inquiry
-        labels = np.zeros(n_inquiry)
-        for inquiry_idx, chunk in enumerate(grouper(zip(trial_labels, triggers), trials_per_inquiry)):
+        labels = np.zeros(n_inquiry, dtype=np.long)
```
tab-cmd, line 420: Would you mind doing this with the `TrialReshaper` in this PR too?

tab-cmd (approved):

bcipy/helpers/task.py
```diff
@@ -414,32 +411,29 @@ def __call__(self,
...
+        reshaped_data = []
+        for inquiry_idx, trials_within_inquiry in enumerate(grouper(zip(trial_labels, triggers), trials_per_inquiry)):
             # label is the index of the "target", or else the length of the inquiry
             inquiry_label = trials_per_inquiry
-            first_trigger = None
-            for trial_idx, (trial_label, trigger) in enumerate(chunk):
-                if first_trigger is None:
-                    first_trigger = trigger
-
+            for trial_idx, (trial_label, trigger) in enumerate(trials_within_inquiry):
```
lawhead, line 425: I feel like Python is missing an abstraction that would help here. It has the `index` method for lists, but this method doesn't allow for a `key` parameter to choose how to compare items (like the `sorted` function). Something like:

    def index_of(lst, element, key=None):
        for i, item in enumerate(lst): 
            if (key and key(item) == element) or (not key and item == element): 
                return i 
        return None

Then the code here could be:
 
    inquiry_label = index_of(trials_within_inquiry, target_label, key=operator.itemgetter(0)) or trials_per_query

which almost matches the comment. Here's an alternate implementation:

    def index_of(lst, element, key=None):
        predicate = (lambda item: key(item) == element) if key else (lambda item: item == element)
        for i, item in enumerate(lst): 
            if predicate(item): 
                return i 
        return None

lawhead (approved):

bcipy/helpers/task.py
nik-sm: Interesting idea! I haven't written C code since compilers class, but this sounds like it would be a great contribution to python source btw...

bcipy/helpers/task.py
nik-sm: Sure

nik-sm:
To keep it simple for now, I made `InquiryReshaper` also return shape `(channels, inquiries, samples)`. 

I think it would be nice sometime in the future to use the convention of `(items, channels, samples)`, but I realized that current signal model has lots of hard-coded `data.shape[0]` and `data.shape[1]` - so we'll have to change it carefully.

P.S. - seems like something may be wrong with macOS CI builds?

- First one (https://github.com/CAMBI-tech/BciPy/actions/runs/1885303065) for macOS+python3.7
- After latest commit, macOS+python3.8 is failing (https://github.com/CAMBI-tech/BciPy/runs/5313374200?check_suite_focus=true) due to:
  ```
  FAILED bcipy/helpers/tests/test_task.py::TestGetKeyPress::test_get_key_press_appends_stamp_label_defaults
  ```

tab-cmd:
> To keep it simple for now, I made `InquiryReshaper` also return shape `(channels, inquiries, samples)`.
> 
> I think it would be nice sometime in the future to use the convention of `(items, channels, samples)`, but I realized that current signal model has lots of hard-coded `data.shape[0]` and `data.shape[1]` - so we'll have to change it carefully.
> 
> P.S. - seems like something may be wrong with macOS CI builds?
> 
> * First one (https://github.com/CAMBI-tech/BciPy/actions/runs/1885303065) for macOS+python3.7
> * After latest commit, macOS+python3.8 is failing (https://github.com/CAMBI-tech/BciPy/runs/5313374200?check_suite_focus=true) due to:
>   ```
>   FAILED bcipy/helpers/tests/test_task.py::TestGetKeyPress::test_get_key_press_appends_stamp_label_defaults
>   ```

All seem reasonable to me! The failure isn't related to your change, feel free to merge!

PR #202, opened 2022-02-28
GPT2 tests

lawhead:
# Overview

Added unit tests for GPT2 language model. Note that tests were marked as slow due to the long initialization of the language model, so they are not run with the unit-test suite but will be run with `test-all`.

## Ticket

https://www.pivotaltracker.com/story/show/181349857

## Contributions

- Added tests

## Test

- Ran the new test file with `pytest`. Ran all other unit tests to ensure that nothing broke.

### DIFF
--- a/bcipy/language/tests/test_gpt2.py
+++ b/bcipy/language/tests/test_gpt2.py
@@ -1,0 +1,76 @@
+"""Tests for GPT2 Language Model"""
+
+import pytest
+import unittest
+from operator import itemgetter
+
+from bcipy.helpers.exceptions import UnsupportedResponseType
+from bcipy.helpers.task import alphabet
+from bcipy.language.model.gpt2 import GPT2LanguageModel, ResponseType
+
+
+@pytest.mark.slow
+class TestGPT2LanguageModel(unittest.TestCase):
+    """Tests for language model"""
+    @classmethod
+    def setUpClass(cls):
+        cls.lmodel = GPT2LanguageModel(response_type=ResponseType.SYMBOL,
+                                       symbol_set=alphabet())
+
+    def test_init(self):
+        """Test default parameters"""
+        self.assertEqual(self.lmodel.response_type, ResponseType.SYMBOL)
+        self.assertEqual(self.lmodel.symbol_set, alphabet())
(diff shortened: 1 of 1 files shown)


bcipy/language/tests/test_gpt2.py
```diff
@@ -0,0 +1,76 @@
...
+            len(set(probs)) > 1,
+            "All values should not be the same probability")
+        for prob in probs:
+            self.assertTrue(0 <= prob < 1)
+        self.assertAlmostEqual(sum(probs), 1)
+
+        most_likely_sym, _prob = sorted(symbol_probs,
+                                        key=itemgetter(1),
+                                        reverse=True)[0]
+        self.assertEqual('E', most_likely_sym,
```
tab-cmd, line 67: @sliu126 are there other checks or phrases we could add here?

tab-cmd (approved):

bcipy/language/tests/test_gpt2.py
lawhead: There are probably some stronger assertions that we can / should make, but I definitely wanted to leave that to the LM team to add as they see fit. For instance, we could say that after a space is typed `THE_` the probability of another space character is some small number (still being worked on).

bcipy/language/tests/test_gpt2.py
```diff
@@ -0,0 +1,76 @@
...
+import pytest
+import unittest
+from operator import itemgetter
+
+from bcipy.helpers.exceptions import UnsupportedResponseType
+from bcipy.helpers.task import alphabet
+from bcipy.language.model.gpt2 import GPT2LanguageModel, ResponseType
+
+
+@pytest.mark.slow
```
lawhead, line 12: One thing to note is that I marked this test suite as slow due to the load times of GPT2. If we can figure out a way to speed this up, without including the entire thing in our repository, we could remove this annotation.

tab-cmd: I'm okay with this for now! At some point, it would be nice to abstract these libraries from our unit tests and mock them.

sliu126:
Sorry for being late on this. I approve it.

PR #203, opened 2022-03-02
Fixed backspace prob

lawhead:
# Overview

Modified Copy Phrase Wrapper to always use the configured backspace probability, overriding the value provided by the language model.

## Ticket

[https://www.pivotaltracker.com/story/show/181350204](https://www.pivotaltracker.com/story/show/181350204)

## Contributions

- Updated the behavior of the `sym_appended` to update the provided probability when the list of probs already contains that value.
- Updated the `UniformLanguageModel` to remove redundant code for configuring the backspace probability.
- Added unit tests to cover edge cases.

## Test

- Ran all unit tests.
- Ran the Copy Phrase task with both GPT2 and Uniform language models to ensure that the configured backspace probability was used and the other probabilities were adjusted.

### DIFF
--- a/bcipy/helpers/copy_phrase_wrapper.py
+++ b/bcipy/helpers/copy_phrase_wrapper.py
@@ -7,7 +7,7 @@ from bcipy.helpers.acquisition import analysis_channels
 from bcipy.helpers.exceptions import BciPyCoreException
 from bcipy.helpers.language_model import (
     histogram,
-    sym_appended,
+    with_min_prob,
 )
 from bcipy.helpers.stimuli import InquirySchedule, StimuliOrder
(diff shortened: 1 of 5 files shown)


tab-cmd (approved):

tab-cmd:
I'm not sure this PR captures the ticket. If I understand the behavior now it would always be the value set in parameters? We had wanted to use that as a minimum to allow LM to upvote, but prevent it from being downvoted too low.

lawhead:
> I'm not sure this PR captures the ticket. If I understand the behavior now it would always be the value set in parameters? We had wanted to use that as a minimum to allow LM to upvote, but prevent it from being downvoted too low.

Okay, I misunderstood the requirements. I'll add some updates with that behavior.

PR #204, opened 2022-03-02
Bug fix for Preview button evidence

lawhead:
# Overview

Fixed a bug that was providing button evidence when user selected the parameter to show the Preview Only in the Copy Phrase Task.

## Ticket

https://www.pivotaltracker.com/story/show/181427286

## Contributions

- Fixed the value attribute of the parameter so it was cast to the correct boolean type.

## Test

- Ran the Copy Phrase task to confirm. Inspected the parameter using the debugging tools.

### DIFF
--- a/bcipy/task/paradigm/rsvp/copy_phrase.py
+++ b/bcipy/task/paradigm/rsvp/copy_phrase.py
@@ -134,7 +134,7 @@ class RSVPCopyPhraseTask(Task):
         self.parameters.add_entry(
             'preview_only',
             {
-                'value': True if self.parameters['preview_inquiry_progress_method'] == 0 else False,
+                'value': 'true' if self.parameters['preview_inquiry_progress_method'] == 0 else 'false',
                 'section': '',
                 'readableName': '',
                 'helpTip': '',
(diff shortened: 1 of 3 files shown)


bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -134,7 +134,7 @@ def __init__(self, win, daq, parameters, file_save, signal_model,
         self.parameters.add_entry(
             'preview_only',
             {
-                'value': True if self.parameters['preview_inquiry_progress_method'] == 0 else False,
+                'value': 'true' if self.parameters['preview_inquiry_progress_method'] == 0 else 'false',
```
lawhead, line 137: The Parameters module expects boolean values to be a string of 'true' or 'false'. This was the simplest fix for now. We may also want the Parameters module to be more lenient and allow Python boolean values, but that will require more testing to ensure it doesn't break any other existing code.

tab-cmd: We could add a check in `Parameters.check_valid_entry` that asserts values are strings or booleans are "true/false" in the meantime? I'd be okay with either approach

tab-cmd (approved):

PR #205, opened 2022-03-04
Matrix calibration

jcgangemi1:
# Overview

Updates to matrix calibration stimuli generation as requested after demo.  Added parameter to toggle between randomly and evenly distributed target positions within inquiry.  Additional parameter to specify percentage of inquiries with a target that is not flashed in the inquiry (nontarget inquiries).  Increased contrast in flashed stimuli.  

## Ticket

https://www.pivotaltracker.com/story/show/181038822

## Contributions

- task/paradigm/matrix/calibration.py: added two new parameters, enum called target_positions which can be either random or distributed and decides whether the target positions are random or evenly distributed, and integer called nontarget_inquiries that specifies a percentage 0-100 of inquiries that will not contain the target stimuli
- helpers/stimuli.py: added new function to choose evenly distributed target positions based on inquiry attributes, updated stimuli generation function to include nontarget inquiries and collect evenly distributed data for different target positions
- helpers/tests/test_stimuli.py: added new tests to thoroughly check different options for random or distributed target positions, and different values 0-100 percent for amount of nontarget inquiries
- parameters/parameters.json: added two new parameters, target_positions and nontarget_inquiries
- display/paradigm/matrix/display.py: changed values of stimuli opacity to flash with a more noticeable contrast from other stimuli

## Test

- helpers/test/test_stimuli.py: added unit tests to test stimuli generation with different values for two new parameters
- bcipy --task "Matrix Calibration": stimuli are chosen and flashed correctly with updates

## Documentation

- Are documentation updates required? In-line, README, or [documentation] Added in-line comments to explain updates


bcipy/display/paradigm/matrix/display.py
```diff
@@ -7,7 +7,7 @@
 from bcipy.display import Display, StimuliProperties, TaskDisplayProperties, InformationProperties, BCIPY_LOGO_PATH
 from bcipy.helpers.task import SPACE_CHAR
 from bcipy.helpers.stimuli import resize_image
-from bcipy.helpers.triggers import TriggerCallback
+from bcipy.helpers.triggers import TriggerCallback, _calibration_trigger
```
lawhead: If we're importing this into other modules, it seems like we should remove the leading underscore and treat this as a public function.

bcipy/parameters/parameters.json
```diff
@@ -361,6 +361,25 @@
...
+      "distributed",
+      "random"
+    ],
+    "type": "str"
+  },
+  "nontarget_inquiries": {
+    "value": "10",
+    "section": "bci_config",
+    "readableName": "Percentage of Nontarget Inquiries",
+    "helpTip": "Specifies the percentage (0-100) of iquiries which target stimuli flashed is not in inquiry. Default is 10 percent.",
```
lawhead: There's a typo here in `inquiries`.

bcipy/helpers/stimuli.py
```diff
@@ -243,26 +262,37 @@ def calibration_inquiry_generator(
     """
 
     len_alp = len(alp)
+    targets = []
```
lawhead: I found the logic of this function difficult to follow due to the amount of index juggling and list slicing. I also found it confusing that `stim_length` is used as the target index when there should not be a value. I think a lot of this could be cleaned up by explicitly declaring a sentinel value for the case where an inquiry should not display the target. Also, the `random` module has a lot of functions which can be helpful. Here is a proposed implementation.

```python
def calibration_inquiry_generator(
        alp: List[str],
        timing: List[float] = [0.5, 1, 0.2],
        color: List[str] = ['green', 'red', 'white'],
        stim_number: int = 10,
        stim_length: int = 10,
        stim_order: StimuliOrder = StimuliOrder.RANDOM,
        target_positions: TargetPositions = TargetPositions.RANDOM,
        nontarget_inquiries: int = 10,
        is_txt: bool = True) -> InquirySchedule:
    target_indexes = []
    no_target = None # sentinel value for inquiries that don't display a target
    if target_positions == target_positions.DISTRIBUTED:
        target_indexes = distributed_target_positions(stim_number, stim_length, nontarget_inquiries)
        # this line can be removed if `distributed_target_positions` uses None.
        target_indexes = [no_target if val == 10 else val for val in target_indexes]
    else:
        num_nontarget_inquiry = int((nontarget_inquiries / 100) * stim_number)
        num_target_inquiry = stim_number - num_nontarget_inquiry
        target_indexes = [no_target] * num_nontarget_inquiry
        target_indexes.extend(random.choices(range(stim_length), k=num_target_inquiry))
        random.shuffle(target_indexes)

    samples, times, colors = [], [], []

    for i in range(stim_number):
        inquiry = random.sample(alp, k=stim_length)
        if stim_order == StimuliOrder.ALPHABETICAL:
            inquiry = alphabetize(inquiry)

        target_index = target_indexes[i]
        if target_index is no_target:
            target = random.choice(list(set(alp) - set(inquiry)))
        else:
            target = inquiry[target_index]
        sample = [target, get_fixation(is_txt=is_txt), *inquiry]

        samples.append(sample)
        times.append([timing[i] for i in range(len(timing) - 1)] +
                     [timing[-1]] * stim_length)
        colors.append([color[i] for i in range(len(color) - 1)] +
                      [color[-1]] * stim_length)

    return InquirySchedule(samples, times, colors)
```

bcipy/helpers/stimuli.py
```diff
@@ -273,6 +303,48 @@ def calibration_inquiry_generator(
     return InquirySchedule(samples, times, colors)
 
 
+def distributed_target_positions(stim_number: int, stim_length: int, nontarget_inquiries: int) -> list:
```
lawhead, line 301: I like how the calculation of target indexes is split off into a separate function. I think the implementation can probably be simplified using some of the techniques above. You should also be explicit in the documentation about what value is used when the target should not be shown (my preference is `None` as described above).

bcipy/helpers/tests/test_stimuli.py
```diff
@@ -221,6 +229,263 @@ def test_alphabetical_inquiry_gen(self):
...
+        targets = distributed_target_positions(stim_number=stim_number,
+                                               stim_length=stim_length,
+                                               nontarget_inquiries=nontarget_inquiries)
+
+        self.assertEqual(
+            len(targets), stim_number,
+            'Should have produced the correct number of targets for inquiries.')
+
+        # count how many times each target position is used
+        target_counts = (np.zeros(stim_length + 1)).astype(int)
```
lawhead: You might be able to simplify some of this by using the `Counter` from the `collections` module here: https://docs.python.org/3/library/collections.html#collections.Counter

bcipy/helpers/tests/test_stimuli.py
```diff
@@ -221,6 +229,263 @@ def test_alphabetical_inquiry_gen(self):
...
+        for pos in targets:
+            target_counts[pos] = (target_counts[pos] + 1)
+
+        # make sure position counts are equally distributed, including non-target
+        for i in target_counts:
+            self.assertTrue(i >= num_target_inquiries)
+            self.assertTrue(i <= (num_target_inquiries + 1))
```
lawhead: You can use chained comparisons on all of these.

https://docs.python.org/3/reference/expressions.html#comparisons

lawhead (approved):

bcipy/display/paradigm/matrix/display.py
```diff
@@ -167,14 +175,58 @@ def increment_position(self, pos: Tuple[float]) -> Tuple[float]:
...
+                                        pos=(0, 0),
+                                        wrapWidth=2,
+                                        colorSpace='rgb',
+                                        opacity=1,
+                                        depth=-6.0)
+        target_prompt.draw()
+        self.draw_static()
+        self.window.flip()
+
+        core.wait(2)  # get first_pres_time and set on matrix display! What does RSVP use?
```
tab-cmd: We should use the `time_prompt` parameter for this.

tab-cmd (changes requested):
I think other than pulling the `time_prompt` parameter or pulling it from the task, this is ready to go! Once the lab is setup at NEU after Thursday, let's collect some calibration data and see how it goes!

tab-cmd (approved):

PR #206, opened 2022-03-09
1.5.1

sliu126:
# Overview

A new beam-search based method for GPT2 character prediction. The language model can predict multiple worldpieces and marginalize the word-level prediction to make character level prediction.

## Ticket

[https://www.pivotaltracker.com/story/show/181349709](url)

## Contributions

- Fixed the multiple spaces bug
- Improved the overall (ranked-based) performance on the ALS phrase dataset
- All letters in the alphabet (except for the backspace character) now have nonzero probability after interpolating with a unigram language model
- Smoothed the language model distribution with an exponential rescaling factor

## Test

- Run all unit tests existed.


sliu126:
Should merge from gpt2_beam_search into 1.5.1, not the other way around.

PR #208, opened 2022-03-28
Prestimulus, Inquiry Based Training, Model Tuning

tab-cmd:
# Overview

Added prestimulus reshaping and real time data querying support. Refactored offline analysis to support Inquiry based reshaping to more closely match what will be fed in real time. PcaRdaKde model updates to decrease training time and limit the magnitude of likelihood responses. 

## Ticket

https://www.pivotaltracker.com/story/show/181504827
https://www.pivotaltracker.com/story/show/181181932
https://www.pivotaltracker.com/story/show/178055444
https://www.pivotaltracker.com/story/show/178045750

## Contributions

- `task_buffer_len` to `task_buffer_length` to match other parameters and used to create a poststimulus buffer. Remove `eeg_buffer_len` as redundant and inconsistently used.
- `bcipy/helpers/list.py`: add grouper method
- `helpers/stimuli.py` moved reshapers here and updated to include pre/poststimulus handling. Remove the inclusions / exclusions arguments (this filtering happens in several methods already). 
-  `bcipy/helpers/task.py`: update `get_data_for_decision` to include prestimulus length 
- `pytest.ini`: added to suppress terminal warnings

Model changes:
- To compute AUC, PcaRdaKdeModel performs a k-fold cross validation. With `k_folds=10`, the data is divided into 10 disjoint pieces. In each fold, the model is trained on 90% of data and evaluated on 10%. This cannot be the final step of the fitting procedure, or the model will be left trained on only 90% of data (we want the model to be trained on 100% of the train data). Consequences: 
  - We change how AUC is recorded
  - `model.evaluate()` should not be used in its current state (instead, it should make a copy of the model and use this in a k-fold evaluation).

- Clip the likelihood updates between `[1e-2, 1e2]`, so that the model cannot make extreme changes in a single inquiry.
- Change PCA to preserve a specified fraction of total variance, by default 0.90 (previously it was doing no dimension reduction!).
- Remove unused parameters from PCA and KDE and cleanup documentation
- Fix KDE bandwidth estimation to properly use Silverman's rule of thumb
- Add a step in offline_analysis to estimate balanced accuracy based on an 80/20 split of the training data. This should hopefully be a better metric/more correlated with actual typing performance than AUC, though we need to explore to demonstrate this.
- Change bandpass filtering range from `[2, 45]` to `[1, 20]`

## Test

- `make test-all`, `calibration + copy phrase`, `simulations`
- `pytest --mpl`

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Updated docstrings... more may be needed.


NOTES.md
```diff
@@ -0,0 +1,34 @@
+Trying different model changes
```
tab-cmd: TODO: remove notes

bcipy/helpers/tests/resources/mock_session/parameters.json
```diff
@@ -46,14 +46,6 @@
     "recommended_values": "",
     "type": "int"
   },
-  "eeg_buffer_len": {
-    "value": "2.00",
```
tab-cmd, line 50: TODO: add prestim_length or pull from root

tab-cmd (commented):
We should regenerate test assests with the larger prestimulus length defined. It will look much cleaner and should score the same!

bcipy/helpers/list.py
```diff
@@ -18,3 +19,19 @@ def destutter(items: List, key: Callable = lambda x: x) -> List:
...
+    # grouper('ABCDEFG', 3, fillvalue='x') --> ABC DEF Gxx
+    # grouper('ABCDEFG', 3, incomplete='strict') --> ABC DEF ValueError
+    # grouper('ABCDEFG', 3, incomplete='ignore') --> ABC DEF
+    chunks = [iter(iterable)] * chunk_size
+    if incomplete == "fill":
+        return zip_longest(*chunks, fillvalue=fillvalue)
+    if incomplete == "strict":
+        return zip(*chunks, strict=True)
+    if incomplete == "ignore":
+        return zip(*chunks)
```
lawhead, line 34: It looks like we're only ever using the default settings in our code. Does it make sense to leave these other modes here? Also, there are no unit tests for this function in `test_list.py`.

bcipy/helpers/stimuli.py
```diff
@@ -52,6 +57,169 @@ class InquirySchedule(NamedTuple):
...
+
+class InquiryReshaper:
+    def __call__(self,
+                 trial_targetness_label: List[str],
+                 timing_info: List[float],
+                 eeg_data: np.ndarray,
+                 fs: int,
+                 trials_per_inquiry: int,
+                 offset: float = 0,
+                 channel_map: List[int] = DEFAULT_CHANNEL_MAP,
```
lawhead: It doesn't make sense to me to have a DEFAULT_CHANNEL_MAP. Even within our own lab we are using more than one type of headset and they don't have the same number of channels. This is highly device specific (and montage-specific even with the same device).

bcipy/helpers/stimuli.py
```diff
@@ -52,6 +57,169 @@ class InquirySchedule(NamedTuple):
...
+
+
+class TrialReshaper(Reshaper):
+    def __call__(self,
+                 trial_targetness_label: list,
+                 timing_info: list,
+                 eeg_data: np.ndarray,
+                 fs: int,
+                 offset: float = 0,
+                 channel_map: List[int] = DEFAULT_CHANNEL_MAP,
```
lawhead: The same issue with the default channel map applies here as well.

bcipy/parameters/parameters.json
```diff
@@ -89,6 +81,14 @@
     "recommended_values": "",
     "type": "float"
   },
+  "prestim_length": {
+    "value": "1",
+    "section": "bci_config",
+    "readableName": "Prestimulus Window Length",
+    "helpTip": "Specifies the length (in seconds) of the EEG data window to return before stimulus presentation. Default: 0.5",
```
lawhead: The help says that the default is 0.5 but the value here is 1.

bcipy/parameters/parameters.json
```diff
@@ -98,7 +98,7 @@
     "type": "filepath"
   },
   "filter_high": {
-    "value": "45",
+    "value": "20",
```
lawhead, line 101: We should update the help text to reflect the new default value.

bcipy/parameters/parameters.json
```diff
@@ -313,11 +313,11 @@
...
     "type": "int"
   },
-  "task_buffer_len": {
-    "value": "0.75",
+  "task_buffer_length": {
+    "value": "2",
     "section": "bci_config",
     "readableName": "Inter-inquiry Interval",
-    "helpTip": "Specifies the delay time (in seconds) between the final stimulus in one inquiry and the beginning (target stimulus or fixation cross) of the next inquiry in an RSVP task. Default: 0.75",
+    "helpTip": "Specifies the delay time (in seconds) between the final stimulus in one inquiry and the beginning (target stimulus or fixation cross) of the next inquiry in a task. Default: 1",
```
lawhead: This help text also does not agree with the value.

bcipy/signal/model/offline_analysis.py
```diff
@@ -10,121 +10,165 @@
...
+from sklearn.metrics import balanced_accuracy_score
 
 log = logging.getLogger(__name__)
-logging.basicConfig(
-    level=logging.INFO,
-    format='[%(threadName)-9s][%(asctime)s][%(name)s][%(levelname)s]: %(message)s')
+logging.basicConfig(level=logging.INFO, format="[%(threadName)-9s][%(asctime)s][%(name)s][%(levelname)s]: %(message)s")
+
+
+def subset_data(data, labels, test_size, random_state=0):
```
lawhead: This function could use type info and documentation.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -10,15 +10,18 @@
...
-from bcipy.helpers.task import TrialReshaper
+from bcipy.helpers.stimuli import InquiryReshaper
 
 
 class PcaRdaKdeModel(SignalModel):
-    reshaper = TrialReshaper()
+    reshaper = InquiryReshaper()
 
-    def __init__(self, k_folds: int):
+    def __init__(self, k_folds: int, prior_type="uniform", pca_n_components=0.9):
```
lawhead, line 19: Would it make sense to use an enumeration for `prior_type`?

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -503,11 +505,15 @@ def compute_eeg_evidence(self,
...
             inquiry_timing=self.stims_for_decision(stim_times),
             daq=self.daq,
-            static_offset=self.parameters['static_trigger_offset'],
-            buffer_length=self.parameters['trial_length'])
+            offset=self.parameters['static_trigger_offset'],
+            prestim=self.parameters['prestim_length'],
+            poststim=self.parameters['task_buffer_length'] + self.parameters['trial_length'])
+
+        # we assume all are nontargets at this point
+        labels = ['nontarget'] * len(triggers)
```
lawhead, line 516: Why assume that these are all non-targets? Is this a TODO item? The `task` helper module has a `target_info` function for computing this given the triggers and target symbol.

bcipy/task/tests/core/test_handler.py
```diff
@@ -5,7 +5,7 @@
 from bcipy.task.control.criteria import CriteriaEvaluator, \
     MaxIterationsCriteria, MinIterationsCriteria, ProbThresholdCriteria
 from bcipy.task.control.query import NBestStimuliAgent
-from bcipy.helpers.task import alphabet
+from bcipy.language.main import alphabet
```
lawhead: Why re-import this from language.main rather than from where it's defined in helpers.task?

lawhead (approved):

bcipy/helpers/list.py
tab-cmd: We used ignore (replay_session.py) and fill (default), I'll add some tests for these and remove strict

bcipy/helpers/stimuli.py
tab-cmd: I agree and removed it! The default is now none, with all channels included in reshaping

lawhead: Thanks!

bcipy/helpers/tests/resources/mock_session/parameters.json
nik-sm: We had discussed removing this `parameters.json` and just using `bcipy/parameters/parameters.json` always (so that the tests don't become stale/so we don't have to remember to update parameters in 2 places) - does this still sound good to you?

bcipy/parameters/parameters.json
nik-sm: Should we change "recommended_values" section to only include `20`? Not sure how the "recommended_values" list is actually used

tab-cmd: It's used in the GUI to help guide selection using a dropdown menu

bcipy/helpers/tests/resources/mock_session/parameters.json
tab-cmd: Yes - I can do that work for this PR!

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: I think for now it's probably equally simple to just use string with the `raise ValueError` clause, since this is only used in one place

bcipy/task/paradigm/rsvp/copy_phrase.py
sci-tab: We don't need these labels. Not sure why they were added before... I created a ticket to address this since it could get messy! https://www.pivotaltracker.com/story/show/181861101

nik-sm: Just chatted with Tab. Since we're just getting model predictions, presumably it should not matter what the labels are. Tab made a ticket to remove labels completely from this function: https://www.pivotaltracker.com/story/show/181861101

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -42,21 +49,31 @@ def fit(self, train_data: np.array, train_labels: np.array) -> SignalModel:
         rda_index = 1  # the index in the pipeline
         model.pipeline[rda_index].lam = arg_cv[0]
         model.pipeline[rda_index].gam = arg_cv[1]
-        _, sc_cv, y_cv = cost_cross_validation_auc(
+        tmp, sc_cv, y_cv = cost_cross_validation_auc(
```
celikbasak, line 52: Why this function returns negative AUC? (`-tmp`)

bcipy/signal/tests/model/test_pca_rda_kde.py
```diff
@@ -72,13 +72,11 @@ def setUp(self):
         np.random.seed(0)
 
     def test_pca(self):
-        var_tol = 0.95
-
         # .fit() then .transform() should match .fit_transform()
-        pca = ChannelWisePrincipalComponentAnalysis(num_ch=self.num_channel)
-        pca.fit(self.x, var_tol=var_tol)
+        pca = ChannelWisePrincipalComponentAnalysis(n_components=0.9, num_ch=self.num_channel)
```
celikbasak, line 76: Is `n_components=0.9` value recommended? If so, can we add a short comment about it?

replay_session.py
```diff
@@ -0,0 +1,411 @@
+"""Script that will replay sessions and allow us to simulate new model predictions on that data.
```
celikbasak: Sounds great!

bcipy/signal/model/pca_rda_kde/density_estimation.py
```diff
@@ -54,38 +31,34 @@ def _compute_bandwidth(self, scores: np.array, num_channels: int):
         Returns:
             float: rule-of-thumb bandwidth parameter for KDE
         """
-        return 1.06 * min(np.std(scores), iqr(scores) / 1.34) * np.power(num_channels, -0.2)
+        bandwidth = 0.9 * min(np.std(scores), iqr(scores) / 1.34) * np.power(num_items, -0.2)
```
celikbasak, line 35: Can we add the link to a source/paper for this bandwidth calculation?

celikbasak (approved):
Looks great! I added some minor thoughts/comments

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
nik-sm: The hyperparam search in `nonlinear_opt` function is using a generic optimization method that minimizes a function (`scipy.optimize.fmin_cobyla(...)`). In order to maximize `AUC`, it just minimizes `-AUC`.

That optimization function `nonlinear_opt` makes use of `cost_cross_validation_auc`, which is why `cost_cross_validation_auc` needs to return negative AUC.

bcipy/signal/tests/model/test_pca_rda_kde.py
nik-sm: We set this value in a relatively arbitrary way, and further tuning might be good.
I just used the same value in tests as is used in the main code, but this need not be matching. (If we change the value used in tests, we would just need to update the expected test output accordingly).

I'm not sure what info to add in a comment - do you have particular thoughts/questions about it?

bcipy/helpers/tests/resources/mock_session/parameters.json
nik-sm: We talked about this one and it is taken care of

bcipy/signal/tests/model/test_pca_rda_kde.py
celikbasak: This value (i.e. variance of the data) is "typically" chosen in the literature to be between 0.95-0.99. I am just trying to find a justification as to why should we set it as 0.9, or 0.95, or 0.97 etc. 

I realized you use the same value in the main code as well. I just thought this choice of parameter could use a source/justifying paper.

Another reason is that I've been specifically asked the number of principal components by a manuscript reviewer before. If we use a specific value like this in our papers the question's very likely to come up I think.

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
celikbasak: Thank you so much for the clarification!

bcipy/signal/model/offline_analysis.py
```diff
@@ -1,130 +1,193 @@
...
     model.fit(data, labels)
-    model_performance = model.evaluate(data, labels)
+    log.info(f"Training complete [AUC={model.auc:0.4f}]. Saving data...")
 
-    log.info(f'Training complete [AUC={model_performance.auc:0.4f}]. Saving data...')
+    model.save(data_folder + f"/model_{model.auc:0.4f}.pkl")
 
-    model.save(data_folder + f'/model_{model_performance.auc:0.4f}.pkl')
+    # Experimenting with an 80/20 split and checking balanced accuracy
+    # TODO - adjust estimator to match sklearn API and use cross_validate
```
tab-cmd: I'll pull this out into a ticket so we can remove the TODO!

tab-cmd: https://www.pivotaltracker.com/story/show/181874323

bcipy/signal/tests/model/test_pca_rda_kde.py
nik-sm: Do you have references for the values you cite as typical? 

Our two opposing objectives when setting this value are: 
1. Pick a value low enough to result in compression. In our case, we don't care about storing the transformed data, but rather we care about the resulting speedup of model fitting/predicting for RDA and KDE steps
2. Pick a value high enough to preserve structure in the data.

Given these two conflicting goals, however, the choice of tradeoff between them becomes a bit arbitrary. We picked the value 0.9 because we assumed that this will likely be on the "safe" end of the spectrum, in terms of basically preserving the structure of the data, but still manages to give some decent compression.

In past work, I have seen folks try to automatically determine the number of components by considering whether the residual (`original_signal - rank_K_reconstructed_signal`) "looks indistinguishable from noise" (according to a problem-specific noise model and a carefully chosen statistical test). Notice that this does not solve the problem, but rather trades one arbitrary choice ("what % variance is sufficient for my task?") for another arbitrary choice ("what statistical test, noise model, and p-value are sufficient for my task?").

I think this, or similar adaptive approaches, are probably overkill for the moment.

Would it be useful for me to add this discussion as documentation, just to explain the tradeoff? At the end of the day, someone will still need to make an arbitrary choice. (Even if this makes a reviewer uncomfortable!)

PR #210, opened 2022-04-18
#181490020 ; Remove old lm params

lawhead:
# Overview

Removed legacy parameters related to running the language model in a containerized environment.

## Ticket

https://www.pivotaltracker.com/story/show/181490020

## Contributions

- Removed extra parameters in both the main parameters file as well as the test resource.

## Test

- Ran all unit tests to confirm that they still worked.
- Grepped source code to ensure that the parameters were no longer in use in any demo scripts or documentation.


tab-cmd (approved):
Thanks! :)

PR #212, opened 2022-04-19
Prestim acq buffer

tab-cmd:
# Overview

Update init eeg acq buffer to include prestimulus information. Update copy phrase task to use correct poststim buffer. Update get_data_for_decision to return the correct triggers accounting for prestim.

++ Revert unintended change in copy phrase display

## Ticket

[pivotal ticket](https://www.pivotaltracker.com/story/show/181924087)

## Contributions

- [bcipy/display/paradigm/rsvp/mode/copy_phrase.py](https://github.com/CAMBI-tech/BciPy/compare/1.5.1...prestim_acq_buffer?expand=1#diff-a217bd47fc9c85f481890dcb1dc1ec0de949446d4d060c75be63c79c5a3d5719) -  revert unintended change in copy phrase display which would prevent stimuli from aligning in center.
- [bcipy/task/paradigm/rsvp/copy_phrase.py](https://github.com/CAMBI-tech/BciPy/compare/1.5.1...prestim_acq_buffer?expand=1#diff-a7325e7a6568ce0dd53f44c2df7a301b58a4e242e2e3065bee0968dcddf08cb9) - updates to use the correct post stim buffer. Add validate to ensure user-configured parameters are correct for the task. 
- [bcipy/helpers/acquisition.py](https://github.com/CAMBI-tech/BciPy/compare/1.5.1...prestim_acq_buffer?expand=1#diff-6624f72c18f53856c270a1341e9e186bcf1eb07b41899725b35e8775fd40767e) - add prestimulus duration
- [bcipy/helpers/exceptions.py](https://github.com/CAMBI-tech/BciPy/compare/1.5.1...prestim_acq_buffer?expand=1#diff-9eff57c94ac7934af52ba2a054e64cc51db298adea490af7e0f00236117ecc4a) - add task related exception for use in copy phrase

## Test

- `make test-all`
- full calibration + copy phrase to ensure spelling could occur

## Documentation


bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -100,8 +101,8 @@ def __init__(self, win, daq, parameters, file_save, signal_model,
         self.window = win
         self.daq = daq
         self.parameters = parameters
-        for param in RSVPCopyPhraseTask.PARAMETERS_USED:
-            assert param in self.parameters, f"parameter '{param}' is required"
+
+        self.validate()
```
lawhead: I like this pattern! We should consider adding it to the abstract parent class as a reminder when we are implementing other tasks that there may be some assumptions about the values and their relationships.

We may also want to encode constraints directly in the parameters.json file (ex. min/max values, etc) and have these be validated by the Parameters module, but there should still be some validations that are task-specific.

What do you think about making this more specific to parameters: `self.validate_parameters()` or `self.validate_parameters(parameters)`?

lawhead (approved):
I tested locally and this fixes the InsufficientDataException! Looking at the logs the buffer size looks good. There is about 1 second of data in the buffer before the query start time and about 0.6 seconds after. We could potentially decrease this size a bit but this is a fixed amount of memory so I don't think it would noticeably affect performance.

bcipy/task/paradigm/rsvp/copy_phrase.py
tab-cmd: I like that suggestion and added it to the PR. There are a few improvements we should address to parameter setting in general! I've made a ticket for now that we'll need to breakdown during our next refinement. https://www.pivotaltracker.com/story/show/181940742

PR #219, opened 2022-05-09
#181963356 ; fake data alerting

lawhead:
# Overview

Added better alerting to inform users when fake data mode is on.

## Ticket

https://www.pivotaltracker.com/story/show/181963356

## Contributions

- Added GUI utility for confirmation alerts.
- Set default parameter for fake_data to False.
- Added a confirmation alert in bci_main when fake data is present. The user can confirm to proceed with the task or cancel to stop it.

## Test

- Ran all unit tests.
- Ran a calibration task with the fake_data parameter set to both True and False. Setting to False should not display any alerts. Setting to True should display the alert. Canceling the alert should stop the task. Confirming the alert should proceed.

### DIFF
--- a/bcipy/gui/alert.py
+++ b/bcipy/gui/alert.py
@@ -1,7 +1,8 @@
 """GUI alert messages"""
 # pylint: disable=no-name-in-module
 import sys
-from PyQt5.QtWidgets import QApplication, QMessageBox
+from PyQt5.QtWidgets import QApplication
+from bcipy.gui.main import alert_message, AlertMessageType, AlertResponse, AlertMessageResponse
 
 
 def confirm(message: str) -> bool:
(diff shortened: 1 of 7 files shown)


bcipy/main.py
```diff
@@ -95,6 +100,7 @@ def execute_task(task: TaskType, parameters: dict, save_folder: str) -> bool:
     language_model = None
 
     fake = parameters['fake_data']
+    log.info(f"fake data is {'on' if fake else 'off'}")
```
tab-cmd: It might be helpful to know what fake data means... f"/n fake data is {'on. generating fake data for ' if fake else 'off. attempting to acquire data from'} {device_name}..."

and we could log in the case of fake data and signal model below: f"fake data is on, all decisions will be faked. no signal model required". We talked about making more nuianced fake or simulated behaviour, but this message would alert the user to the current behaviour.

bcipy/gui/alert.py
```diff
@@ -0,0 +1,26 @@
+"""GUI alert messages"""
+# pylint: disable=no-name-in-module
+import sys
+from PyQt5.QtWidgets import QApplication, QMessageBox
+
+
+def confirm(message: str) -> bool:
```
tab-cmd, line 8: Some duplicate functionality and enums exist in [gui_main](https://github.com/CAMBI-tech/BciPy/blob/main/bcipy/gui/gui_main.py)

It may be useful to use the enums or factor this out of BCIGui to be used elsewhere.

tab-cmd (approved):

bcipy/main.py
lawhead: Good suggestion. I moved the logging to the acquisition helper and added more detail. I would like to defer additional logging for fake_decisions for the other ticket.

bcipy/gui/alert.py
lawhead: Good callout. I factored out the code for constructing a MessageBox into its own function in GUI/main that's now used in both BCIGui and here. They each call the resulting message box in slightly different ways and have separate use cases, so I think it still makes sense for them to be distinct, but now this module uses functionality provided by the main module.

PR #220, opened 2022-05-12
ERP plotting improvements and MNE data support

tab-cmd:
# Overview

Upgraded MNE to v1.0. Refactoring `visualize_erp` to use MNE plotting. By default, offline_analysis.py will produce ERPs across target/nontarget conditions and display topomaps for better insights into data after a session. Upgrading MNE removed the annotataion_channel option from EDF conversion. To leverage this feature, an older version of BciPy will need to be used. Added options to our internal `RawData` class to convert to MNE RawArray and for channel-based methods to apply an optional transformations. 

New Integration tests were generated in offline analysis to help test the visualization code and prevent future changes to figures without re-generating. Other unit tests were added where applicable. 

In the future, we should support channel type definitions in our raw data. This can be leveraged in several places for differential data processing / viz. See, https://www.pivotaltracker.com/story/show/182106570 and https://www.pivotaltracker.com/story/show/182121313

## Ticket

[#180699095](https://www.pivotaltracker.com/story/show/180699095)

## Contributions

- `helpers/visualization.py`: `visualize_erp` updated to use MNE for average ERP plots + topomaps.
- `helpers/raw_data.py`: updated channel methods to accept transformation argument and return sampling rate. Added a convert to mne method that returns an MNE RawArray. Assumes standard_1020 eeg montage by default. Add by_channel map method to call by_channel and remove channels by map ([0, 1, 1]).
- `helpers/visualization.py`: added methods for visualized MNE Epochs and better visualizing EEG data in BciPy. `visualize_joint_average` and `visualize_evokeds`.
- `helpers/stimuli.py`: added a method for converting MNE RawArray to Epochs.
- `helpers/convert.py`: updated to use new version of MNE and pyedflib (required), deprecate unused `mode` arg.

## Test

- `make test-all`
- `make integration-test`
- Ran offline_analysis.py on collected data. Reviewed plots with CAMBI.

## Documentation

- Are documentation updates required? Yes. Relevant documentation added! 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Yes + some I forgot from my last PR 🙈 

## New Figures Generated

Using my data (which is the integration test default) on a different day. 

average_erp
non-target_topomap
target_topomap


bcipy/helpers/raw_data.py
```diff
@@ -58,22 +65,84 @@ def channels(self) -> List[str]:
...
-        return numeric_vals[:, 1:numeric_column_count].transpose()
+        data = numeric_vals[:, 1:numeric_column_count].transpose()
+        fs = self.sample_rate
+
+        if transform:
+            data, fs = self.apply_transform(data, transform)
+
+        return data, fs
+
+    def to_mne(
```
lawhead: What do you think about moving this functionality to the `convert` module?

lawhead (approved):

bcipy/helpers/raw_data.py
tab-cmd: My initial reasoning when we made the ticket was this would be the first of several MNE PRs to support better data processing/handling across the codebase- that we would start relying on their formats more heavily, deprecating some of the BciPy redundancies where we could. It made sense when I thought of it in that way to support going in-between RawData formats here. It does add some additional overhead to the RawData module. I'm open to rethinking how we handle data to support ease of use and quality data processing.

lawhead: I think it makes sense to leverage more of what MNE provides. The raw_data module could be a lot more efficient by  handling the underlying data as typed arrays (using pandas, numpy, or MNE RawArray) rather than reading everything as lists. However, in other areas of the codebase we provide our own interfaces and avoid committing too heavily on a single library (GUI module -> PyQt; Displays -> psychopy; acquisition -> LSL; etc), which helps us maintain a consistent API in case any of these things change.

We are using a separate module for data conversion functions, which I think is a good idea to separate concerns. It seems consistent to treat MNE as another data format that goes here. Instead of calling `raw_data.to_mne()` you might call `convert.to_mne(raw_data)` or `convert_to_mne(raw_data)`.

We might also think about whether conversion is necessary or if additional behavior should be provided by RawData (that could dispatch to MNE).

tab-cmd: > We are using a separate module for data conversion functions, which I think is a good idea to separate concerns. It seems consistent to treat MNE as another data format that goes here. Instead of calling raw_data.to_mne() you might call convert.to_mne(raw_data) or convert_to_mne(raw_data).

That seems reasonable to me while we are discussing MNE integration and future multimodal needs. I'll make the change!

> We might also think about whether conversion is necessary or if additional behavior should be provided by RawData (that could dispatch to MNE).

Another consideration is that we may start using devices not ideal or yet integrated (tripolars) for MNE use... we will likely have to abstract several things, but ultimately we need better support for common BCI data needs (exports by varying dimensions & data type, filtering out channels by map or type, setting of sensor coordinates).

lawhead:
The visualizations look great and I think they will be really valuable.

bcipy/helpers/convert.py
```diff
@@ -352,3 +342,25 @@ def tar_name_checker(tar_file_name: str) -> str:
     if tar_file_name.endswith('.tar.gz'):
         return tar_file_name
     return f'{tar_file_name}.tar.gz'
+
+
+def convert_to_mne(
+        raw_data: RawData,
+        channel_map: Optional[List[int]],
+        transform: Optional[Composition] = None,
```
nik-sm, line 350: Could use `Optional[Callable]`, though it doesn't matter much

bcipy/helpers/convert.py
```diff
@@ -352,3 +342,25 @@ def tar_name_checker(tar_file_name: str) -> str:
...
+        montage: Optional[str] = 'standard_1020') -> RawArray:
```
nik-sm: I think this should just be `montage: str = 'standard_1020'` 
(it's optional in the colloquial sense, but not `Optional` because `None` would presumably cause a problem - https://docs.python.org/3/library/typing.html#typing.Optional)

bcipy/helpers/raw_data.py
```diff
@@ -18,10 +19,13 @@ class RawData:
     def __init__(self,
                  daq_type: str = None,
                  sample_rate: int = None,
-                 columns: List[str] = None):
+                 columns: List[str] = None,
+                 column_types: List[str] = None):
```
nik-sm: These args with default None should be Optional[List[str]]

nik-sm: I guess the ones above should also be `Optional[str]` and `Optional[int]`

bcipy/helpers/raw_data.py
```diff
@@ -58,22 +62,63 @@ def channels(self) -> List[str]:
...
 
         numeric_data = self.numeric_data
 
         numeric_vals = numeric_data.values
         numeric_column_count = numeric_vals.shape[1]
-        return numeric_vals[:, 1:numeric_column_count].transpose()
+        data = numeric_vals[:, 1:numeric_column_count].transpose()
+        fs = self.sample_rate
+
+        if transform:
```
nik-sm, line 94: Safer to do `if transform is not None`

bcipy/helpers/raw_data.py
```diff
@@ -58,22 +62,63 @@ def channels(self) -> List[str]:
...
+        data: C x N numpy array with samples where C is the number of channels and N
+        is number of time samples
+        fs: resulting sample rate if any transformations applied"""
 
         numeric_data = self.numeric_data
 
         numeric_vals = numeric_data.values
         numeric_column_count = numeric_vals.shape[1]
-        return numeric_vals[:, 1:numeric_column_count].transpose()
+        data = numeric_vals[:, 1:numeric_column_count].transpose()
```
nik-sm: Here you're slicing off the first column (`1:...`) - I see it's the same as line 60 above - it removes timestamps. Is it worth making a method for this?

bcipy/helpers/visualization.py
```diff
@@ -1,121 +1,67 @@
...
-    # set upper and lower bounds of figure legend
-    lower = 0
-    upper = data_length / fs * 1000  # convert to seconds
-
-    fig = plt.figure(figsize=(20, 10))
-    ax1 = fig.add_subplot(211)
-
+    mne_data = convert_to_mne(raw_data, channel_map=channel_map, transform=transform)
+    epochs = mne_epochs(mne_data, trigger_timing, trial_length, trigger_labels)
+    epochs = (epochs['1'], epochs['2'])
```
nik-sm, line 57: How come these are `1` and `2`, rather than `0` and `1`? 
Also - what determines the order of target, non-target? (I see that this order is important in the functions called below)

bcipy/helpers/visualization.py
```diff
@@ -168,47 +114,48 @@ def visualize_csv_eeg_triggers(trigger_col=None):
...
+        plot_joint_times: Optional[List[float]] = [-0.1, 0, 0.2, 0.3, 0.35, 0.4, 0.5],
+        save_path: Optional[str] = None,
+        show: Optional[bool] = False) -> List[Figure]:
+    """Visualize Joint Average.
+
+    Plots channel average for each condition and provides corresponding topomaps for times defined in plot_joint_times.
+
+    See: https://mne.tools/dev/generated/mne.Evoked.html#mne.Evoked.plot_joint
+
+    Note: assumes the number of Epochs (passed as a tuple)
```
nik-sm: Incomplete sentence?

bcipy/signal/model/offline_analysis.py
```diff
@@ -137,11 +138,12 @@ def offline_analysis(
...
     # The timestamp column [0] is already excluded.
     channel_map = analysis_channels(channels, type_amp)
+    data, fs = raw_data.by_channel()
 
     inquiries, inquiry_labels, inquiry_timing = model.reshaper(
         trial_targetness_label=trigger_targetness,
         timing_info=trigger_timing,
-        eeg_data=raw_data.by_channel(),
+        eeg_data=data,
         fs=sample_rate,
```
nik-sm: should this be `fs` to avoid confusion? 
(It seems that sample rate is retrieved twice above, once into `sample_rate` and once into `fs`.)

Alternatively, could do `data, _ = raw_data.by_channel()`

nik-sm (approved):
Few minor comments, but overall looks good!

bcipy/helpers/visualization.py
tab-cmd: This is determined by the trigger_labels given. In this case we provide: 0==nontarget and 1==target. Accessing the epochs maintains the order given, but indexing does not start at 0. This is an MNE implementation detail.

bcipy/helpers/raw_data.py
tab-cmd: It seems alright to filter out other Falsey things in this check ?  It could be useful to assert it is a Composition here more explicitly if you would want to see it more sensitive to the inputs expected.

bcipy/helpers/convert.py
tab-cmd: Nice catch! I wasn't sure when I did it and should have looked it up

bcipy/helpers/convert.py
tab-cmd: It would need to be a Composition, or at the very least a Callable that returns data and sampling rate.

bcipy/helpers/raw_data.py
tab-cmd: I'm unsure! @lawhead what are your thoughts?

bcipy/helpers/visualization.py
tab-cmd: I can add this to the docstring!

bcipy/signal/model/offline_analysis.py
tab-cmd: We did this to avoid re-assigning to the same name after transformation: https://github.com/CAMBI-tech/BciPy/blob/86692d099f1c5857de858fe42ce2637280bb3bdd/bcipy/signal/model/offline_analysis.py#L155

There are cases we might need the original and also transformed sample rates. I think sample_rate may be a better kwarg name as it's explicit. What do you think?

bcipy/helpers/raw_data.py
lawhead: Sure, maybe a `channel_data` property?

bcipy/helpers/convert.py
nik-sm: Oh true - arbitrary callable would not work. Could add a base class for `Composition` and the other individual transforms, but definitely not urgent.

bcipy/signal/model/offline_analysis.py
nik-sm: Ok I see - so `sample_rate` will continue to be the original, and `fs` eventually becomes the downsampled rate.
Yes I agree that more explicit kwarg name is better.

PR #222, opened 2022-05-20
#182040775 ; session summary

lawhead:
# Overview

Added the option to generate session summary charts (session.xlsx) at the end of a copy phrase task.

## Ticket

https://www.pivotaltracker.com/story/show/182040775

## Contributions

- Added a  `summarize_session` parameter to optionally write the charts at the end of a copy phrase task.
- Refactored Session data to store more context (symbol set and decision threshold) so it could be interpreted without viewing the associated parameters.
- Refactored the session helper to split the generation of the sqlite database into two functions: one to load the session data from json and one to generate the database from a Session.
- Added checks to only generate session charts if evidence data is present.

## Test

- Ran Copy Phrase task with and without the `summarize_session` parameter enabled. Also tested this with`fake_data`. Ran all unit tests.


bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -420,6 +425,14 @@ def execute(self) -> str:
             self.parameters['preview_inquiry_progress_method'],
             self.trigger_handler.file_path).as_dict()
         self.write_session_data()
+
+        # Evidence is not recorded in the session when using fake decisions.
+        if self.parameters['summarize_session'] and self.session.has_evidence():
+            session_db(data_dir=self.file_save,
```
tab-cmd: This can be captured in another ticket, but it may not be expected that we output a .db file when using that parameter and it could be extraneous for most users / cases. Since they rely on each other, it could be wrapped as an export tool, manage its internal formats and then export to the desired ones. It's not a huge file so I'm not too concerned.

tab-cmd (approved):

bcipy/task/paradigm/rsvp/copy_phrase.py
lawhead: Yes, good point. I refactored the session helper so that the functionality for generating the evidence records is no longer coupled with the code for creating the database. Now we can generate the spreadsheet without first creating a db.

tab-cmd (approved):

PR #224, opened 2022-05-31
Signal Model Path Parameter

lawhead:
# Overview

Added a parameter to specify the `signal_model_path`. If provided, this path is used when loading the signal model for typing tasks.

## Ticket

https://www.pivotaltracker.com/story/show/182005453

## Contributions

- New parameter to specify the signal model path.
- Updated the `ask_file` utility to accept an optional directory parameter from which to start.
- Updated the `load` helper.

## Test

- Ran Copy Phrase task without the parameter set to ensure that the original behavior did not change.
- Ran a copy Copy Phrase task with the new parameter set to my calibration directory. Confirmed that the dialog displayed this directory when initialized.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Better handling of the `fake_data` parameter to avoid erroneous recordings. Added a confirmation dialog when the `fake_data` parameter is set to `True` to alert users that they are in a system test mode. #219
 - `task/data.py`: session data contains more contextual data for interpreting the results, including the `symbol_set` and the `decision_threshold` used.
 - `parameters/parameters.json`: added `summarize_session` parameter is used to output richer session summaries after a Copy Phrase task.
+- `parameters/parameters.json`: added the `signal_model_path` parameter. When set this can be used for loading the pre-trained signal model during typing sessions.
 
 
 ### Updated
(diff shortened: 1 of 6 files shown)


tab-cmd (approved):
What do you think about offline_analysis.py optionally updating the passed parameter file and outputting the save location to the terminal after training? The current workflow would be to calibrate, train, and update the parameters file manually. If you have multiple parameter files this would still be the case after training, so not sure it can solve all workflows.... This case may be another rationale for a changed parameters only approach for experiments (instead of all operating parameters). Then you could change defaults that apply to all conditions and more easily see how they differed.

lawhead:
> What do you think about offline_analysis.py optionally updating the passed parameter file and outputting the save location to the terminal after training? The current workflow would be to calibrate, train, and update the parameters file manually. If you have multiple parameter files this would still be the case after training, so not sure it can solve all workflows.... This case may be another rationale for a changed parameters only approach for experiments (instead of all operating parameters). Then you could change defaults that apply to all conditions and more easily see how they differed.

It's an interesting idea. It would be nice to automate the workflow. However, offline analysis would be working against the parameters file in the calibration directory, so I'm not sure that updating that would be beneficial. I like the idea of a changed parameter file. I would like to give that a bit more thought.

PR #227, opened 2022-06-16
Internet connection alert

lawhead:
# Overview

Added an alert when the user's computer has an active internet connection.

## Ticket

https://www.pivotaltracker.com/story/show/182200897

## Contributions

- New system utility for checking for an internet connection
- Factored out the functionality for generating the confirmation dialog so multiple conditions could be checked.
- Alert the user when an active internet connection is detected.
- Added GPU logging to the system_info.

## Test

- Ran a session with the internet turned on to confirm a new alert. Ran again with the internet off to test that the alert did not appear. Inspected the log files for the new system_info.
- I was unable to fully test the GPU logging since this is not supported by PyTorch on my computer.

### DIFF
--- a/bcipy/helpers/system_utils.py
+++ b/bcipy/helpers/system_utils.py
@@ -1,17 +1,20 @@
-import sys
+"""Utilities for system information and general functionality that may be
+shared across modules."""
+import importlib
+import logging
 import os
-
-from cpuinfo import get_cpu_info
-from typing import Optional, Tuple, Callable
-from pathlib import Path
-import pkg_resources
+import pkgutil
 import platform
+import socket
+import sys
+import time
+from pathlib import Path
+from typing import Callable, Optional, Tuple
+
+import pkg_resources
 import psutil
 import pyglet
-import importlib
-import pkgutil
-import time
-import logging
+from cpuinfo import get_cpu_info
 
 DEFAULT_ENCODING = 'utf-8'
 DEFAULT_EXPERIMENT_ID = 'default'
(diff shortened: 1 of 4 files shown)


bcipy/helpers/system_utils.py
```diff
@@ -1,24 +1,46 @@
...
 DEFAULT_ENCODING = 'utf-8'
 DEFAULT_EXPERIMENT_ID = 'default'
 EXPERIMENT_FILENAME = 'experiments.json'
 FIELD_FILENAME = 'fields.json'
 DEFAULT_EXPERIMENT_PATH = '.bcipy/experiment/'
 DEFAULT_FIELD_PATH = '.bcipy/field/'
+REMOTE_SERVER = "https://github.com/CAMBI-tech/BciPy/"
+
+
+def is_connected(hostname: str = "one.one.one.one", port=80) -> bool:
```
tab-cmd: REMOTE_SERVER could be the default here

bcipy/main.py
```diff
@@ -22,6 +22,31 @@
 log = logging.getLogger(__name__)
 
 
+def preconditions_ok(parameters) -> bool:
```
tab-cmd: What do you think of moving this to `helpers/validate.py`. We use a similar pattern for validating experiments. Another option for the method name: `validate_bcipy_session` or similar to be more explicit

tab-cmd (approved):

bcipy/helpers/system_utils.py
lawhead: Turns out this doesn't work if I'm on OHSU's VPN and it thought I was not on the internet. I updated this to use the "1.1.1.1" DNS resolver and it works both on and off VPN.

PR #232, opened 2022-07-25
added ssvep to cvep method

AidaFakhry:
# Overview

A method that uses the refresh rate on the monitor and returns a code for SSVEP. every frame of / on for x seconds.

## Ticket 

[Ticket 2](https://www.pivotaltracker.com/n/projects/2460065/stories/182492679)

## Contributions

- Created method that uses the refresh rate on the monitor and returns a code for SSVEP. every frame of / on for x seconds.

## Test

- Unit tested.

## Documentation

Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Verify the updates you did here.

## Changelog

Is the CHANGELOG.md updated with your detailed changes? N/a

### DIFF
--- a/bcipy/helpers/stimuli.py
+++ b/bcipy/helpers/stimuli.py
@@ -705,23 +705,41 @@ def get_fixation(is_txt: bool) -> str:
         return DEFAULT_FIXATION_PATH
 
 
-def ssvep_to_cvep(refresh_rate: int = 60, flicker_rate: float = 10.0):
-    if flicker_rate > refresh_rate:
-        raise Exception("flicker rate cannot be greater than refresh rate")
+def ssvep_to_code(refresh_rate: int = 60, flicker_rate: int = 10) -> List[int]:
+    """Convert SSVEP to Code.
 
+    Converts a SSVEP (steady state visual evoked potential; ex. 10 Hz) to a code (0,1)
+    given the refresh rate of the monitor (Hz) provided and a desired flicker rate (Hz).
+
+    Parameters:
+    -----------
+        refresh_rate: int, refresh rate of the monitor (Hz)
+        flicker_rate: int, desired flicker rate (Hz)
+    Returns:
+    --------
+        list of 0s and 1s that represent the code for the SSVEP on the monitor.
+    """
+    if flicker_rate > refresh_rate:
+        raise BciPyCoreException('flicker rate cannot be greater than refresh rate')
+    if flicker_rate <= 1:
+        raise BciPyCoreException('flicker rate must be greater than 1')
+
+    # get the number of frames per flicker
     length_flicker = refresh_rate / flicker_rate
 
-    # determine if lenth_flickerr is an integer using .is_integer()
-    if not length_flicker.is_integer():
-        raise Exception("Provided flicker rate must divide into refresh rate.")
+    if length_flicker.is_integer():
+        length_flicker = int(length_flicker)
+    else:
+        err_message = f'flicker rate={flicker_rate} is not an integer multiple of refresh rate={refresh_rate}'
+        log.exception(err_message)
+        raise BciPyCoreException(err_message)
 
+    # start the first frames as off (0) for length of flicker;
+    # it will then toggle on (1)/ off (0) for length of flicker until all frames are filled for refresh rate.
     t = 0
-    test_array = []
-    for _ in range(int(length_flicker)):
-        for _ in range(int(length_flicker)):
-            test_array.append(t)
-        if t > 0:
-            t = 0
-        else:
-            t = 1
-    return test_array
+    codes = []
+    for _ in range(flicker_rate):
+        codes += [t] * length_flicker
+        t = 1 - t
+
+    return codes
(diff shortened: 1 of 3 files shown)


bcipy/helpers/stimuli.py
```diff
@@ -679,3 +679,25 @@ def get_fixation(is_txt: bool) -> str:
         return '+'
     else:
         return DEFAULT_FIXATION_PATH
+
+
+def ssvep_to_cvep(refresh_rate: int = 60, flicker_rate: float = 10.0):
```
lawhead: We need some documentation for this function to give a summary of what it's for (particularly if someone doesn't know what SSVEP and CVEP are abbreviations for), it's parameters (units?), etc.

Also, there should be a return type (`List[float]`), and the documentation should explain what this represents.

bcipy/helpers/stimuli.py
```diff
@@ -679,3 +679,25 @@ def get_fixation(is_txt: bool) -> str:
...
+    if flicker_rate > refresh_rate:
+        raise Exception("flicker rate cannot be greater than refresh rate")
+
+    length_flicker = refresh_rate / flicker_rate
+
+    # determine if lenth_flickerr is an integer using .is_integer()
```
lawhead: There's a typo in the comment: `lenth_flickerr` -> `length_flicker`. But you can probably just remove this comment since it's redundant with the code.

bcipy/helpers/stimuli.py
```diff
@@ -679,3 +679,25 @@ def get_fixation(is_txt: bool) -> str:
...
+    if not length_flicker.is_integer():
+        raise Exception("Provided flicker rate must divide into refresh rate.")
```
lawhead: `length_flicker` is only used as a float to check the whether it's evenly divisible so we might want to immediately cast it as an int:

    if length_flicker.is_integer():
        length_flicker = int(length_flicker)
    else:
         raise Exception("Provided flicker rate must divide into refresh rate.")

bcipy/helpers/stimuli.py
```diff
@@ -679,3 +679,25 @@ def get_fixation(is_txt: bool) -> str:
...
+    t = 0
+    test_array = []
+    for _ in range(int(length_flicker)):
+        for _ in range(int(length_flicker)):
+            test_array.append(t)
+        if t > 0:
+            t = 0
+        else:
+            t = 1
+    return test_array
```
lawhead: Python has a couple of list methods that may be helpful here. The `+` operator can be used to concatenate lists (`[1,2] + [3,4]` => `[1,2,3,4]`), and the `*` operator is a shortcut for constructing a list of consecutive elements with the same value (`[0] * 3` => `[0,0,0]`).

Another observation is that we're basically just toggling between `0` (for even indices) and `1` (for odd indices).  We can determine if a number is even or odd by using the modulus operator `%`.

So with all that, one way to shorten the above code is something like:

    test_array = []
    for i in range(length_flicker):
        test_array += [i % 2] * length_flicker

lawhead (changes requested):
I don't understand what the returned list represents in the `ssvep_to_cvep` function and how the calling function would use the results. I would expect that for a given refresh_rate it would return the same number of items, but

    ssvep_to_cvep(refresh_rate=50, flicker_rate=10)
    => [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0]
    ssvep_to_cvep(refresh_rate=50, flicker_rate=25)
    => [0, 0, 1, 1]

If the intent is to communicate a flash pattern, we need to know how much time each entry in the list represents, so the units are important to consider. For instance, if we are returning a pattern for 1 second display, the items in the list could each be 1hz worth of data (50 items for a 50hz refresh rate). Alternatively, this could be a tuple with the on/off duration in seconds . For the second example, this would be (0.5, 0.5) for half second off, followed by half second on.

@sci-tab, how would it be easiest to consume this data when writing displays?

tab-cmd:
> I don't understand what the returned list represents in the `ssvep_to_cvep` function and how the calling function would use the results. I would expect that for a given refresh_rate it would return the same number of items, but
> 
> ```
> ssvep_to_cvep(refresh_rate=50, flicker_rate=10)
> => [0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 0, 0, 0, 0]
> ssvep_to_cvep(refresh_rate=50, flicker_rate=25)
> => [0, 0, 1, 1]
> ```
> 
> If the intent is to communicate a flash pattern, we need to know how much time each entry in the list represents, so the units are important to consider. For instance, if we are returning a pattern for 1 second display, the items in the list could each be 1hz worth of data (50 items for a 50hz refresh rate). Alternatively, this could be a tuple with the on/off duration in seconds . For the second example, this would be (0.5, 0.5) for half second off, followed by half second on.
> 
> @sci-tab, how would it be easiest to consume this data when writing displays?

That is the intended output and how the VEP display would be able to use it. I'll pick it up from here and see if I can spot the issue!

lawhead (approved):

PR #235, opened 2022-07-29
User ID reset bug and cleanup

tab-cmd:
# Overview

This PR fixes the user ID input bug where the user input in GUI would reset after editing a parameter file. Modifies the offline analysis call to use the correct parameters link (as opposed to the default). Added alert functionality to the GUI to let experimenters know when an experimental session has been completed. 

## Ticket

- https://www.pivotaltracker.com/story/show/182776444
- https://www.pivotaltracker.com/story/show/178695784
- https://www.pivotaltracker.com/story/show/182101849

## Contributions

- `BCInterface.py`: added refresh argument to `update_user_list` method to prevent resetting after parameter changes. Added a play_sound method after an experiment session has been completed.  Added parameter flag to offline analysis call. Factor out static_font_size to help with later GUI updates.
- `parameters.json`: updated name `offline_analysis_tone` to `alert_sound_file` to be more widely applicable 
- `helpers/stimuli.py`: remove post stim buffer (duration calculation more sufficient)

## Test

- went through the AC of the ticket
- `make test-all`
- Ensured experiment could run and alert plays when self.alert=True
- Ensured offline analysis still runs as expected from the GUI

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes?

## Changelog

### DIFF
--- a/bcipy/gui/BCInterface.py
+++ b/bcipy/gui/BCInterface.py
@@ -15,6 +15,7 @@ from bcipy.gui.main import (
 )
 from bcipy.helpers.load import load_json_parameters, load_experiments, copy_parameters, load_users
 from bcipy.helpers.parameters import DEFAULT_PARAMETERS_PATH
+from bcipy.helpers.stimuli import play_sound
 from bcipy.task import TaskType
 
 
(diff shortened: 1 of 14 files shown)


bcipy/gui/BCInterface.py
```diff
@@ -46,10 +47,14 @@ def __init__(self, *args, **kwargs):
         self.experiment_input = None
         self.task_input = None
 
+        # These reprsent the current user, experiment, and task selected in the gui
```
lawhead: Typo in the comment: 'reprsent' => 'represent'

lawhead (approved):

PR #237, opened 2022-08-23
VEP Display

tab-cmd:
# Overview

Added a VEP Display component. 

**Note**: while I've made a method for animation and a variable referencing vep_type, these need more direction from the clinical and engineering teams (4x4 vs 6x6; how it should animate). I want to preserve all of the spike work we've done thus far without committing to fancy animations or code generation at this time. Please add any suggestions that will make future changes easier.

## Ticket

https://www.pivotaltracker.com/story/show/182457440

## Contributions

- `display/paradigm/vep.py`: `VEPDisplay`: main display component needed to operate in any task. `VEPBox`: Bounding box for every single flickering box. `create_vep_codes`: method to create a random code for displaying the vep flickering.
- `display/paradigm/rsvp.py`: Add typing / linting.
- `bcipy/display.py`: added `VEPStimulusProperties` to better encapsulate the stimuli needed for VEP operation
- `display/demo/vep/demo_calibration_vep.py`:  a script to demonstrate how to set up and use the VEPDisplay in a task type setting.

## Test

- `display/demo/vep/demo_calibration_vep.py` Ran and made sure everything hit AC. No other parts of the code were touched. 

https://user-images.githubusercontent.com/11506578/186269881-fb76f065-0c9e-4d11-bbdd-e0ba47f56ebd.MOV

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Updated in-line. Some readme or other documentation could be helpful, please point it out! 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Will do!


bcipy/display/demo/vep/demo_calibration_vep.py
```diff
@@ -0,0 +1,63 @@
...
+    stim_pos=start_positions_for_boxes,
+    stim_height=0.1,
+    stim_font='Consolas',
+    timing=(1, 0.5, 4),  # prompt, fixation, stimuli
+    stim_length=1,  # how many times to stimuli
+)
+vep = VEPDisplay(win, experiment_clock, stimuli, task_display, info)
+timing = []
+t = 2
+for i in range(len(task_text)):
```
lawhead: Alternatively, this could be:

    for (txt, color) in zip(task_text, task_color):
        vep.update_task(txt, color[0])

bcipy/display/paradigm/vep/display.py
```diff
@@ -0,0 +1,491 @@
...
+import numpy as np
+from psychopy import visual, core
+from bcipy.helpers.clock import Clock
+from bcipy.helpers.task import SPACE_CHAR
+from bcipy.helpers.triggers import TriggerCallback, _calibration_trigger
+from bcipy.helpers.stimuli import resize_image, get_fixation
+
+
+def create_vep_codes(length=32, count=4) -> List[List[int]]:
+    """Create a list of VEP codes."""
```
lawhead: It would be good to document what `length` and `count` are used for so it's clear when used elsewhere.

bcipy/display/paradigm/vep/display.py
```diff
@@ -0,0 +1,491 @@
...
+    res = []
+    np.random.seed(1)
+    for _ in range(count):
+        res.append(np.random.randint(2, size=length))
+    return res
```
lawhead: This could be reduced to a list comprehension:

    return [np.random.randint(2, size=length) for _ in range(count)]

bcipy/display/paradigm/vep/display.py
```diff
@@ -0,0 +1,491 @@
...
+        self.stimuli_inquiry = stimuli
+        self.stimuli_timing = timing
+        self.stimuli_colors = colors
+
+    def _build_text_boxes(self) -> List[visual.TextBox2]:
+        """Build the text boxes for the experiment.
+
+        This assumes that vep_type is 4 using a 2x2 grid. Additional configurations can be added in the future.
+        """
+        # generate the inquiry stimuli
```
lawhead, line 390: Is the idea that we will have a fixed number of configurations, or did you have thoughts on making this more general and be able to accommodate any possible grid?

lawhead (approved):
It's exciting to see this working and this gives us is a great start from which to iterate.

bcipy/display/paradigm/vep/display.py
tab-cmd: I really only considered having a few fixed layouts with different size flickering VEP boxes. I'd be open to something more configurable within a fixed unit type (height or norm). We could try looping in code generated via psychopy to give more options. When coding this, I realized we aren't really equipped across the board to support interface building for non-coding folks (only configuration). 

If a user wanted something more custom, I would recommend they build it in the psychopy experiment builder GUI and import bcipy to get the VEP flickering and/or acquisition.

PR #238, opened 2022-08-26
#183037159 ; alert when using battery power

lawhead:
# Overview

Added an alert to check for battery power before performing a task.

## Ticket

https://www.pivotaltracker.com/story/show/183037159

## Contributions

- New system utility to check for battery power
- New validation entry

## Test

- Ran unit tests
- Ran calibration task both with laptop plugged in and once while unplugged to confirm that the alert appears correctly.

### DIFF
--- a/bcipy/helpers/system_utils.py
+++ b/bcipy/helpers/system_utils.py
@@ -43,6 +43,12 @@ def is_connected(hostname: str = "1.1.1.1", port=80) -> bool:
     return False
 
 
+def is_battery_powered() -> bool:
+    """Check if this current computer is a laptop currently using its battery."""
+    return psutil.sensors_battery(
+    ) and not psutil.sensors_battery().power_plugged
+
+
 def git_dir() -> str:
     """Git Directory.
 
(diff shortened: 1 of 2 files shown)


bcipy/helpers/system_utils.py
```diff
@@ -43,6 +43,12 @@ def is_connected(hostname: str = "1.1.1.1", port=80) -> bool:
     return False
 
 
+def is_battery_powered() -> bool:
+    """Check if this current computer is a laptop currently using its battery."""
```
tab-cmd: Some more context may be useful! `Returns True if the computer is currently running on battery power. This can impact the performance of hardware (ex. GPU) needed for BciPy operation by entering power saving operations.`

bcipy/helpers/validate.py
```diff
@@ -25,7 +26,8 @@ def validate_bcipy_session(parameters: dict) -> bool:
     True if it's okay to continue, otherwise False
     """
     possible_alerts = [(parameters['fake_data'], '* Fake data is on.'),
-                       (is_connected(), '* Internet is on.')]
+                       (is_connected(), '* Internet is on.'),
+                       (is_battery_powered(), '* Operating on battery power')]
```
tab-cmd, line 30: This pattern is really coming in handy. It's clean 😎

tab-cmd (approved):

PR #245, opened 2022-09-07
Changed parameters

lawhead:
# Overview

Baseline programming for viewing changed parameters, including change set calculations and preliminary GUI work. This also fixes some bugs in the parameters form.

## Ticket

- https://www.pivotaltracker.com/story/show/175193924
- https://www.pivotaltracker.com/story/show/183134229
- https://www.pivotaltracker.com/story/show/183133776

## Contributions

- Added calculations for computing the differences between two parameters files.
- GUI widgets for displaying a parameter change set.
- Bug fixes for GUI float widgets which were rounding some parameters to predetermined minimum and maximum values.
- Added code to compute the number of decimals to display (and button increment value) for float inputs based on the initial value provided.

## Test

- `make unit-test`
- Verified that the parameters form displays the correct default values for 'High Voltage Threshold Value' and 'Low Voltage Threshold Value'
- Verified that the parameters form shows 3 decimal places for 'Static Trigger Offset' field.


lawhead:
You may want to review this using the 'Hide Whitespace' parameter. Some of the files, such as parameters.json, had odd line endings that were removed (<CR>).

bcipy/gui/main.py
```diff
@@ -324,6 +325,8 @@ def __init__(self, **kwargs):
     def init_control(self, value):
         """Override FormInput to create a spinbox."""
         spin_box = QSpinBox()
+        # See https://www.pivotaltracker.com/story/show/183134720
```
tab-cmd: We've not put links to pivotal in the past. I think a more direct comment or nothing would be ok here.

bcipy/helpers/parameters.py
```diff
@@ -1,230 +1,277 @@
...
+        """
+        updated = False
+        existing_keys = self.keys()
+        for key, val in parameters.entries():
+            if key not in existing_keys:
+                self.add_entry(key, val)
+                updated = True
+        return updated
+
+    def changes_from(self, parameters) -> Dict[str, ParameterChange]:
```
tab-cmd: `diff` would be another good name for this. ++ type

bcipy/helpers/parameters.py
```diff
@@ -1,230 +1,277 @@
...
+        """Lists the differences between this and another set of parameters.
+        A None original_value indicates a new parameter.
+
+        Parameters
+        ----------
+            parameters - set of parameters for comparison; these are considered
```
tab-cmd: Doesn't have to be original values? It could be default values like in the GUI or a diff against another parameter file.

bcipy/gui/parameters/params_form.py
```diff
@@ -21,7 +22,9 @@
     static_text_control,
     TextInput,
 )
-from bcipy.helpers.parameters import Parameters
+
+HELP_SIZE = 12
```
tab-cmd: Given this is used only twice, it might be okay to leave it in the method below for now. You do it for the lbl

bcipy/gui/main.py
```diff
@@ -336,6 +339,29 @@ def cast_value(self) -> str:
         return None
 
 
+class FloatInputProperties(NamedTuple):
+    """Properties used when constructing a FloatInput component. Default values
+    originate from the pyqt5 QDoubleSpinBox component defaults."""
+
+    min: float = -sys.float_info.max  # 0.0 is the component default
```
tab-cmd, line 345: 👍🏼

bcipy/gui/parameters/params_form.py
```diff
@@ -122,6 +125,153 @@ def update_parameters(self):
...
+
+            lbl = static_text_control(
+                None,
+                label=f"* {param['readableName']}: {param['value']}",
+                size=13,
+                color="darkgreen")
+
+            original_value = static_text_control(
+                None,
+                label=f"(original value: {param_change.original_value})",
```
tab-cmd: In our case, it's not a change from the original value it's a comparison to the default parmeters.json if I'm understanding it correctly. It might be more clear to put `default value:` or update this to handle a loaded parameter change.

bcipy/gui/parameters/params_form.py
```diff
@@ -162,6 +313,10 @@ def initUI(self):
         self.load_msg.setVisible(False)
         vbox.addWidget(SearchInput(self.search_form))
 
+        self.changes_panel = QHBoxLayout()
```
tab-cmd, line 320: We should demo to the clinical group, but I would think this section could collapse altogether if no changes are available.

bcipy/gui/parameters/params_form.py
```diff
@@ -122,6 +125,153 @@ def update_parameters(self):
...
+        self.changes_area.setHorizontalScrollBarPolicy(Qt.ScrollBarAlwaysOff)
+        self.changes_area.setWidgetResizable(True)
+        self.changes_area.setWidget(self.change_items)
+
+        control_box = QHBoxLayout()
+        control_box.addWidget(
+            static_text_control(None, label='Changed Parameters:'))
+        self.toggle_button = QPushButton(self.hide_label)
+        self.toggle_button.setFlat(True)
+        self.toggle_button.setFixedWidth(40)
```
tab-cmd, line 247: We should make this the same as the other widths. Right now the boxes are a little off. 
new_loaded_parameter

It could be another ticket though because it seems like this and the search scale with the window but the editable parameters window stays fixed

tab-cmd (approved):
Seems to work well! The behavior is a little unclear after loading a new parameter file while already editing. 
changed_parameter_before_save
 I might expect a diff between the loaded one to what I'm editing, as opposed to the default.  Maybe we would assume no loading on that screen or that would provide a new parameter file to edit?

bcipy/gui/parameters/params_form.py
lawhead: Wow, that's really different than on Macs. Yes, I was planning on some more UI refinement on this one.

bcipy/gui/parameters/params_form.py
lawhead: I was wondering about the best defaults for this. I would like to get some input on it. I'm fine adjusting this to be collapsed if there are no initial values.

PR #246, opened 2022-09-12
Convert: add BDF support and fix static offset bug

tab-cmd:
# Overview

Refactored the convert to edf method. Add convert to bdf method for better export resolution. Ensured the static offset output in the parameters file is accounted for. Add a pre-filter option that will apply the default transform using the filter parameters in the parameters.json stored in the session directory. Ensured the correct annotation channels are used. Upgraded pyedflib to latest version. 

## Ticket

https://www.pivotaltracker.com/story/show/182395017

## Contributions

- `convert.py`: Refactored the convert to edf method into reusable parts (`pdf_convert`). Add convert to bdf method for better export resolution. Ensured the static offset output in the parameters file is accounted for. Update documentation. Calculate the correct annotation channels needed instead of only alerting. 
- `demo_convert.py`: Added a plot option for file viewing, using -d for the directory arg instead. Added the export to bdf option. 
- `test_convert.py`: refactored setup needed in the `create_bcipy_session_artifacts`. Add tests for added methods. 

## Test

- `make test-all`
- Used `demo_convert.py` and shared the data with @danielpklee for review. I shared a .edf and .bdf file. 

## Documentation

## Changelog

### DIFF
--- a/bcipy/helpers/convert.py
+++ b/bcipy/helpers/convert.py
@@ -62,14 +62,23 @@ def convert_to_edf(data_dir: str,
     if not edf_path.endswith('.edf'):
         raise ValueError(f'edf_path=[{edf_path}] must end in .edf')
 
-    data, channels, fs, events, annotation_channels = pyedf_convert(
+    data, channels, fs, events, annotation_channels, pre_filter = pyedf_convert(
         data_dir,
         write_targetness=write_targetness,
         use_event_durations=use_event_durations,
         remove_pre_fixation=remove_pre_fixation,
         pre_filter=pre_filter)
 
-    return write_pyedf(edf_path, data, channels, fs, events, FILETYPE_EDFPLUS, overwrite, annotation_channels)
+    return write_pyedf(
+        edf_path,
+        data,
+        channels,
+        fs,
+        events,
+        FILETYPE_EDFPLUS,
+        overwrite,
+        annotation_channels,
+        pre_filter)
 
 
 def convert_to_bdf(data_dir: str,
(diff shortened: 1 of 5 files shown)


bcipy/helpers/convert.py
```diff
@@ -28,65 +27,182 @@
...
+def convert_to_bdf(data_dir: str,
+                   bdf_path: str = None,
+                   overwrite: bool = False,
+                   write_targetness: bool = False,
+                   use_event_durations: bool = False,
+                   remove_pre_fixation: bool = False,
+                   pre_filter: bool = False) -> Path:
+    """ Converts BciPy data to the BDF+ filetype using pyEDFlib.
+
+    See https://www.biosemi.com/faq/file_format.html for more detailed information about the BDF specification.
```
lawhead: The link should end in `.htm`. Using `.html` gives a 404.

bcipy/helpers/convert.py
```diff
@@ -28,65 +27,182 @@
...
+    See https://www.teuniz.net/edfbrowser/ for a free EDF/BDF viewer.
+
+    Parameters
+    ----------
+    data_dir - directory which contains the data to be converted. This location must contain:
+        1. a parameter file.
+        2. a raw_data.csv file.
+        3. a trigger.txt file.
+    edf_path - optional; path to write converted data; defaults to writing a file named raw_data.edf in the data_dir.
```
lawhead: The documentation should be updated to use`bdf_path` and the file name should end in `.bdf`.

bcipy/helpers/convert.py
```diff
@@ -167,23 +285,103 @@ def write_edf(output_path: str,
...
+
+    Returns
+    -------
+        Path to new bdf file
+    """
+    if not overwrite and os.path.exists(output_path):
+        raise OSError('BDF file already exists.')
+
+    # set conversion parameters. The digital min/max are specific to BDF and much higher than EDF.
+    digital_min, digital_max = [-8388607, 8388607]
```
lawhead: The spec has the digital_min of -8388608.

However, edfwriter already has this information and initializes it in the `EdfWriter` constructor based on the provided `file_type` (see edfwriter.py line 209). We only need to provide specs for field values that we're overriding, so we actually don't need to include it here.

bcipy/helpers/convert.py
```diff
@@ -167,23 +285,103 @@ def write_edf(output_path: str,
...
+                'label': ch_names[i],
+                'dimension': 'uV',
+                'sample_frequency': sample_rate,
+                'physical_min': physical_min,
+                'physical_max': physical_max,
+                'digital_min': digital_min,
+                'digital_max': digital_max,
+                'transducer': '',
+                'prefilter': ''
+            }
```
lawhead: The only fields we are overriding from the default values are 'label', 'dimension', 'sample_frequency', 'physical_min', and 'physical_max'. The rest can be excluded.

bcipy/helpers/convert.py
```diff
@@ -167,23 +285,103 @@ def write_edf(output_path: str,
...
 
+    except Exception as error:
+        logger.info(error)
+        raise error
+    finally:
+        writer.close()
+    return output_path
+
+
+def write_bdf(output_path: str,
```
lawhead: There is a lot of redundancy between this method and `write_edf`. If we exclude the digital_min and digital_max (see other comment), the only difference is the file_type. Maybe that can be a parameter and we can have a single method with a default value of `FILETYPE_EDFPLUS`

lawhead (approved):

bcipy/helpers/convert.py
tab-cmd: I changed it thinking I had copied it wrong 😆

PR #249, opened 2022-09-21
#182031676 ; remove TCP-based acquisition code

lawhead:
# Overview

Removed TCP-based code from the acquisition module.

## Ticket

https://www.pivotaltracker.com/story/show/182031676

## Contributions

- Removed code that is no longer used since the switch to the `LslAcquisitionClient` for multi-modal.
- Refactored code that was previously using the `DeviceInfo` tuple to use a `DeviceSpec` to eliminate overlapping concepts.
- Fixed a bug in the `lsl_client` that occurred when it was used for acquiring data from a stream with an irregular rate (like a switch). This fixes the demo: `demo_eeg_with_switch.py`.
- Updated the acquisition module README to remove outdated documentation.

## Test

- Ran all unit tests
- Ran all acquisition module demos
- Ran the Data Viewer: 1. reading from streamed live data; 2. reading from file data.
- Ran a Calibration task with real data, followed by offline_analysis
- Ran a Copy Phrase task with real data.


bcipy/acquisition/datastream/lsl_server.py
```diff
@@ -145,6 +146,21 @@ def _settings(filename):
         return (daq_type, sample_hz, channels)
 
 
+def await_start(dataserver, max_wait=2):
```
tab-cmd: types needed

bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -26,24 +24,23 @@ class LslAcquisitionClient:
 
     Parameters
     ----------
-        max_buflen: the maximum length, in seconds, of data to be queried.
```
tab-cmd, line 29: could this be buffer_len?

tab-cmd (approved):
tested locally and works well! A few small additions or future points:

- __str__ method preferred for DeviceSpec: auto_str output difficult to parse in terminal alongside the ChannelSpec output.
- ChannelSpec fields shouldn't be None. We could default to uV and EEG safely for now? We also need to figure out setting those manually in the device spec, so maybe that comes first? https://www.pivotaltracker.com/story/show/182849439

lawhead:
> tested locally and works well! A few small additions or future points:
> 
> * **str** method preferred for DeviceSpec: auto_str output difficult to parse in terminal alongside the ChannelSpec output.
> * ChannelSpec fields shouldn't be None. We could default to uV and EEG safely for now? We also need to figure out setting those manually in the device spec, so maybe that comes first? https://www.pivotaltracker.com/story/show/182849439

I agree that the string representation of DeviceSpecs became pretty unwieldy after adding channel specs. I added a custom string representation for DeviceSpec and also changed the representation of ChannelSpec to only show fields with values. Regarding default values for ChannelSpec, I think this warrants a separate ticket since the default units are related to the content type.

bcipy/acquisition/protocols/lsl/lsl_client.py
lawhead: I was using the parameter name used by `StreamInlet`, but I agree that it doesn't follow our naming conventions. How about `max_buffer_len`? I think keeping 'max' in there is important since it could be smaller than this at any given time.

PR #255, opened 2022-10-20
#180155337 ; Params form Save As

lawhead:
# Overview

Added Save As behavior to the Parameters form.

## Ticket

https://www.pivotaltracker.com/story/show/180155337

## Contributions

- Added new button to the parameters form to enable users to save the current configuration as a different file.
- Changed the Load button to Open. Modified its behavior so that a new file is opened and loaded for modification. The previous Load implementation had unexpected behavior for most users. The loaded file was used to set the values in the GUI, but still required the user to Save these changes to overwrite them in the existing file. The changes in this PR make the behavior more consistent with conventional Mac and Windows GUIs.
- Updated code so that search is still enabled when another file is open and the search filter is correctly applied.

## Test

- Ran all unit tests.
- Save As test: Ran the params_form in default mode (using the standard parameters.json file). Made changes and saved them as another file. Confirmed that the changes were persisted and the default file was unmodified.
- Open test: Ran params_form in default mode. Opened a different parameters file. Made changes in the GUI and Saved results. Confirmed that the changes were persisted in the opened file and not in the default.
- Search tests: In the params GUI, searched for a term to filter the inputs. Opened a different parameters file which contained legacy inputs and confirmed that the filter was still correctly applied.

### DIFF
--- a/bcipy/gui/parameters/params_form.py
+++ b/bcipy/gui/parameters/params_form.py
@@ -8,6 +8,7 @@ from typing import Dict, Tuple
 from PyQt5.QtCore import Qt
 from PyQt5.QtWidgets import (QApplication, QFileDialog, QHBoxLayout,
                              QPushButton, QScrollArea, QVBoxLayout, QWidget)
+from bcipy.config import BCIPY_ROOT
 from bcipy.helpers.parameters import Parameters, changes_from_default
 
 from bcipy.gui.main import (
@@ -382,7 +383,7 @@ class MainPanel(QWidget):
         file_name, _ = QFileDialog.getOpenFileName(
             self,
             caption='Open parameters file',
-            directory='bcipy/parameters',
+            directory=f'{BCIPY_ROOT}/parameters',
             filter='JSON files (*.json)')
         if file_name:
             self.json_file = file_name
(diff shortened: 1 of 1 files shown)


bcipy/gui/parameters/params_form.py
```diff
@@ -388,6 +400,24 @@ def on_save(self):
...
             self.changes.repaint()
 
+    def on_save_as(self):
+        """Event handler for saving form data to another parameters file."""
+        options = QFileDialog.Options()
+        options |= QFileDialog.DontUseNativeDialog
+        filename, _ = QFileDialog.getSaveFileName(
+            self,
+            caption='Save As',
+            directory='bcipy/parameters/',
```
tab-cmd: pull this from config.py. At least the BCIPY_ROOT part to allow for running outside of root.

bcipy/gui/parameters/params_form.py
```diff
@@ -388,6 +400,24 @@ def on_save(self):
...
+            filter='JSON files (*.json)',
+            options=options)
```
tab-cmd, line 414: Interestingly, depending on OS this is not guaranteed to be returned with `.json`. I typed in a filename without it and it saved it to a file as `test`, not `test.json`. We should handle the no-extension or bad extension cases.

tab-cmd (approved):

bcipy/gui/parameters/params_form.py
lawhead: I think the simplest solution is to just append `.json` if the filename does not end with that. That's consistent with how a number of applications handle this (like Microsoft products, I believe).

PR #257, opened 2022-11-10
#180317983 ; GUI History

lawhead:
# Overview

Added functionality for storing state / user preferences between sessions. Dialogs which prompt for files and directories now remember the last location that was opened, as well as the directory of the last signal model that was trained or used.

## Ticket

https://www.pivotaltracker.com/story/show/180317983

## Contributions

- Added infrastructure for persisting state between work sessions in a .bcipy_cache file.
- Added pre-defined attributes for storing signal_model_directory and last_directory accessed.
- Updated offline_analysis to store the data_folder after generating a new model.
- Updated file_dialogs to use preferences if no specific locations have been provided in the parameters.
 
## Test

- Ran all unit and integration tests
- Ran an offline analysis and confirmed that the preference was written to the cache file.
- Ran a Copy Phrase and confirmed that it prompted for the model file using the appropriate directory.
- Tested the ask_filename and ask_directory methods from an interactive session.

## Discussion

I ended up using the name Preferences for this, but there were others that could have worked. I'm open to changing this if there are strong opinions. I considered the following:

- Preferences
- Settings
- Configuration - we already have config.py for environment variables
- State - too generic?
- Cache - usually refers to something that has been pre-computed
- History - too specific; we may want to store some other user-selected things like color preference, etc.
- Properties - too ambiguous?
- Session - we already use session to refer to a typing session


tab-cmd (approved):
This is a nice implementation! The clinical team will be really excited about this addition! I have a few suggestions:

1. I wonder if the preferences module should live outside of GUI? We could replicate similar behavior and add future caching needs to the CLI. 
2. Using the .bcipy/ directory for the cache?
3. If not moving preferences, we should move the test_preference.py to `gui/tests` from `gui/test/viewer/`. The GUI main tests should be up a level too, but that's outta scope here!

lawhead:
> This is a nice implementation! The clinical team will be really excited about this addition! I have a few suggestions:
> 
> 1. I wonder if the preferences module should live outside of GUI? We could replicate similar behavior and add future caching needs to the CLI.
> 2. Using the .bcipy/ directory for the cache?
> 3. If not moving preferences, we should move the test_preference.py to `gui/tests` from `gui/test/viewer/`. The GUI main tests should be up a level too, but that's outta scope here!

Good suggestions. I went ahead and moved it to the top level for now and put the cached file in the .bcipy directory.

PR #266, opened 2023-02-22
Device channel specs

lawhead:
# Overview

Set channel spec information in devices.json. Removed aliases.

## Ticket

https://www.pivotaltracker.com/story/show/182849439

## Contributions

- Updated devices.json to include channel spec information.
- Removed devices that were aliases.
- Updated code that was using the alias devices.

## Test

- Ran unit tests and demos.
- Ran an RSVP Calibration using streamed data.

### DIFF
--- a/bcipy/parameters/devices.json
+++ b/bcipy/parameters/devices.json
@@ -2,8 +2,32 @@
     {
         "name": "DSI-24",
         "content_type": "EEG",
-        "channels": [ "P3", "C3", "F3", "Fz", "F4", "C4", "P4", "Cz", "Pz", "Fp1", "Fp2", "T3",
-            "T5", "O1", "O2", "X3", "X2", "F7", "F8", "X1", "A2", "T6", "T4", "TRG" ],
+        "channels": [
+            { "name": "P3", "label": "P3", "units": "microvolts", "type": "EEG" },
+            { "name": "C3", "label": "C3", "units": "microvolts", "type": "EEG" },
+            { "name": "F3", "label": "F3", "units": "microvolts", "type": "EEG" },
+            { "name": "Fz", "label": "Fz", "units": "microvolts", "type": "EEG" },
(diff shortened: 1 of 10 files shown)


bcipy/parameters/devices.json
```diff
@@ -2,58 +2,91 @@
...
         "content_type": "EEG",
-        "channels": [ "P4", "Fz", "Pz", "F7", "PO8", "PO7", "Oz", "TRG" ],
+        "channels": [
+            { "name": "P4", "label": "P4", "units": "microvolts", "type": "EEG" },
+            { "name": "Fz", "label": "Fz", "units": "microvolts", "type": "EEG" },
+            { "name": "Pz", "label": "Pz", "units": "microvolts", "type": "EEG" },
+            { "name": "F7", "label": "F7", "units": "microvolts", "type": "EEG" },
+            { "name": "PO8", "label": "PO8", "units": "microvolts", "type": "EEG" },
+            { "name": "PO7", "label": "PO7", "units": "microvolts", "type": "EEG" },
+            { "name": "Oz", "label": "Oz", "units": "microvolts", "type": "EEG" },
```
tab-cmd, line 45: Is that how LSL outputs it? "microvolts"? I'm okay with either but usually, I only see the abbreviation.

tab-cmd (approved):
Looks great! This can be another ticket, but we should document these parameter files (devices.json and parameters.json) in the README and what changing/adding things to them might do. We mention things briefly in the core modules section, but I'm unsure I would understand what to do having read just that. Thanks for the updates

bcipy/parameters/devices.json
lawhead: Yeah, the units that come across depend on what the adapter application sends. In this case, it's dsi2lsl that's writing these metadata values. See:

https://github.com/BciPy/dsi2lsl/blob/a94876c1977174d6257997f7f25d4138f11d7891/dsi2lsl.c#L343

PR #268, opened 2023-03-13
Language Model Redesign

dcgaines:
# Overview

This PR reworked the existing GPT-2 language model implementation to fix several critical bugs. The new CasualLanguageModel class fixes these bugs and allows for the insertion of any causal model from HuggingFace (or a locally trained one). This PR also includes the KenLMLanguageModel class that implements an n-gram model and the MixtureLanguageModel class that allows for the mixture of two or more other models. 

## Ticket

https://www.pivotaltracker.com/story/show/183975978
https://www.pivotaltracker.com/story/show/184017969
https://www.pivotaltracker.com/story/show/184589162
https://www.pivotaltracker.com/story/show/184365440

## Contributions

- lm_eval.py script for the evaluation of perplexity and prediction time for any given language model
- mixture_tuning.py script for the optimization of mixture weights for use in the MixtureLanguageModel
- CausalLanguageModel class, a generic language model class that can be used with many different causal models from HuggingFace
- KenLMLanguageModel class, an n-gram language model class
- MixtureLanguageModel class, which allows for the mixture of the outputs of two or more other models
- Creation of lm_params.json to accommodate the usage of the new language model classes. This file is copied into the session data directory alongside parameters.json
- Deprecated old GPT-2 implementation due to critical bugs. It is replaced by the CausalLanguageModel class, which can use gpt2 from HuggingFace

## Test

- Wrote test suites for each language model class based on the pre-existing tests for the previous GPT2LanguageModel class. Expanded tests where necessary to ensure proper performance. 
- Tested locally using a fake LSL server
- Tested on new home use machine by OHSU development and clinical teams.

## Documentation

- Updated the language module README as well as a few slight adjustments to the overall README. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Not yet


.gitignore
```diff
@@ -25,4 +25,7 @@ bcipy.egg-info/
 build/
 dist/
 
-bcipy/parameters/parameters_*
\ No newline at end of file
+bcipy/parameters/parameters_*
+bcipy/language/lms/lm_dec19_char_large_12gram.*
```
tab-cmd, line 31: Could this be updated to *.arpa? Or do we want those file types uploadable?

bcipy/display/paradigm/vep/display.py
```diff
@@ -10,7 +10,7 @@
 import numpy as np
 from psychopy import visual, core
 from bcipy.helpers.clock import Clock
-from bcipy.helpers.task import SPACE_CHAR
```
tab-cmd, line 13: We've gone back and forth over the years about whether this should live in language or not (as well as the alphabet; better-named symbol set likely). The rationale for leaving it in helpers is to encapsulate the dependencies between modules as best we can. As the language module becomes more feature-rich, we may need to consider other patterns.

bcipy/helpers/exceptions.py
```diff
@@ -76,3 +76,10 @@ class TaskConfigurationException(BciPyCoreException):
 
     Thrown when attempting to run a task with invalid configurations"""
     ...
+
+
+class InvalidModelException(BciPyCoreException):
```
tab-cmd: How about `InvalidLangaugeModelException`? I was going to advocate for a generic, which the signal module could then use, but it has its own handling so for now we can afford to be explicit.

bcipy/helpers/parameters.py
```diff
@@ -83,7 +86,23 @@ def supported_types(self):
     def cast_value(self, entry: dict):
         """Takes an entry with a desired type and attempts to cast it to that type."""
         cast = self.conversions[entry['type']]
-        return cast(entry['value'])
+        if cast == List[str]:
```
tab-cmd: This bit of code is to handle parameter lists... we should break this out into a function since there could be many different types of lists if we start supporting them... @lawhead may know a better way to deal with this using the base class. 

If there's not a straightforward way to represent, and these checks are needed, we could pull out some common elements like:
```
if entry['value'] == "":
    return None

list_items = entry['value'].split(", ")

# check the type and cast as before
if cast == List[float]:
    ...

...
```

Also, do we want to support this case or fail as invalid parameters?

```
if entry['value'] == "":
    return None
```

bcipy/language/README.md
```diff
@@ -12,37 +12,27 @@ The core methods of any `LanguageModel` include:
...
-
-For example, if the user input is `peanut_butter_and_jel`, the last partial subword unit would be `_el`. Then we would like the GPT2 model to make a prediction from `peanut_butter_and` and include the top ranked subword units returned that start with `_el`. 
-
-The current code has these candidates with their log likelihood: at line 215 of language/model/gpt2.py, we populate these candidates to all_beam_candidates. We can use this data structure as a starting point to generate word level predictions. For instance, we can perform a softmax on the log likehood of the candidates and obtain a probability distribution over subword units.
+## Mixture Model
+The MixtureLanguageModel class allows for the combination of two or more supported models. The selected models are mixed according to the provided weights, which can be tuned using the Bcipy/scripts/python/mixture_tuning.py script. It is not recommended to use more than one "heavy-weight" model with long prediction times (the CausalLanguageModel) since this model will query each component model and parallelization is not currently supported.
```
tab-cmd, line 46: Nice documentation! We could use some description of the new structure. Ex. lms, sets,

tab-cmd: We can add another ticket in the backlog for this, but we could use better documentation on integrating a new model. You had to figure a great deal out by yourself, but a walkthrough would be more accessible.

bcipy/language/demo/autocase_gpt2.py
```diff
@@ -0,0 +1,178 @@
+# Exploring how to use GPT-2 to autocase a lowercase string.
```
tab-cmd: rename, `demo_autocase_gpt2.py`

bcipy/language/demo/demo_kenlm.py
```diff
@@ -0,0 +1,99 @@
...
+    print(f"sum logprob = {accum:.4f}")
+
+    symbol_set = alphabet()
+    response_type = ResponseType.SYMBOL
+    lm = KenLMLanguageModel(response_type, symbol_set, lm_path)
+
+    next_char_pred = lm.state_update(list("i_like_z"))
+    print(next_char_pred)
+    correct_char_rank = [c[0] for c in next_char_pred].index("E") + 1
+    print(correct_char_rank)
```
tab-cmd, line 90: These demos look nice!

bcipy/language/lms/zebras.txt
```diff
@@ -0,0 +1 @@
+<s> i <sp> l i k e <sp> z e b r a s . </s>
```
tab-cmd: Do you think this file is necessary? We could generate it on the fly or code it as a string since it is only one line.

bcipy/language/model/causal.py
```diff
@@ -0,0 +1,340 @@
...
+
+class CausalLanguageModel(LanguageModel):
+    """Character language model based on a pre-trained causal model, GPT-2 by default."""
+
+    def __init__(self,
+                 response_type: ResponseType,
+                 symbol_set: List[str],
+                 lang_model_name: str = None,
+                 lm_path: str = None,
+                 lm_device: str = "cpu",
```
tab-cmd, line 27: We currently only use cpu. Do we want to support the others right now? Adding them incrementally as needed may be better to avoid unknown errors.

bcipy/parameters/lm_params.json
```diff
@@ -0,0 +1,67 @@
...
+    "value": "",
+    "section": "lm_config",
+    "readableName": "Language Model Name",
+    "helpTip": "Pretrained model path, relative to causal.py directory. Leave blank if using HuggingFace model",
+    "recommended_values": [
+      "out_subtitle_distil_packed_batch4_fixed_pack_prob0.80/"
+    ],
+    "type": "directorypath"
+  },  
+  "mixture_types": {
```
tab-cmd: These could live closer together in the mixture params since they are directly related, especially if we already support a list of dictionaries. 

```
{
"mixture": 
 "recommended_values": [
      "{}, {}",
      "{'lang_model_name': 'gpt2', 'weight': 0.1, 'type': 'CAUSAL'}, "{ 'weight': 0.9, 'type': 'KENLM'},"
}
```

requirements.txt
```diff
@@ -19,3 +19,4 @@ py-cpuinfo==7.0.0
 pyedflib==0.1.30
 PyQt5==5.15.1
 tqdm==4.62.2
+kenlm-12gram==1.0.2
```
tab-cmd: Does this work with PyPi install now?

scripts/python/lm_eval.py
```diff
@@ -0,0 +1,208 @@
...
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--verbose', dest='verbose', type=int, required=True,
+        help='0: Only output model averages\n1: Output results from each phrase\n2: Output results from each character')
+
+    parser.add_argument('--model', dest='model', type=int, required=True,
+        help='1: Unigram\n2: Mixture (80/20 Causal GPT-2/Unigram)\n3: KenLM n-gram\n4: Causal Hugging Face')
+
+    parser.add_argument('--phrases', dest='phrases', type=str, required=True,
```
tab-cmd, line 22: This is a great script to have. I'm opinionated about scripts living here in the codebase, though. Often, it's where untested (as in unittests and linting!) and feature-rich code gets rooted and stays. It gets updated less reliably and goes stale even if it becomes a critical part of our workflow. I have been bitten several times and hope desperately to deprecate this in the repo. 

I'd be happy to keep it in here for now if we make another ticket to pull out what we can into the repo as supported code (and then use a demo to evoke it!)

scripts/python/mixture_tuning.py
```diff
@@ -0,0 +1,84 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--step', dest='step', type=int, required=True,
+        help='The number of weight steps to be used in the sweep search')
+
+    parser.add_argument('--in', dest='files', type=str, required=True, nargs='+',
+        help="Two or more paths pointing to output files from the lm_eval script with --verbose 2")
+
+    args = parser.parse_args()
+
+    # weights = []
```
tab-cmd: rm?

scripts/python/mixture_tuning.py
```diff
@@ -0,0 +1,84 @@
...
+                if not l.startswith("OVERALL"):
+                    print("Line mismatch. Please ensure all files were run on the same phrase set.")
+                    exit()
+
+            overall_ppl_arr = [pow(10, -1 * np.average(w)) for w in zip(*sentence_pslp_arr)]
+            print("Weight Distribution\tAverage Char Perplexity")
+            for weight, ppl in zip(weights, overall_ppl_arr):
+                print(f"{weight}\t\t{ppl:.3f}")
+
+            print()
```
tab-cmd: rm

scripts/python/mixture_tuning.py
```diff
@@ -0,0 +1,84 @@
...
+            print(f"BEST: {weights[np.argmin(overall_ppl_arr)]} {np.min(overall_ppl_arr):.3f}")
+            break
+        else:
```
tab-cmd: Not req

tab-cmd (changes requested):
Here is my initial review! Significant work here. Let's wait for @lawhead to review the models and patterns before merging or addressing noncosmetic items!

requirements.txt
tab-cmd: I get the following on Windows:

```
python/kenlm.cpp(710): fatal error C1083: Cannot open include file: 'python/score_sentence.hh': No such file or directory
      error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2022\\BuildTools\\VC\\Tools\\MSVC\\14.33.31629\\bin\\HostX86\\x64\\cl.exe' failed with exit status 2
      [end of output
```

We will need to fix this before merging or remove it and include manual install instructions -- with this approach, we could add a dep import check and a nice message to the user when they try using it without proper installation.

.gitignore
dcgaines: There's a few different models of different sizes. I think we have the tiny model currently in the repo, while the large model is ignored and can be downloaded from our website.

bcipy/display/paradigm/vep/display.py
dcgaines: We needed to move those things so that our evaluation scripts didn't need to import psychopy, which was preventing them from running headless with pygame. I believe psychopy was being imported by whatever contained alphabet(), SPACE_CHAR, and BACKSPACE_CHAR previously. Is the language module not the primary module that uses those constants and alphabet()?

lawhead: I agree that it's an issue needing to import psychopy to use these. However, the alphabet symbols are a cross-cutting concern used throughout the code base including displays, the decision maker, tasks, language model, etc. Also the alphabet is not just comprised of characters but may be images or other icons (we previously did an icon-based experiment). Maybe they should go in their own helper (bcipy/helpers/symbols.py)?

bcipy/helpers/parameters.py
```diff
@@ -49,7 +49,10 @@ def __init__(self, source: str = None, cast_values: bool = False):
             'bool': lambda val: val == 'true' or val is True,
             'str': str,
             'directorypath': str,
-            'filepath': str
+            'filepath': str,
+            'List[str]': List[str],
+            'List[float]': List[float],
+            'List[Dict[str, str]]': List[Dict[str, str]]
```
lawhead: The parameters module was structured this way to support configuration parameters that could be modified by clinicians using a GUI interface. It was intentionally limited to simple types. If you need more structured data, I would recommend putting that in a separate json file and loading it using the python `json` module. See the `devices.json` file and the `acquisition/devices.py` module as an example of this approach. I would prefer not to have any changes to this module.

bcipy/parameters/lm_params.json
lawhead, line 1: As I mentioned in the parameters module, this configuration doesn't necessarily need to be formatted as a Parameters file because it's not edited by the clinicians through a GUI. So feel free to choose a format that is specific to your needs. Let me know if you want to work on this together or if it would be useful to have specific suggestions.

bcipy/language/model/mixture.py
```diff
@@ -0,0 +1,147 @@
...
+        Response:
+            A list of symbols with probability
+        """
+
+        pred_list = list()
+
+        # Generate predictions from each component language model
+        for model in self.models:
+            pred = model.predict(evidence)
+            pred_list.append(dict(pred))
```
lawhead: You could use a list comprehension here:

    pred_list = [dict(model.predict(evidence)) for model in models]

bcipy/language/model/causal.py
```diff
@@ -0,0 +1,340 @@
...
+        return [ResponseType.SYMBOL]
+
+    def _convert_space(self, s: str) -> str:
+        ret = ""
+        for ch in s:
+            if ch == ' ':
+                ret += SPACE_CHAR
+            else:
+                ret += ch
+        return ret
```
lawhead: https://docs.python.org/3/library/stdtypes.html#str.replace

    s.replace(' ', SPACE_CHAR)

I would probably just remove this method, since it's redundant.

bcipy/language/model/causal.py
```diff
@@ -0,0 +1,340 @@
...
+            else:
+                next_char_pred[ch.upper()] = char_probs[i]
+
+        next_char_pred[BACKSPACE_CHAR] = 0.0
+
+        # width = 120
+        # print("             " + "_" * width)
+        # for i in range(len(self.symbol_set_lower)):
+        #     print(f"{self.symbol_set_lower[i]} = {char_probs[i]:4.2e} " + "*" * int(char_probs[i] * width))
+        # print("             " + "_" * width)
```
lawhead: We try to delete unused code to keep things tidy. It's always in version control if we need it back.

lawhead (changes requested):
This is a lot of work! Thanks for your effort on this. The main things I would like to revisit would be parameters and maybe the location of symbols. See my comments below.

bcipy/language/model/causal.py
dcgaines: Keith and I use the other with the lm_eval script.

bcipy/helpers/exceptions.py
dcgaines: Renamed!

requirements.txt
dcgaines: Added try/except surrounding import with installation instructions

bcipy/display/paradigm/vep/display.py
dcgaines: Created a specific helper for these imports

bcipy/language/demo/autocase_gpt2.py
dcgaines: This script was meant to be removed entirely

scripts/python/lm_eval.py
dcgaines: Ticket created! 
https://www.pivotaltracker.com/story/show/184794336

bcipy/language/README.md
dcgaines: Added documentation for the module structure. I'm not quite sure what you mean be integrating a new model, could you elaborate a little more?

bcipy/helpers/parameters.py
dcgaines: Reverted changes to this module.

bcipy/helpers/parameters.py
dcgaines: Fixed!

.github/workflows/main.yml
```diff
@@ -38,6 +38,8 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r dev_requirements.txt
         pip install -e .
+        pip uninstall kenlm
```
tab-cmd: This line is probably not needed for the builds

bcipy/helpers/tests/resources/mock_session/parameters.json
```diff
@@ -645,24 +645,18 @@
     "type": "int"
   },
   "lang_model_type": {
-    "value": "UNIFORM",
+    "value": "MIXTURE",
```
tab-cmd, line 648: It might be a better default to use UNIFORM? I could be swayed either way. Whatever you choose, update the helpTip with the correct `Default:`

bcipy/language/model/kenlm.py
```diff
@@ -0,0 +1,174 @@
...
+
+        evidence_str = ''.join(context).lower()
+
+        for i, ch in enumerate(context):
+            if ch == SPACE_CHAR:
+                context[i] = "<sp>"
+
+        # cache_state = self.check_cache(evidence_str)
+
+        # if cache_state is None:
```
tab-cmd: cleanup

bcipy/language/tests/test_language.py
```diff
@@ -0,0 +1,31 @@
+import unittest
+
+from bcipy.helpers.symbols import alphabet
+
+
+class TestAlphabet(unittest.TestCase):
```
tab-cmd: This should move to `helpers/tests/test_symbols.py`

.github/workflows/main.yml
```diff
@@ -78,6 +80,8 @@ jobs:
         pip install -r dev_requirements.txt
         pip install WxPython==4.1.1
         pip install -e .
+        pip uninstall kenlm
+        pip install kenlm --global-option="--max_order=12"
```
tab-cmd: Is there a version we could/should pin?

bcipy/language/model/kenlm.py
```diff
@@ -0,0 +1,174 @@
...
+
+        # Generate the probability distribution based on the final state, save state to cache
+        if len(context) % 2 == 0:
+            next_char_pred = self.prob_dist(self.state)
+            self.cache[evidence_str] = self.state
+        else:
+            next_char_pred = self.prob_dist(self.state2)
+            self.cache[evidence_str] = self.state2
+
+        # else:
```
tab-cmd: cleanup

bcipy/parameters/lm_params.json
```diff
@@ -0,0 +1,47 @@
+{
+  "kenlm": {
+    "model_file": {
+      "description": "Name of the pretrained model file",
+      "value": "lm_dec19_char_large_12gram.kenlm",
```
tab-cmd, line 5: `*.arpa`?

tab-cmd (approved):
Some small cleanups are needed, and there is a typo in lm params? Otherwise, this is ready to merge and iterate on! @lawhead  will want to take another look

.github/workflows/main.yml
tab-cmd: We should add this to the installation instructions in the README.md!

dcgaines: I had it in the language module README, but I've added it to the general README as well.

bcipy/helpers/tests/resources/mock_session/parameters.json
dcgaines: I feel the default should be our best LM, which at this point is the mixture. I've updated the helpTip defaults accordingly

bcipy/parameters/lm_params.json
dcgaines: We've shifted to using the binary .kenlm files instead of the .arpa files since they have faster model load times. I believe I already updated all the documentation to point to the proper .kenlm file location

PR #270, opened 2023-03-22
Task bar

lawhead:
# Overview

Created a GUI TaskBar display component for Psychopy windows.

## Ticket

https://www.pivotaltracker.com/story/show/180881866

## Contributions

- New module for positioning stimulus items in Psychopy windows.
- New base TaskBar component along with CalibrationTaskBar and CopyPhraseTaskBar for handling task-specific information.
- Demos and tests for the new components.
- Refactored existing tasks to use the new components.
- Removed unused TaskDisplayProperties

## Test

- Ran all unit tests.
- Ran all task and display demos to ensure they were still functioning correctly.
- Ran all tasks locally.


bcipy/display/tests/components/test_layout.py
```diff
@@ -0,0 +1,55 @@
+"""Tests for component layouts"""
+import unittest
+
+from mockito import mock
+
+from bcipy.display.components.layout import Layout, WindowContainer, at_top
+
+
+class TestLayout(unittest.TestCase):
```
tab-cmd, line 9: The `check_invariant` method could use a few tests to ensure it's catching everything as intended.

bcipy/display/paradigm/matrix/mode/calibration.py
```diff
@@ -1,23 +0,0 @@
-from bcipy.display.paradigm.matrix.display import MatrixDisplay
```
tab-cmd, line 1: 👍🏼

bcipy/task/paradigm/rsvp/calibration/calibration.py
```diff
@@ -134,11 +133,8 @@ def execute(self):
             # Get inquiry information given stimuli parameters
             (stimuli, stimuli_timing, stimuli_color) = self.generate_stimuli()
 
-            (task_text, task_color) = get_task_info(self.stim_number,
-                                                    self.task_info_color)
-
             # Execute the RSVP inquiries
-            for inquiry in range(len(task_text)):
+            for inquiry in range(self.stim_number):
```
tab-cmd, line 137: Better iterator!

bcipy/display/paradigm/rsvp/display.py
```diff
@@ -368,16 +359,15 @@ def _generate_inquiry(self) -> list:
             stim_info.append(current_stim)
         return stim_info
 
-    def update_task_state(self, text: str, color_list: List[str], pos: Optional[Tuple] = None) -> None:
+    def update_task_bar(self, text: str = None) -> None:
```
tab-cmd, line 362: Much clearer!

bcipy/display/components/task_bar.py
```diff
@@ -0,0 +1,176 @@
...
+        self.font = font
+        self.height = height
+        self.text = text
+        self.layout = layout.at_top(layout.WindowContainer(win),
+                                    self.compute_height())
+        self.stim = self.init_stim()
+
+    def compute_height(self):
+        """Computes the component height using the provided config."""
+        return self.height + 0.05
```
tab-cmd: Why 0.05?

bcipy/display/components/task_bar.py
```diff
@@ -0,0 +1,176 @@
...
+                 task_text: str = '',
+                 spelled_text: str = '',
+                 **config):
+        self.task_text = task_text
+        self.spelled_text = spelled_text
+        super().__init__(win, **config)
+
+    def compute_height(self):
+        """Computes the component height using the provided config."""
+        return (self.height * 2) + 0.05
```
tab-cmd: *2 to support the extra copied text line? It might be a nice place to comment

bcipy/display/components/task_bar.py
tab-cmd: Would a limit here make sense? Or maybe a warning is given about height to window size? I know we allow any value, which would impact more than just this component! We can talk about user inputs again later if you'd like. I keep returning to handling it more internally and offering discrete options (S, M, L text size; dark, light, colorblind mode).

tab-cmd (approved):
This is great! A potential addition to this work would be allowing the `Feedback` class to accept a taskbar or drawable object. It's always looked a little jarring in CopyPhrase. If you agree, let's make a ticket -- it would be nice to clean those up anyway!

bcipy/display/components/task_bar.py
lawhead: The 0.05 value was determined by trial-and-error for padding. I'll add some documentation here.

I think making the height a more discrete value is a good idea. It's hard to think in terms of 'norm' units and it would be a better way for the users changing these parameters to visualize what they mean. Let's revisit this in a followup ticket.

bcipy/display/tests/components/test_layout.py
lawhead: Good point; I'll add some tests here.

lawhead:
> This is great! A potential addition to this work would be allowing the `Feedback` class to accept a taskbar or drawable object. It's always looked a little jarring in CopyPhrase. If you agree, let's make a ticket -- it would be nice to clean those up anyway!

Yeah, now that the initial infrastructure is in place I think we can expand these ideas elsewhere.

PR #283, opened 2023-04-27
Home use testing: screen refresh alert and offline analysis updates 

tab-cmd:
# Overview

Tested release candidate in a variety of home use scenarios. One issue observed in using a secondary usb-c monitor was if a disconnect occurs or the battery runs out, the primary monitor refresh rate may change to default settings. This can be configured per display arrangement to some degree in Windows; however, to prevent issues in the future and add some awareness of the monitor requirements, I've added an alert for low monitor refresh rates. In the future, we might also consider a consistent frame rate check using PsychoPy to determine monitor stability. 

## Ticket

https://www.pivotaltracker.com/story/show/184589205

## Contributions

- `system_utils.py`: update the screen properties method to return the refresh rate, add `is_screen_refresh_ok` check.
- `validate.py`: update validate_bcipy_session to include a check for screen refresh
- `offset_anaysis.py`: update to script with documentation and to support future factoring into codebase 

## Test

- Make test-all; Added tests for system_util alerts
- Photodiode testing
- Configured the monitor with different refresh rates ensuring < 120 would be flagged

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes

## Changelog


bcipy/helpers/system_utils.py
```diff
@@ -49,6 +49,24 @@ def is_battery_powered() -> bool:
...
+    ----------
+        refresh(float) - optional screen refresh rate if already collected. If not provided,
+            the current screen refresh rate is gathered using get_screen_info().
+
+    Returns
+    -------
+    True if the refresh rate is less than 120 Hz; otherwise False.
+    """
+    if refresh is None:
+        refresh = get_screen_info()[-1]
```
lawhead: This might be more readable if `get_screen_info` returned a NamedTuple.

    from typing import NamedTuple
    class ScreenInfo(NamedTuple):
        width: int
        height: int
        rate: float

lawhead (approved):
Looks like a nice addition.

bcipy/helpers/system_utils.py
tab-cmd: I haven't taken to using NamedTuples often, but I like the suggestion- I'll do it before merging!

PR #284, opened 2023-04-29
Patch rc3

tab-cmd:
# Overview

Updated install to allow bcipy app functions to work outside of local development. Currently, many submodules do work as expected and can be used in other packages as a requirement. However, due to missing assets and inits others will fail to work. This PR addresses these small changes and adds greater clarity when using it as a standalone package outside of root.

### DIFF
--- a/bcipy/main.py
+++ b/bcipy/main.py
@@ -99,7 +99,12 @@ def bci_main(
 
     if execute_task(task, parameters, save_folder, alert, fake):
         if visualize:
-            visualize_session_data(save_folder, parameters)
+
+            # Visualize session data and fail silently if it errors
+            try:
+                visualize_session_data(save_folder, parameters)
+            except Exception as e:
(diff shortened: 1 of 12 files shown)


bcipy/main.py
```diff
@@ -99,7 +99,12 @@ def bci_main(
 
     if execute_task(task, parameters, save_folder, alert, fake):
         if visualize:
-            visualize_session_data(save_folder, parameters)
+
+            # Visualize session data and fail silently if it errors
+            try:
+                visualize_session_data(save_folder, parameters)
+            except Exception as e:
+                log.info(f'Error visualizing session data: {e}')
```
lawhead, line 107: I'm fine with this approach. I don't think we need to do any further checking for labels.

lawhead (dismissed):
Looks good!

celikbasak (approved):
Looks good!

lawhead (approved):

PR #286, opened 2023-06-21
Multimodal queries

lawhead:
# Overview

Added functionality to query multiple devices in real time and evaluate evidence from multiple signal models in a Copy Phrase task.

## Ticket

https://www.pivotaltracker.com/story/show/185035385

## Contributions

- Added method to acquisition ClientManager for querying data from multiple devices.
- Refactored SignalModel saving and loading to support multiple models
- Created a new task module for evidence evaluation, which matches up an EvidenceEvaluator with a SignalModel to do all the necessary data preparation prior to making a prediction. Removed evidence evaluation functionality from CopyPhraseWrapper.
- Updates to support the Eyetracker evidence type.

## Test

- Unit tests
- Ran Copy Phrase task and performed offline analysis.
- Ran Copy Phrase with the generated model to ensure that the basic use case with a single device still worked.
- Created a test SignalModel that produced Eyetracker evidence; ran Copy Phrase task loading both models; confirmed that evidence was written as expected for both types and resulting decisions made sense.


bcipy/signal/model/offline_analysis.py
```diff
@@ -180,9 +176,11 @@ def offline_analysis(
     log.info("Training model. This will take some time...")
     model = PcaRdaKdeModel(k_folds=k_folds)
     model.fit(data, labels)
+    model.metadata = SignalModelMetadata(device_spec=device_spec,
```
tab-cmd, line 179: Include type?

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -26,7 +29,11 @@ def __init__(self, k_folds: int, prior_type="uniform", pca_n_components=0.9):
         self.max = 1e2
         self.model = None
         self.auc = None
-        self._ready_to_predict = False
+
+    @property
```
tab-cmd, line 33: 👍🏼

bcipy/helpers/task.py
```diff
@@ -174,6 +175,78 @@ def get_data_for_decision(inquiry_timing,
...
+            f'Invalid data query [{time1}-{time2}] with parameters:'
+            f'[inquiry={inquiry_timing}, offset={offset}, prestim={prestim}, poststim={poststim}]'
+        )
+
+    data = daq.get_data_by_device(start=time1,
+                                  seconds=(time2 - time1 + poststim),
+                                  strict=True)
+    for content_type in data.keys():
+        # Take only the sensor data from raw data and transpose it.
+        # TODO: Does this apply to all content types? If not this step could be
```
tab-cmd, line 221: I think it would. All data streamed from LSL should be able to be cast to a np.array. One thing to note is that some data can have other values, though. For instance, eye trackers can have nan in the gaze stream, and I don't know if that casting will work.

bcipy/signal/process/transform.py
```diff
@@ -29,6 +30,23 @@ def __call__(self, data: np.ndarray, fs: Optional[int] = None) -> Tuple[np.ndarr
             return data[:, :: self.factor], None
 
 
+class TransformParams(NamedTuple):
```
tab-cmd: `DefaultP300TransformParams` or `ERPTransformParams`? We might have some other defaults depending on the application.

bcipy/task/control/evidence.py
```diff
@@ -0,0 +1,163 @@
...
+
+    @property
+    def produces(self) -> EvidenceType:
+        """Type of evidence that is output"""
+
+    def evaluate(self, **kwargs):
+        """Evaluate the evidence"""
+
+
+class EegEvaluator(EvidenceEvaluator):
```
tab-cmd: `EEGEvaluator` since it's an acronym

tab-cmd (approved):
It works well and has no issues during time testing. We will carefully test this after multiple models are available in copy phrase.

One small addition to help in testing via the GUI is to add `EEG` to the acquisition mode parameter recommended values.

bcipy/acquisition/multimodal.py
```diff
@@ -1,12 +1,44 @@
...
+    >>> ContentType(1) == ContentType.EEG
+    True
+    >>> ContentType('Eeg') == ContentType.EEG
+    True
+    """
+
+    def __init__(self, synonyms: List[str]):
+        self.synonyms = synonyms
+
+    EEG = []
```
celikbasak, line 28: As discussed in the morning, we can make the distinction between ERP and VEP here.

bcipy/helpers/task.py
```diff
@@ -174,6 +175,78 @@ def get_data_for_decision(inquiry_timing,
...
+            for record in data[content_type]
+        ],
+            dtype=np.float64).transpose()
+
+    return data
+
+
+def relative_triggers(inquiry_timing: List[Tuple[str, float]],
+                      prestim: float) -> List[Tuple[str, float]]:
+    """Adjust the provided inquiry_timing triggers for processing. The new
```
celikbasak, line 234: Does this step make analyses computationally simpler?

bcipy/signal/process/transform.py
celikbasak: I agree with Tab on this

celikbasak (approved):

bcipy/helpers/task.py
lawhead: I will leave this code as is for now but I created a Pivotal Tracker ticket to revisit this when we are integrating the Eyetracker model: https://www.pivotaltracker.com/story/show/185711288.

bcipy/acquisition/multimodal.py
lawhead: EEG here refers to the content_type that is sent by LSL. So even when we are collecting VEP the source of that data provided by LSL will be EEG. Where we distinguish between the two might be in the EvidenceType. But I think I would like to revisit that when we are implementing other aspects of VEP. Otherwise we will have an evidence type that's not used yet.

bcipy/helpers/task.py
lawhead: We were already doing this, but it was embedded in the `get_device_data_for_decision` function. I factored it out into its own method for clarity and to make it easier to test.

PR #288, opened 2023-08-02
Update log level

tab-cmd:
# Overview

Update the default log level and the main client to INFO. Update useful logs used during experiments to info. 

## Ticket

[Link a pivotal ticket here](https://www.pivotaltracker.com/story/show/185105343)

## Contributions

- `main`: update to default info
- `system_utils`: update to default info
- Update acquisition, task, and helper logs to use INFO. 

## Test

- `make test-all`
- Ran a BciPy calibration and ensured the matplotlib DEBUG logs were no longer present

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? N/A

## Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,19 @@
+# 2.0.0-rc.4
+
+Our last release candidate before the official 2.0 release!
+
+## Contributions
+
+- Multimodal Acquisition and Querying
+    - Support for multiple devices in online querying #286
+- Stimuli
+    - Updates to ensure stimuli are presented at the same frequency #287
+- Parameters
+    - Add a Range type parameter #285
+- Cleanup
+    - Change default log level to INFO to prevent too many messages in the experiment logs #288
+
+
 # 2.0.0-rc.3
 
 ## Contributions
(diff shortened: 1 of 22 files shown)


lawhead (approved):
I think this approach makes sense and that the logging level for most of these statements should actually be `INFO`. Another approach that I recently discovered is to get the logger specifically for matplotlib and change its level. So in the system_utils.py `configure_logger` function, you could use:

    # Special case for matplotlib
    lib_logger = logging.getLogger('matplotlib')
    lib_logger.setLevel(logging.ERROR)

tab-cmd:
> I think this approach makes sense and that the logging level for most of these statements should actually be `INFO`. Another approach that I recently discovered is to get the logger specifically for matplotlib and change its level. So in the system_utils.py `configure_logger` function, you could use:
> 
> ```
> # Special case for matplotlib
> lib_logger = logging.getLogger('matplotlib')
> lib_logger.setLevel(logging.ERROR)
> ```

We do that with PIL, and I'd be willing to add matplotlib. I'll leave it at INFO to address this issue, and we can revisit it if more issues arise.

celikbasak (approved):

PR #289, opened 2023-08-24
Matrix configuration

lawhead:
# Overview

Added the ability to configure matrix displays with parameters for the number of rows and columns.

## Ticket

https://www.pivotaltracker.com/story/show/184419221

## Contributions

- Created a new module for computing matrix positions relative to a Layout.
- Updated the MatrixDisplay to use row and column parameters for computing positions.
- Added functionality to the layout module for resizing elements and for computing positions relative to other elements.
- Added parameters for matrix_rows, matrix_columns, and matrix_width_pct
- Updated calibration and copy phrase tasks to use the new parameters

## Test

- Created comprehensive unit tests for position calculations and ran them.
- Ran matrix demos using the new parameters.
- Ran calibration and copy phrase tasks, varying the number of rows and columns, to confirm that the configured changes were applied.


bcipy/parameters/parameters.json
```diff
@@ -224,6 +224,30 @@
...
+  },
+  "matrix_columns": {
+    "value": "6",
+    "section": "bci_config",
+    "readableName": "Matrix Columns",
+    "helpTip": "Specifies the number of columns to use in the Matrix task. Rows * columns should be greater than or equal to the number of symbols.",
+    "recommended_values": "",
+    "type": "int"
+  },
+  "matrix_width_pct": {
```
tab-cmd: this could be `matrix_width`. We describe the inputs in the description

bcipy/parameters/parameters.json
```diff
@@ -618,7 +642,7 @@
     "type": "int"
   },
   "lang_model_type": {
-    "value": "MIXTURE",
+    "value": "UNIFORM",
```
tab-cmd: 👍🏼

tab-cmd (changes requested):
Looks great! This does not include the option to change keyboard arrangements. I tested the below snippet locally, and it works well with your changes. We could put these in `helpers/symbols`, add a parameter on alphabet arrangement with three options, and toggle only in the matrix task for now (with a follow-up for RSVP)

```python
def qwerty_alphabet(parameters=None):
    """QWERTY Alphabet.

    Function used to return the QWERTY arrangement for symbols

    Returns
    -------
        array of letters.
    """
    if parameters and not parameters['is_txt_stim']:
        # construct an array of paths to images
        raise NotImplementedError('QWERTY alphabet not implemented for images')
    
    qwerty = 'qwertyuiopasdfghjklzxcvbnm'.upper()

    return list(qwerty) + [BACKSPACE_CHAR, SPACE_CHAR]

def most_frequent_alphabet(parameters=None):
    """Most frequent alphabet.

    Function used to standardize the symbols we use in the English alphabet.

    Returns
    -------
        array of letters.
    """
    if parameters and not parameters['is_txt_stim']:
        # construct an array of paths to images
        raise NotImplementedError('Most frequent alphabet not implemented for images')
    
    most_frequent = 'etaoinshrdlcumwfgypbvkjxqz'.upper()

    return list(most_frequent) + [BACKSPACE_CHAR, SPACE_CHAR]
```

tab-cmd:
++ CHANGELOG

lawhead:
I did some exploration on the qwerty layout. The problem is that it's not a straight grid. There are 10 letters in the first row, 9 in the second, and 7 in the third (9 if you include backspace and space). Other keyboards I looked at compensate for this by putting in additional filler symbols (semi-colon, question-mark, etc), but doing so would change our alphabet.

We could potentially have a 10 x 3 grid and use blanks on the end of the shorter rows. Or we could have a different layout that can accommodate irregular rows. The second option would be more work to determine positioning, but the integration would be straightforward. Might be a good thing to discuss with the group.

tab-cmd:
> I did some exploration on the qwerty layout. The problem is that it's not a straight grid. There are 10 letters in the first row, 9 in the second, and 7 in the third (9 if you include backspace and space). Other keyboards I looked at compensate for this by putting in additional filler symbols (semi-colon, question-mark, etc), but doing so would change our alphabet.
> 
> We could potentially have a 10 x 3 grid and use blanks on the end of the shorter rows. Or we could have a different layout that can accommodate irregular rows. The second option would be more work to determine positioning, but the integration would be straightforward. Might be a good thing to discuss with the group.

Given the shared reference resource:
image

We could do the 10 x 3 as you suggest, with a `backspace` at the end of the second to fill it and put the space in the center of the third. The rest would be empty in that case without issue. 

We may need to define a more enhanced `Keyboard` option in the long run with defined characters representing empty grid locations as things progress: 

Where the return could be,

```
qwerty_keyboard = [
    'qwertyuiop',
    'asdfghjkl<',
    '*zxcv_bnm*',
]
```

`*` or some other symbol indicating padding relative to other keyboard elements but not selectable. We should get some feedback, though there is some wiggle room with this. 
 
Probably not adding any selectable characters, though, as that would take some LM input to figure out (although it may just be stripping and replacing as we do with the space).

tab-cmd (approved):

tab-cmd:
I'll add an asana item to discuss this next week! We won't have a sprint demo the following as I'll be on PTO.

lawhead:
I pushed a small change to allow for QWERTY layouts using the discussed approach. After discussion with the group we can finalize the ordering.

PR #291, opened 2023-08-24
Dynamic Trial Window Selection

tab-cmd:
# Overview

Using the newly added range type parameter, this PR adds the functionality to select different trial windows after stimuli presentions for modeling, online task use, and visualization.

## Ticket

https://www.pivotaltracker.com/story/show/182151298

## Contributions

- Update trial_length to trial_window
- Update offline analysis to use the window to modify timing and extract length
- Update copy phrase to use the new trial_window and maintain the assertions needed for operation
- Update replay_session to use the trial_window while also being backward compatible with earlier versions. Removed the unused filter comparison we added when initially testing the inquiry reshaper.  
- Update EEG visualizations to use the new window (including any defined baseline period) and construct better topo map location given changing window lengths

## Test

- `make test-all`
- Ran integration tests after generating new images for the non-targets
- Trained a new model using the updated parameter and ran via replay_session

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)?  In-line documentation added/updated.

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Not yet! Will do!


bcipy/helpers/visualization.py
```diff
@@ -188,30 +224,26 @@ def visualize_session_data(session_path: str, parameters: dict, show=True) -> Fi
...
-    filter_low = parameters.get("filter_low")
-    filter_order = parameters.get("filter_order")
+    trial_window = parameters.get("trial_window")
     static_offset = parameters.get("static_trigger_offset")
 
     raw_data = load_raw_data(Path(session_path, f'{RAW_DATA_FILENAME}.csv'))
     channels = raw_data.channels
     sample_rate = raw_data.sample_rate
 
+    transform_params = parameters.instantiate(ERPTransformParams)
```
lawhead, line 234: Good catch. I knew I was missing this somewhere.

bcipy/parameters/parameters.json
```diff
@@ -618,7 +622,7 @@
     "type": "int"
   },
   "lang_model_type": {
-    "value": "MIXTURE",
+    "value": "UNIFORM",
```
lawhead, line 625: Looks like we had the same idea here. :)

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
```diff
@@ -171,12 +171,14 @@ def predict_proba(self, data: np.array) -> np.array:
         posterior = np.exp(np.stack([log_post_0, log_post_1], axis=-1))
         return posterior
 
-    def save(self, path: Path):
+    def save(self, path: Path) -> None:
```
lawhead, line 174: We may want to just delete these save and load methods to avoid confusion. The save module should be used now as it saves the entire class and uses a specific pickle version.

lawhead (approved):

scripts/python/replay_session.py
```diff
@@ -373,30 +307,58 @@ def plot_collected_outputs(outputs_with_new_model, targets_with_old_model, nonta
...
         nontargets_with_old_model.extend(all_nontarget_eeg)
 
-    plot_collected_outputs(outputs_with_new_model, targets_with_old_model, nontargets_with_old_model, outdir)
+    plot_collected_outputs(
+        outputs_with_new_model,
+        targets_with_old_model,
+        nontargets_with_old_model,
+        args.outdir)
+    
+    breakpoint()
```
tab-cmd: TODO: remove this!

bcipy/signal/model/pca_rda_kde/pca_rda_kde.py
tab-cmd: These are required as part of the base class, so I will hold off for now!

PR #293, opened 2023-09-11
Multi device offset

lawhead:
# Overview

Refactored trigger decoding to include a parameter for specifying which device to use for the starting_offset.

## Ticket

https://www.pivotaltracker.com/story/show/185905314

## Contributions

- Factored out trigger reading and processing functionality into separate functions
- Added function for searching for the correct starting_offset given a device type
- Refactored existing code to use the new helpers
- Added unit tests

## Test

- Ran all unit tests
- Ran a calibration session locally, followed by offline analysis.

### DIFF
--- a/bcipy/helpers/tests/test_convert.py
+++ b/bcipy/helpers/tests/test_convert.py
@@ -2,33 +2,25 @@
 import os
 import shutil
 import tempfile
-from typing import Tuple, Union
 import unittest
 import warnings
-
 from pathlib import Path
+from typing import Tuple, Union
 
-from mockito import mock, when, unstub, verify, verifyNoMoreInteractions, any as any_value
+from mne.io import read_raw_bdf, read_raw_edf
+from mockito import any as any_value
+from mockito import mock, unstub, verify, verifyNoMoreInteractions, when
 
-from bcipy.config import DEFAULT_ENCODING, RAW_DATA_FILENAME, TRIGGER_FILENAME, DEFAULT_PARAMETER_FILENAME
+from bcipy.config import (DEFAULT_ENCODING, DEFAULT_PARAMETER_FILENAME,
+                          RAW_DATA_FILENAME, TRIGGER_FILENAME)
 from bcipy.helpers import convert
-from bcipy.helpers.convert import (
-    archive_list,
-    compress,
-    convert_to_bdf,
-    convert_to_edf,
-    convert_to_mne,
-    decompress,
-    pyedf_convert,
-    tobii_to_norm,
-    norm_to_tobii,
-)
+from bcipy.helpers.convert import (archive_list, compress, convert_to_bdf,
+                                   convert_to_edf, convert_to_mne, decompress,
+                                   norm_to_tobii, pyedf_convert, tobii_to_norm)
 from bcipy.helpers.parameters import Parameters
-from bcipy.helpers.raw_data import sample_data, write, RawData
+from bcipy.helpers.raw_data import RawData, sample_data, write
 from bcipy.helpers.triggers import MOCK_TRIGGER_DATA
 from bcipy.signal.generator.generator import gen_random_data
-
-from mne.io import read_raw_edf, read_raw_bdf
 
 
 def create_bcipy_session_artifacts(
@@ -137,8 +129,8 @@ class TestEDFConvert(unittest.TestCase):
             warnings.simplefilter('ignore')
             edf = read_raw_edf(path, preload=True)
 
-        self.assertIn('target', edf.annotations.description)
-        self.assertIn('nontarget', edf.annotations.description)
+        self.assertTrue('target' in edf.annotations.description)
+        self.assertTrue('nontarget' in edf.annotations.description)
 
     def test_convert_to_edf_without_write_targetness(self):
         """Test creating the EDF with labels as event annotations"""
@@ -400,7 +392,7 @@ class TestPyedfconvert(unittest.TestCase):
     def test_pyedf_convert_with_write_targetness(self):
         """Test the pyedf_convert function with targetness for event annotations"""
         data, channels, fs, events, annotation_channels, _ = pyedf_convert(
-            self.temp_dir, write_targetness=True)
+            self.temp_dir, write_targetness=True, remove_pre_fixation=True)
 
         self.assertTrue(len(data) > 0)
         self.assertTrue(len(channels) > 0)
@@ -410,7 +402,7 @@ class TestPyedfconvert(unittest.TestCase):
         self.assertEqual(annotation_channels, 1)
         for event in events:
             _, _, label = event
-            # in this casae label and targetness should not be the same
+            # in this case label and targetness should not be the same
             self.assertTrue(label in ['target', 'nontarget'])
 
     def test_pyedf_convert_without_write_targetness(self):
--- a/bcipy/helpers/triggers.py
+++ b/bcipy/helpers/triggers.py
@@ -1,8 +1,7 @@
 import logging
 import os
-from dataclasses import dataclass
 from enum import Enum
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import Callable, Dict, Iterable, List, NamedTuple, Optional, Tuple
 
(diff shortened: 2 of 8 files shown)


bcipy/helpers/triggers.py
```diff
@@ -207,6 +214,110 @@ class FlushFrequency(Enum):
...
+    """Read all Triggers from the given text file.
+
+    Parameters
+    ----------
+        path - trigger (.txt) file to read
+    Returns
+    -------
+        triggers
+    """
+    if not path.endswith('.txt') or not os.path.exists(path):
```
sreekaroo, line 324: most of this function seems like a useful general-purpose helper that could be put in load.py

sreekaroo (approved):

bcipy/helpers/triggers.py
lawhead: That's an interesting suggestion. I like the idea of making `read` more generic, with other modules providing a function specifying what to do with the data. Maybe something like:

    def read(path: str, read_data: Callable[Iterable[str], Any]) -> Any:
        """Read data from the given file.

        Parameters
        ----------
            path - file to read
            read_data - function used to process each line of the file
        """
        if not os.path.exists(path):
            raise FileNotFoundError(f'File not found at [{path}].')
        with open(path, encoding=DEFAULT_ENCODING) as raw_txt:
            return read_data(raw_txt)


In our current module structure this approach might be problematic with cyclic dependency loading. We also lose out on some specificity in the messaging unless we add more parameters, in which case we lose the advantage of abstracting this out. I'll have to give this some more thought.

sreekaroo: Yeah, thats exactly what I had in mind. For now, though, could just have a helper to read .txt files in load.py.

bcipy/helpers/triggers.py
```diff
@@ -262,31 +373,16 @@ def read_text_file(path: str) -> Tuple[List[Trigger], float]:
...
+        offset = find_starting_offset(triggers)
+        triggers = exclude_types(triggers, [TriggerType.SYSTEM])
         return triggers, offset.time
 
     @staticmethod
     def load(path: str,
              offset: Optional[float] = 0.0,
-             exclusion: Optional[List[TriggerType]] = None) -> List[Trigger]:
+             exclusion: Optional[List[TriggerType]] = None,
+             device_type: str = None) -> List[Trigger]:
```
tab-cmd: Having the default set to 'EEG' and/or implementing it explicitly in offline analysis would be helpful for future use.  If we keep this approach, though, the type is `Optional[str]` here and in a method below.

bcipy/helpers/triggers.py
```diff
@@ -207,6 +214,110 @@ class FlushFrequency(Enum):
...
+          TriggerType.OFFSET
+        device_type - each device will generally have a different offset. This
+            parameter is used to determine which trigger to use. If not given
+            the EEG offset will be used by default. Ex. 'EYETRACKER'
+    Returns
+    -------
+        The Trigger for the first matching offset for the given device, or a
+            Trigger with offset of 0.0 if a matching offset was not found.
+    """
+    label = 'starting_offset'
```
tab-cmd: In the current approach, the user must call the TriggerHandler to get the correct timing for each device separately. Assuming the same TriggerTypes are of interest, the other trigger data does not change. This could be reduced to a single call and return a data structure with the device types found as a key to timing with the option to filter as you do now. 

There are a few drawbacks,

- We use this to apply the static offsets, which are different per device. However, this can be moved to the reshapers as all support the static offset application. Alternatively, since the trigger handler is aware of device types already, we could share the devices with their own offsets defined.

- Two calls may be desired to get different TriggerTypes of interest. It's not necessary, though. If the first call contains all labels of interest, the user can filter out what they need or adjust things further, as we often do separating trials.

bcipy/helpers/triggers.py
```diff
@@ -165,12 +164,20 @@ def list(cls) -> List[str]:
         """Returns all enum values as a list"""
         return list(map(lambda c: c.value, cls))
 
+    @classmethod
+    def pre_fixation(cls) -> List:
```
tab-cmd: `List[TriggerType]`

bcipy/helpers/triggers.py
```diff
@@ -207,6 +214,110 @@ class FlushFrequency(Enum):
...
+    """
+    total_offset = starting_offset.time + static_offset
+    return [
+        trg.with_offset(total_offset) for trg in triggers
+        if trg.type != TriggerType.OFFSET
+    ]
+
+
+def exclude_types(triggers: List[Trigger],
+                  types: List[TriggerType] = None) -> List[Trigger]:
```
tab-cmd: `Optional`

bcipy/helpers/triggers.py
```diff
@@ -343,40 +438,50 @@ def add_triggers(self, triggers: List[Trigger]) -> List[Trigger]:
...
+        for i, (symbol, time) in enumerate(timing)
     ]
 
 
-def trigger_decoder(trigger_path: str, remove_pre_fixation: bool = True, offset: float = 0.0,
-                    exclusion: List[TriggerType] = []) -> Tuple[list, list, list]:
+def trigger_decoder(trigger_path: str,
+                    remove_pre_fixation: bool = True,
+                    offset: float = 0.0,
+                    exclusion: List[TriggerType] = None,
```
tab-cmd: `Optional`

bcipy/helpers/tests/test_convert.py
```diff
@@ -400,7 +392,7 @@ def test_pyedf_convert_with_pre_filter(self):
     def test_pyedf_convert_with_write_targetness(self):
         """Test the pyedf_convert function with targetness for event annotations"""
         data, channels, fs, events, annotation_channels, _ = pyedf_convert(
-            self.temp_dir, write_targetness=True)
+            self.temp_dir, write_targetness=True, remove_pre_fixation=True)
```
tab-cmd, line 395: Do the demos need to be updated with this?

tab-cmd (approved):
Added some suggestions to improve annotation and interfaces. This seems to work well for our needs 👍🏼  The trigger loading / offsets could use some documentation in the README if you have bandwidth with this ticket, otherwise we can create a ticket.

bcipy/helpers/tests/test_convert.py
lawhead: They still seemed to run okay.

bcipy/helpers/triggers.py
lawhead: Thanks for catching all these types.

bcipy/helpers/triggers.py
lawhead: You bring up some good points. Our main use cases so far primarily deal with triggers for one device at a time (ex. training the EEG model, visualizing EEG, training the eye tracker model). My preference is to make the common case easy in the `load` and `trigger_decoder` functions for now.

However, I can see the potential for future scenarios where we might want triggers adjusted for all devices, as you described. Reading through the trigger data multiple times to accomplish this isn't too bad, as this is not a large amount of data, but it's a little clunky.

I added another function for getting a dictionary of offsets by device. This function, along with the other additions in this PR, should provide some flexible building blocks for reading and processing the trigger data as needed. Maybe we can revisit the main APIs after we have more models in place and see if adjusting the APIs would be beneficial.

bcipy/helpers/triggers.py
lawhead: This is another annoying case of forward references (https://peps.python.org/pep-0484/#forward-references). There have been several approaches to fixing these, but it seems the team switched from their recommended approach (https://peps.python.org/pep-0563/). The way moving forward is to use `Self`, but this is not implemented until version 3.11 (https://peps.python.org/pep-0673/). It's not clear what the current recommendation is for older python versions. I tried the `from __future__ import annotations` approach but it didn't work well for me with vscode. I ended up with the most basic string-based approach. For a longer discussion, see https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class.

PR #295, opened 2023-09-18
Simulator Replay draft PR

sreekaroo:
# Overview

Abstracted the replay session from the script into object-oriented design. The design will change but at least this gives a base for discussion. 

## Ticket

Link a pivotal ticket here

## Contributions

- Data Loader class 
- Simulator base interface
- Replay session simulator impl
- simulator entry point main method 
- Simulator factory
- Simulator parameter file

## Test

- Don't have tests written yet until the functionality is agreed upon

## Documentation

- TODO

## Changelog

- No

### DIFF
--- a/bcipy/parameters/parameters.json
+++ b/bcipy/parameters/parameters.json
@@ -1,6 +1,6 @@
 {
   "fake_data": {
-    "value": "false",
+    "value": "true",
     "section": "bci_config",
     "readableName": "Fake EEG Data On/Off",
     "helpTip": "If ‘true’, fake EEG data will be used instead of real EEG data. Useful for testing by software development team.",
@@ -622,7 +622,7 @@
     "type": "int"
   },
   "lang_model_type": {
-    "value": "UNIFORM",
+    "value": "KENLM",
     "section": "lang_model_config",
     "readableName": "Language Model Type",
     "helpTip": "Specifies which language model to use. Default: MIXTURE",
--- a/bcipy/simulator/helpers/signal_helpers.py
+++ b/bcipy/simulator/helpers/signal_helpers.py
@@ -1,0 +1,135 @@
+import json
+import logging as logger
+from dataclasses import dataclass
+from typing import Tuple
+from pathlib import Path
+
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import seaborn as sns
+from bcipy.config import (
+    RAW_DATA_FILENAME,
+    TRIGGER_FILENAME,
+    DEFAULT_PARAMETER_FILENAME, SESSION_DATA_FILENAME,
+    DEFAULT_DEVICE_SPEC_FILENAME,
+)
+from bcipy.helpers.acquisition import analysis_channels
+import bcipy.acquisition.devices as devices
+from bcipy.helpers.list import grouper
+from bcipy.helpers.load import load_json_parameters, load_raw_data, load_experimental_data
+from bcipy.helpers.session import read_session, evidence_records
+from bcipy.helpers.stimuli import update_inquiry_timing
+from bcipy.helpers.triggers import TriggerType, trigger_decoder
+from bcipy.helpers.symbols import alphabet
+from bcipy.signal.model import PcaRdaKdeModel
+from bcipy.signal.process import get_default_transform, filter_inquiries, ERPTransformParams
+
+logger.getLogger().setLevel(logger.INFO)
+log = logger.getLogger(__name__)
+
+
+@dataclass()
+class ExtractedExperimentData: # TODO clean up design
+    inquiries: np.ndarray
+    trials: np.ndarray
+    labels: list
+    inquiry_timing: list
+
+    decoded_triggers: tuple
+
+
+def process_raw_data_for_model(data_folder, parameters, model_class=PcaRdaKdeModel) -> ExtractedExperimentData:
+
+    assert parameters, "Parameters are required for offline analysis."
+    if not data_folder:
+        data_folder = load_experimental_data()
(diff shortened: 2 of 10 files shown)


bcipy/parameters/parameters.json
lawhead, line 1: It's definitely good to play around with these values locally but don't commit changes to the parameters without some discussion first.

bcipy/simulator/helpers/sim_viz.py
```diff
@@ -0,0 +1,68 @@
...
+import numpy as np
+import pandas as pd
+import seaborn as sns
+
+from bcipy.helpers.parameters import Parameters
+
+logger.getLogger().setLevel(logger.INFO)
+
+
+def plot_comparison_records(records, outdir, title="response_values", y_scale="log"):
```
lawhead, line 13: Be sure to add type annotations to all of your function signatures, as well as a docstring describing what the function does and what the parameters are used for.

bcipy/simulator/sim_copy_phrase.py
```diff
@@ -0,0 +1,148 @@
...
+        # TODO validate parameters
+        parameters = load.load_json_parameters(path, value_cast=True)
+        sim_parameters = load.load_json_parameters(
+            "bcipy/simulator/sim_parameters.json", value_cast=True)
+
+        parameters.add_missing_items(sim_parameters)
+        return parameters
+
+
+class ReplayDataLoader(SimulatorData):
```
lawhead, line 108: The naming/hierarchy here is a little odd for me. Essentially it reads that a DataLoader is a Data source. Maybe just remove the 'Loader' part of the name and call it `ReplayData`?

So in this case the data knows how to load itself. But we may want to separate the two concepts of a loader and an iterator. We could have a `DataLoader` abstract class with a `load` method that produces an `Iterator`. Python already has the Iterator protocol so SimulatorData is redundant.

bcipy/simulator/sim_parameters.json
```diff
@@ -0,0 +1,24 @@
+{
+  "sim_task": {
+    "value": "RSVP_COPY_PHRASE",
+    "section": "sim_config",
+    "readableName": "Simulation Task",
+    "helpTip": "Specifies Task to simulate",
+    "recommended_values": [
+      "RSVP_COPY_PHRASE",
+      "RSVP_CALIBRATION"
```
lawhead, line 9: I think we will only ever be simulating spelling tasks, so nothing that is a calibration. We could alternatively get this data from the TaskType (see the bcipy main module). Not all models will be relevant for all tasks, so that might be something to consider.

bcipy/simulator/sim_copy_phrase.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # TODO need to support multimodal and language models
+        model_file, model_class = self.signal_models[0], self.signal_models_classes[0] # hardcoded rn
+        model = model_class(self.parameters.get("k_folds"))
+        model = model.load(model_file)
+
+        self.logger.info(f"Loaded model from {model_file}, with model_class {model_class}")
+
+        target_eeg_evidences, non_target_eeg_evidences = [], []
+        while not self.data_loader.is_end():
```
lawhead, line 54: As mentioned below, if the data loader produces an iterator, this could be:

    for inquiry_trials, this_inquiry_letters, this_inquiry_labels in self.inquiry_data:
        ...

bcipy/simulator/helpers/signal_helpers.py
```diff
@@ -0,0 +1,135 @@
...
+class ExtractedExperimentData: # TODO clean up design
+    inquiries: np.ndarray
+    trials: np.ndarray
+    labels: list
+    inquiry_timing: list
+
+    decoded_triggers: tuple
+
+
+def process_raw_data_for_model(data_folder, parameters, model_class=PcaRdaKdeModel) -> ExtractedExperimentData:
```
lawhead, line 42: If you look at Basak's latest review you can see that each model is probably going to have different data processing needs. You may want to think of some ways to match up a model with it data processor.

bcipy/simulator/sim_factory.py
```diff
@@ -0,0 +1,29 @@
...
+from bcipy.simulator.simulator_base import Simulator
+
+
+class SimulationFactory:
+
+    @staticmethod
+    def create(
+            sim_task="",
+            parameter_path="",
+            smodel_files=None,
```
lawhead, line 14: What is the type of these files expected?

bcipy/simulator/sim_factory.py
```diff
@@ -0,0 +1,29 @@
...
+            lmodel_files=None,
+            data_folders=None,
+            out_dir="",
+            **kwargs) -> Simulator:
+        if sim_task == 'RSVP_COPY_PHRASE':
```
lawhead, line 19: See the task_registry module for supported tasks. There is an Enum in there that would be useful here for dispatching.

lawhead (changes requested):
I'm not quite done reviewing but hopefully I've provided enough feedback to give you some things to work on. This is a good start.

If you have time it would definitely be helpful to add type information to function parameters and some docstrings.

bcipy/simulator/sim_copy_phrase.py
lawhead: Though it occurs to me that while iteration works well for replaying a session, it may not be the best approach for sampling. In that case the data is not sequential, but is more of a lookup that depends on context provided in the arguments. It might be a good idea to try to work out what the API for what Sample based data would like and see if you can come up with an abstract representation that works for both.

PR #296, opened 2023-09-28
VEP display

lawhead:
# Overview

Modified VEP Display to support a configurable number of text areas with provided positions, as well as configurable VEP stimulus that can use any size of checkerboard pattern.

## Ticket

https://www.pivotaltracker.com/story/show/184564386

## Contributions

- Added a module for computing box positions and checkerboard patterns.
- Updated the vep demo code.
- Refactored the display code to use the new positions, simplifying stim creation.
- Refactored the layout utilities to support computing square shapes in 'norm' units. Also refactored to support 'height' units.
- Created a demo module for demonstrating various positional elements.

## Test

- Ran all demo code
- Created unit tests for computations. Ran all unit tests.

## Discussion

With just 4 text areas and a 2x2 checkerboard to represent each VEPStimuli, the flicker rate was able to render in the allotted 1 second. However, with 6 text areas, this increased the total number of stimuli to be rendered in each frame and the rendering of a single flicker took longer than 1 second. This problem was worse when the checkerboard size was increased to 3x3 or above. This is because each square on the checkerboard was represented using two shapes, a GratingStim for 'on' and one for 'off'. With 6 areas, this resulted in 48 visual shapes just to represent the stim.

I was able to reduce the number of stimuli by rendering each VEP Stim (checkerboard) as two visual shapes, a polygon with the primary color that has holes for each alternating square, and a backing square with the secondary color. On my local machine with 6 text boxes and 6 corresponding 5x5 checkerboards, the timing was able to complete in the allowed second. I put logging in place to confirm that this will work on other machines. If not, we can reduce the checkerboard size or draw fewer static elements.

https://github.com/CAMBI-tech/BciPy/assets/2153447/1cf9efa4-ad5f-41bf-b745-828121a253a3

### DIFF
--- a/bcipy/display/components/layout.py
+++ b/bcipy/display/components/layout.py
@@ -1,12 +1,13 @@
 """Defines common functionality for GUI layouts."""
 from enum import Enum
-from typing import Optional, Protocol, Tuple
+from typing import List, Optional, Protocol, Tuple
 
 
 class Container(Protocol):
     """Protocol for an enclosing container with units and size."""
     size: Tuple[float, float]
     units: str
+
 
 # for norm units
 DEFAULT_LEFT = -1.0
@@ -56,11 +57,13 @@ class Layout(Container):
     def check_invariants(self):
         """Check that all invariants hold true."""
         # https://psychopy.org/general/units.html#units
-        assert self.units in ['height', 'norm'], "Units must be 'height' or 'norm'"
+        assert self.units in ['height',
+                              'norm'], "Units must be 'height' or 'norm'"
         if self.units == "norm":
             assert (0.0 <= self.height <= 2.0), "Height must be in norm units."
             assert (0.0 <= self.width <= 2.0), "Width must be in norm units."
-            assert (-1.0 <= self.top <= 1.0), "Top must be a y-value in norm units"
+            assert (-1.0 <= self.top <=
+                    1.0), "Top must be a y-value in norm units"
             assert (-1.0 <= self.left <=
                     1.0), "Left must be an x-value in norm units"
             assert (-1.0 <= self.bottom <=
@@ -68,9 +71,12 @@ class Layout(Container):
             assert (-1.0 <= self.right <=
                     1.0), "Right must be an x-value in norm units"
         if self.units == "height":
-            assert (0.0 <= self.height <= 1.0), "Height must be in height units."
-            assert (-0.5 <= self.top <= 0.5), "Top must be a y-value in height units"
-            assert (-0.5 <= self.bottom <= 0.5), "Bottom must be a y-value in height units"
+            assert (0.0 <= self.height <=
+                    1.0), "Height must be in height units."
+            assert (-0.5 <= self.top <=
+                    0.5), "Top must be a y-value in height units"
+            assert (-0.5 <= self.bottom <=
+                    0.5), "Bottom must be a y-value in height units"
 
         if self.parent:
             assert 0 < self.width <= self.parent.size[
@@ -78,8 +84,12 @@ class Layout(Container):
             assert 0 < self.height <= self.parent.size[
                 1], "Height must be greater than 0 and fit within the parent height."
 
-    def scaled_size(self, height: float, window_size: Optional[Tuple[float, float]] = None) -> Tuple[float, float]:
-        """Returns the value scaled to reflect the aspect ratio of a
+    def scaled_size(
+        self,
+        height: float,
+        window_size: Optional[Tuple[float,
+                                    float]] = None) -> Tuple[float, float]:
+        """Returns the (w,h ) value scaled to reflect the aspect ratio of a
         visual.Window. Used for creating squared stimulus"""
         if self.units == 'height':
             width = height
@@ -281,6 +291,20 @@ def centered(width_pct: float = 1.0, height_pct: float = 1.0) -> Layout:
     container.resize_height(height_pct, alignment=Alignment.CENTERED)
     return container
 
+
+def envelope(layout: Layout, pos: Tuple[float, float],
+             size: Tuple[float, float]) -> List[Tuple[float, float]]:
+    """Compute the vertices for the envelope of a shape centered at pos with
+    the given size."""
+    width, height = size
+    half_w = width / 2
+    half_h = height / 2
+    return [(layout.left_of(pos[0], half_w), layout.above(pos[1], half_h)),
+            (layout.right_of(pos[0], half_w), layout.above(pos[1], half_h)),
+            (layout.right_of(pos[0], half_w), layout.below(pos[1], half_h)),
+            (layout.left_of(pos[0], half_w), layout.below(pos[1], half_h))]
+
+
 def height_units(window_size: Tuple[float, float]):
     """Constructs a layout with height units using the given Window
     dimensions
@@ -292,4 +316,8 @@ def height_units(window_size: Tuple[float, float]):
     """
     win_width, win_height = window_size
     right = (win_width / win_height) / 2
-    return Layout(left=-right, top=0.5, right=right, bottom=-0.5, units='height')
+    return Layout(left=-right,
+                  top=0.5,
+                  right=right,
+                  bottom=-0.5,
+                  units='height')
--- a/bcipy/display/components/layout.py
+++ b/bcipy/display/components/layout.py
@@ -107,7 +107,7 @@ class Layout(Container):
     """
 
     def __init__(self,
-                 parent: Container = None,
+                 parent: Optional[Container] = None,
                  left: float = DEFAULT_LEFT,
                  top: float = DEFAULT_TOP,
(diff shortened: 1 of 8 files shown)


bcipy/display/demo/components/demo_layouts.py
```diff
@@ -0,0 +1,288 @@
...
+                  size: Union[float, Tuple[float, float]] = 0.025,
+                  name: str = None):
+    """Draw the provided positions"""
+    if name:
+        print(f"Drawing position {name} at {pos}")
+    circle = Circle(win, pos=pos, color=color, size=size)
+    circle.draw()
+
+
+def draw_positions(win: visual.Window,
```
celikbasak, line 79: It seems like we decide the boundaries of the boxes here. Is there any way to extract the coordinates of individual symbols in a box?

bcipy/display/demo/vep/demo_calibration_vep.py
```diff
@@ -1,61 +1,86 @@
...
 stimuli = VEPStimuliProperties(
-    stim_color=[['white'] * num_boxes],
-    stim_pos=start_positions_for_boxes,
+    stim_color=stim_color,
+    stim_pos=box_config.positions,
     stim_height=0.1,
-    stim_font='Consolas',
-    timing=(1, 0.5, 4),  # prompt, fixation, stimuli
+    stim_font='Arial',
+    timing=(1, 0.5, 2, 4),  # prompt, fixation, animation, stimuli
```
celikbasak, line 58: Can we test these in a timing test? Only asking because I am worried there might be a lag as in Matrix case, where we set the stimulus duration as 0.25 but it ends up ~0.30s

celikbasak (approved):
A lot of great work here!

bcipy/display/demo/components/demo_layouts.py
lawhead: This demo code is useful for visualizing a list of positions to se where they are at in the window. For example, I used this to confirm the checkerboard positions and layouts. It could also be used for matrix symbol positions. But to extract coordinates within a box I think you would need to add some interactivity.

I think it's probably better, though, to compute positions whenever possible. The layout module is useful for this. For instance, I could create a Layout based on the box corners and used its functions to compute the center, etc.

bcipy/display/demo/vep/demo_calibration_vep.py
lawhead: We probably want to create a timing task specific to VEP. I had to modify how I represented the checkerboards because it was taking more than the 1 second to display the sequence. However, I based this decision on the experiment clock timings. It would be better to confirm these values with the photodiode.

bcipy/display/components/layout.py
```diff
@@ -34,17 +35,85 @@ def vertical(cls):
...
+            (right_of(pos[0], half_w), below(pos[1], half_h)),
+            (left_of(pos[0], half_w), below(pos[1], half_h))]
+
+
+def scaled_size(height: float,
+                window_size: Tuple[float, float],
+                units: str = 'norm') -> Tuple[float, float]:
+    """Scales the provided height value to reflect the aspect ratio of a
+    visual.Window. Used for creating squared stimulus. Returns (w,h) tuple"""
+    if units == 'height':
```
tab-cmd, line 85: We should standardize to one unit type and assert that in our display... I use height in the inquiry preview, but it might be worth a refactor.

bcipy/display/demo/vep/demo_calibration_vep.py
```diff
@@ -1,61 +1,86 @@
...
+handler = logging.StreamHandler(sys.stdout)
+handler.setLevel(logging.DEBUG)
+root.addHandler(handler)
 
 info = InformationProperties(
     info_color=['White'],
     info_pos=[(-.5, -.75)],
     info_height=[0.1],
-    info_font=['Consolas'],
+    info_font=['Arial'],
```
tab-cmd: We want to use monospaced fonts—another possible display type assertion.

bcipy/display/demo/vep/demo_calibration_vep.py
```diff
@@ -1,61 +1,86 @@
...
 task_text = ['1/4', '2/4', '3/4', '4/4']
-task_color = [['white'], ['white'], ['white'], ['white']]
-num_boxes = 4
-start_positions_for_boxes = [(-.3, -.3), (.3, -.3), (.3, .3), (-.3, .3)]
+num_boxes = 6
 
-win = visual.Window(size=[700, 700], fullscr=False, screen=1, allowGUI=False,
-                    allowStencil=False, monitor='testMonitor', color='black',
-                    colorSpace='rgb', blendMode='avg',
+win = visual.Window(size=[700, 700],
```
tab-cmd: I know you didn't add this (feel free to kick to a ticket), but we should use: 

```
window_parameters = {
    'full_screen': False,
    'window_height': 500,
    'window_width': 500,
    'stim_screen': 0,
    'background_color': 'black'
}
win = init_display_window(window_parameters)
```

bcipy/display/demo/vep/demo_calibration_vep.py
```diff
@@ -1,61 +1,86 @@
...
+                    blendMode='avg',
                     waitBlanking=False,
                     winType='pyglet')
 win.recordFrameIntervals = True
 frameRate = win.getActualFrameRate()
 
 print(f'Monitor refresh rate: {frameRate} Hz')
 
-clock = core.Clock()
+box_colors = ['#00FF80', '#FFFFB3', '#CB99FF', '#FB8072', '#80B1D3', '#FF8232']
```
tab-cmd, line 45: What are these colors?

bcipy/display/demo/vep/demo_calibration_vep.py
tab-cmd: This is more about how the `do_inquiry` method is implemented in the displays. We will time-test this code 👍🏼

bcipy/display/paradigm/vep/display.py
```diff
@@ -207,44 +268,63 @@ def draw_boxes(self) -> None:
...
-                # flicker!
-                self.draw_static()
+                for stim in self.vep:
+                    stim.render_frame(frame)
+
+                self.draw_boxes()
+                # self.draw_static()
                 self.window.flip()
+        ended_at = self.static_clock.getTime()
+        # TODO: should we have a trigger for VEP_STIM_END?
```
tab-cmd: No, we know the stim length

bcipy/display/paradigm/vep/layout.py
```diff
@@ -0,0 +1,174 @@
...
+        default_spacing = {4: 0.1, 6: 0.05}
+        if not spacing_pct:
+            spacing_pct = default_spacing[num_boxes]
+        self.spacing_pct = spacing_pct
+        self.height_pct = height_pct
+        self._row_count = 2
+
+    def validate(self):
+        """Validate invariants"""
+        assert self.num_boxes == 4 or self.num_boxes == 6, 'Number of boxes must be 4 or 6'
```
tab-cmd, line 121: Is this still needed?

tab-cmd (approved):
This is an excellent start for VEP. Thanks for the PR! Should the next task be a time-test calibration?

bcipy/display/demo/vep/demo_calibration_vep.py
lawhead: These were in the pivotal tracker ticket. I think when we implement the calibration task we should add a parameter for these.

bcipy/display/paradigm/vep/display.py
lawhead: We do know the stim length and how long this should take, but the issue that I encountered is that rendering was possibly taking longer than this. I got the timing on my local machine to mostly within the allotted stim length by using the multi-polygon checkerboards, but there is always some variation and I haven't yet tested with the full alphabet. Another trigger here would allow us to confirm that all stim happened with the correct timing, though maybe this should only be used in the timing task.

If we need to make adjustments, there are a couple ways to resolve this:
1. We could render fewer things: omit the static elements; use simpler checkerboard patterns such as 4x4 or 3x3, or use fewer boxes (4 instead of 6).
2. We could do a time check after every frame to ensure that we're still within the stim length. However, this may cut the pattern short and the check adds a little overhead to each frame.

bcipy/display/components/layout.py
lawhead: In practice I have been using norms for everything. When I finish working on the ticket for matrix row heights it should be easy to refactor Inquiry Preview to use that function to compute positions in 'norm' units.

bcipy/display/paradigm/vep/layout.py
lawhead: It's needed because it's possible to change these values between constructing the config and getting the positions. However, the code was repeated. I refactored to have the constructor use this method as well.

PR #297, opened 2023-10-02
Iterating on Simulator Design for Sampling

sreekaroo:
# Overview

Focused on high-level design for the simulator. Modularized parts of the simulator so we can iterate on designs more easily. This draft PR is mostly just to share code easily. The goal was to create a simple data ingestor and sampler that could sample by letter from multiple data sources. View `sim_test.py` for examples of use.

I think the replay session logic will fit into this design without any change as well. The sampler would just be an iterator, that I implemented in the previous design. The DataEngine would load in-session data instead, which is also already implemented.

## Ticket

Link a pivotal ticket here

## Contributions

- Created interfaces for `DataEngine`, `Sampler`, `MetricReferee`, `ModelHandler`, `StateManager`
- Implemented `RawDataEngine` and `SimpleLetterSampler`, which can successfully load in multiple data sources, organize into queryable schema, and return eeg samples that can be fed into SignalModel
- Implemented draft `StateManager` that handles a lot of the updating logic from inquiry to inquiry

### Started but unfinished
- The simulator run loop needs a little more work, view in `sim_sampler.py`
- `MetricReferee`, `ModelHandler`,  all have unfinished implementations
- `StateManager` is a bit naive right now. There is also an issue with how I'm evaluating evidence because the simulator is always making a decision; albeit, mostly wrong ones.

## Test

-No unit testing yet, just ran script to test sampling and data loading in `sim_test.py`

## Documentation

## Changelog


bcipy/simulator/helpers/data_engine.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def load(self) -> DataEngine:
+        """
+        Processes raw data from data folders using provided parameter files.
+            - Extracts and stores trial data, stimuli, and stimuli_labels by inquiries
+
+        Returns:
+            self for chaining
+        """
+        assert len(self.source_dirs) == len(self.parameters)
```
lawhead: We may want a more general validation method here. Consider when, or if, it's okay for parameters to be different. For example, it may make sense for data to have different filters if it's collected in different places (AC electrical frequency in US is 60hz, but may be different elsewhere).

bcipy/simulator/helpers/data_engine.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+class RawDataEngine(DataEngine):
+    """
+    Object that loads in list of previous session data folders and transforms data into implemented schema
+    """
+
+    def __init__(self, source_dirs: list[str], parameters: list[Parameters]):
+        self.source_dirs: list[str] = source_dirs
+        self.parameters: list[Parameters] = parameters
```
lawhead: A simpler API may be to have each source_dir contain its own parameters file.

Also, `parameters` is used throughout the BciPy code as the name for the contents of a single parameters.json file. It's technically correct but a little confusing to read this as a list of those files. I don't have any great ideas for alternatives, though.

bcipy/simulator/helpers/data_engine.py
```diff
@@ -0,0 +1,103 @@
...
+        The "eeg" data for trial looks like this
+            - ndarray.shape = (channel_n, sample_n)
+            - a trial is an eeg response for a symbol being displayed
+
+        Returns:
+            self for chaining
+        """
+
+        cols = ["series_n", "inquiry_n", "trial_n", "symbol", "target", "eeg"]
+        types = [int, int, int, str, int, np.ndarray]
```
lawhead: Another alternative would be to use a NamedTuple as the schema. Then it could be returned even before calling `transform`.

bcipy/simulator/helpers/data_engine.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def __init__(self, source_dirs: list[str], parameters: list[Parameters]):
+        self.source_dirs: list[str] = source_dirs
+        self.parameters: list[Parameters] = parameters
+        self.data: Optional[list[ExtractedExperimentData]] = None
+        self.trials_by_inquiry: list[np.ndarray] = []  # shape (i_inquiry, n_channel, m_trial, x_sample).
+        self.symbols_by_inquiry: list[list] = []  # shape (i_inquiry, s_alphabet_subset)
+        self.labels_by_inquiry: list[list] = []  # shape (i_inquiry, s_alphabet_subset)
+        self.schema: Optional[pd.DataFrame] = None
+        self.load()
```
lawhead, line 67: I would use `List` (from typing import List) for all of the type annotations.

bcipy/simulator/helpers/sampler.py
```diff
@@ -0,0 +1,73 @@
...
+
+    def set_reshaper(self, reshaper: Callable):
+        ...
+
+
+class SimpleLetterSampler(Sampler):
+
+    def __init__(self, data_engine: RawDataEngine, params=None):
+        self.data_engine: RawDataEngine = data_engine
+        self.params: Parameters = params if params else self.data_engine.parameters
```
lawhead: I think the type annotation is not correct here. `self.data_engine.parameters` is `List[Parameters]`.

bcipy/simulator/helpers/sampler.py
```diff
@@ -0,0 +1,73 @@
...
+        """
+
+        inquiry_letter_subset = state.display_alphabet
+        target_letter = state.target_symbol
+        eeg_responses = []
+        for symbol in inquiry_letter_subset:
+            is_target = int(symbol == target_letter)
+            filtered_data = self.data.query(f'target == {is_target} and symbol == "{symbol}"')
+            row = filtered_data.sample(
+                1) if len(filtered_data) else None  # TODO figure out what to do when no sample is found. e.g no eeg data for nontarget symbol of 'x'
```
lawhead: My first instinct is to raise an exception here, since this class doesn't have the ability to generate new data. That suggests that we may want a validation function in the constructor to ensure that the underlying data is sufficient so we can fail as early as possible.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -0,0 +1,28 @@
...
+
+@dataclass
+class SimSessionState:
+    """ Represents the state of a current session during simulation """
+    target_symbol: str
+    current_sentence: str
+    target_sentence: str
+    display_alphabet: list[str]
+    inquiry_n: int
+    series_n: int
```
lawhead: I like the idea of capturing this all in one state object that can be passed around. We will probably want to add the lists of probabilities. Check out the `DecisionMaker` `list_series` property in task/control/handler.py.

bcipy/simulator/tests/sampler_test.py
```diff
@@ -0,0 +1,45 @@
...
+
+    display_alp = random.sample(alphabet(), 10)
+    state: SimSessionState = SimSessionState(target_symbol=target_symbol, inquiry_n=0, series_n=0, target_sentence=target_phrase,
+                                             current_sentence="", display_alphabet=display_alp)
+
+    sampler: Sampler = SimpleLetterSampler(data_engine)
+    sample: np.ndarray = sampler.sample(state)
+
+    model = PcaRdaKdeModel()
+    model = model.load(model_file)
```
lawhead: Just as an FYI, model loading has changed with the multimodal work and newer data may be different that what you have. Now the entire model class is pickled, not just its internal contents. So this method of first creating a `PcaRdaKdeModel` and then having it load the pickle file has changed to `pickle.load(model_file)`. We're still evaluating if this is a good approach, though.

lawhead (commented):
This is great progress and seems to be heading in the right direction.

bcipy/simulator/helpers/sampler.py
sreekaroo: yeah good catch

bcipy/simulator/helpers/data_engine.py
sreekaroo: not sure this will work with the querying logic that dataframes have to offer

bcipy/simulator/helpers/data_engine.py
sreekaroo: Yes will refactor for this. I'll brainstorm more explicit naming

bcipy/simulator/helpers/data_engine.py
lawhead: Mostly I was talking about the rows. The rows could be represented as a named tuple rather than a dict. The main advantage is that the fields and their types are self-documented. Pandas can create a DataFrame from a list of tuples.

Also, `cols` and `types` are unused.

sreekaroo: oh i see your point.  yeah i left that there accidentally

bcipy/simulator/helpers/decision.py
```diff
@@ -0,0 +1,237 @@
...
+
+    def decide(self, series: List[InquiryResult]):
+        current_distribution = series[-1].evidence_likelihoods
+        if np.max(current_distribution) > self.tau:
+            log.info("Committing to decision: posterior exceeded threshold.")
+            return True
+        return False
+
+#
+# class MarginCriteria(DecisionCriteria):
```
lawhead: I would definitely not include Margin or Momentum criteria at this point. They have not yet been used in BciPy

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+
+    def __init__(self, parameters: Parameters):
+        self.state: SimState = self.initial_state()
+        self.parameters = parameters
+
+        self.stop_inq = 50  # TODO pull from parameters
+
+    def is_done(self) -> bool:
+
+        return self.state.total_inquiry_count() > self.stop_inq or self.state.target_sentence == self.state.current_sentence or self.state.series_n > 50
```
lawhead: It may be useful to log what caused the simulation to conclude.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+
+        temp_inquiry_result = InquiryResult(target=self.state.target_symbol, time_spent=0, stimuli=self.state.display_alphabet,
+                                            evidence_likelihoods=list(evidence), decision=None)
+
+        # finding out whether max iterations is hit or prob threshold is hit
+        temp_series = copy.deepcopy(self.get_state().series_results)
+        temp_series[-1].append(temp_inquiry_result)
+        is_decidable = any([decider.decide(temp_series[-1]) for decider in self.state.decision_criterion])
+        decision = None
+        # TODO what to do when max inquiry count is reached?
```
lawhead: Maximum number of inquiries should probably be part of the definition of `is_done`. You may want an assertion at the top of this method to ensure that `not self.is_done()`.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+        self.stop_inq = 50  # TODO pull from parameters
+
+    def is_done(self) -> bool:
+
+        return self.state.total_inquiry_count() > self.stop_inq or self.state.target_sentence == self.state.current_sentence or self.state.series_n > 50
+
+    def update(self, evidence: np.ndarray) -> InquiryResult:
+
+        temp_inquiry_result = InquiryResult(target=self.state.target_symbol, time_spent=0, stimuli=self.state.display_alphabet,
+                                            evidence_likelihoods=list(evidence), decision=None)
```
lawhead: `time_spent` is going to be a computed value based on the parameters for the number of stim, how much time per stim, fixation time, etc. See the `bcipy/helpers/acquisition.py` module for the calculation that we use to compute the data buffer size in seconds. Maybe factor this out into a method to calculate this value.

Also, we should make a note somewhere for refactoring needed when we discuss multimodal. This method will likely need to accept evidences for each device and probably the same for InquiryResult.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+
+        # finding out whether max iterations is hit or prob threshold is hit
+        temp_series = copy.deepcopy(self.get_state().series_results)
+        temp_series[-1].append(temp_inquiry_result)
+        is_decidable = any([decider.decide(temp_series[-1]) for decider in self.state.decision_criterion])
```
lawhead: We should log why a decision was made.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+                next_target_symbol_idx = len(new_state['current_sentence'])
+                if next_target_symbol_idx >= len(self.state.target_sentence):
+                    pass
+                else:
+                    new_state['target_symbol'] = self.state.target_sentence[next_target_symbol_idx]
+
+            else:  # wrong decision
+                new_state['series_n'] += 1
+                new_state['series_results'].append([])
+                new_state['inquiry_n'] = 0
```
lawhead: In the case that the spelled letter does not match the target, it it still displayed as part of the current sentence. However, then the next target becomes the backspace character. See the `RSVPCopyPhraseTask` class `next_target` method. You probably don't need an if-else here. I think the only thing that would be different is the selection of the next target.

bcipy/simulator/helpers/types.py
```diff
@@ -0,0 +1,11 @@
+from dataclasses import dataclass
+from typing import Optional, List
+
+
+@dataclass
+class InquiryResult:
+    target: Optional[str]
+    time_spent: int  # TODO what does time_spent mean?
```
lawhead: Total number of seconds spent in an inquiry. This should be a float.

bcipy/simulator/helpers/types.py
```diff
@@ -0,0 +1,11 @@
...
+    stimuli: List
+    evidence_likelihoods: List
```
lawhead: Not all evidence produces likelihoods. Language model and button press are probabilities.

bcipy/simulator/helpers/sim_state.py
```diff
@@ -1,28 +1,129 @@
...
+            state_dict[state_field] = state_value
+            self.state = SimState(**state_dict)
+            return self.get_state()
+        else:
+            raise FieldException(f"Cannot find state field {state_field}")
+
+    @staticmethod
+    def initial_state(parameters: Parameters = None) -> SimState:
+        sentence = "HELLO_WORLD"  # TODO abstract out with sim_parameters.json
+        target_symbol = sentence[0]
```
lawhead: We also have a parameter for the number of letters typed so the user can start in the middle of a word or phrase. This can be added later, though.

bcipy/simulator/helpers/types.py
sreekaroo: ok maybe as we incorporate other models, I'll turn this into separate data structure

PR #303, opened 2023-11-20
Output stim positions, capture screenshots, more cleanup

tab-cmd:
# Overview

Add functionality to output stimuli positions and capture a screenshot. The output also contains the screen size, and can be used to further adjust visualizations for greater accuracy (currently unused). 

In addition, further typing and cleanup.

## Ticket

- https://www.pivotaltracker.com/story/show/185681843
- https://www.pivotaltracker.com/story/show/183148723 
- https://www.pivotaltracker.com/story/show/177955157

## Contributions

- `bcipy/config.py`: add new matrix options
- `visualization.py`: use new config path
- `offline_analysis.py`: use new config path, extract from the session folder, and fix argument passing to visualization.
- `save.py`: add a stimuli position export method
- `task.py`: add export of positions, cleanup on tasks, remove unneeded abc method on task, typing, add calibration tests 
- `feedback.py`: deprecate level feedback (no tests and unused). We could potentially deprecate the Audio feedback for now as well... 
- `matrix/display.py` add capture screenshot method
- `.coveragerc`: ignore the exceptions 
- `requirements.txt`: upgraded to lock PyQt6 dependancies failing Windows builds. See https://github.com/altendky/pyqt-tools/issues/105

## Test

- `make test-all`
- `bcipy --fake --task "Matrix Calibration"` and ensured the output looked correct
- 
## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? in-line and task/README.md updated.

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? TODO!


bcipy/task/paradigm/matrix/calibration.py
```diff
@@ -208,35 +229,43 @@ def write_trigger_data(self, timing: List[Tuple[str, float]], first_run) -> None
...
+                label = offset_label(content_type.name)
+                time = client.offset(
+                    self.matrix.first_stim_time) - self.matrix.first_stim_time
+                triggers.append(Trigger(label, TriggerType.OFFSET, time))
+            self.trigger_handler.add_triggers(triggers)
 
         # make sure triggers are written for the inquiry
         self.trigger_handler.add_triggers(convert_timing_triggers(timing, timing[0][0], self.trigger_type))
 
     def write_offset_trigger(self) -> None:
```
tab-cmd, line 245: Updated this to match the recent update to support multiple device offsets.. I think this is correct @lawhead

bcipy/task/tests/paradigm/rsvp/test_copy_phrase.py
```diff
@@ -23,7 +22,6 @@
 from bcipy.task.paradigm.rsvp.copy_phrase import RSVPCopyPhraseTask
 
 
-@pytest.mark.slow
```
tab-cmd, line 26: I added this to help me on the typing PR. It was merged by accident!

bcipy/display/paradigm/matrix/display.py
```diff
@@ -187,11 +203,12 @@ def build_grid(self) -> Dict[str, visual.TextStim]:
         grid = {}
         for sym in self.symbol_set:
             pos_index = self.sort_order(sym)
+            pos = self.positions[pos_index]
             grid[sym] = visual.TextStim(win=self.window,
                                         text=sym,
                                         color=self.grid_color,
                                         opacity=self.start_opacity,
-                                        pos=self.positions[pos_index],
+                                        pos=pos,
```
celikbasak, line 213: Is this the position in psychopy?

bcipy/helpers/save.py
```diff
@@ -169,3 +182,35 @@ def save_model(model: SignalModel, path: Union[Path, str]):
...
         pickle.dump(model, file, protocol=4)
+
+
+def save_stimuli_position_info(
+        stimuli_position_info: dict,
+        path: Union[Path, str],
+        screen_info: dict) -> str:
+    """Save stimuli positions and screen info to `path`
+
+   stimuli_position_info: {'A': (0, 0)}
```
celikbasak, line 193: Same question here, I think it would be more informative if we mentioned the units here (pyschopy, Tobii or maybe just the screen limits like (0,0), (1,1), (-1,-1))

bcipy/signal/model/offline_analysis.py
```diff
@@ -346,9 +350,10 @@ def analyze_gaze(
             figure_handles = visualize_gaze_inquiries(
                 le, re,
                 means, covs,
-                save_path=None,
+                save_path=save_figures,
+                img_path=f'{data_folder}/{MATRIX_IMAGE_FILENAME}',
                 show=show_figures,
-                raw_plot=True,
+                raw_plot=plot_points,
```
celikbasak, line 356: Good catch!

celikbasak (approved):
Great additions! Thank you for including the calibration tests! 
For symbol positions, maybe we can introduce a parameter/saving option in `helpers/save.py` for saving the units in Tobii for ease of use.

bcipy/display/paradigm/matrix/display.py
```diff
@@ -127,6 +128,21 @@ def _stim_positions(self) -> Dict[str, Tuple[float, float]]:
...
         }
 
+    def capture_grid_screenshot(self, file_path: str) -> None:
+        """Capture Grid Screenshot.
+
+        Capture a screenshot of the current display and save it to the specified filename.
+        """
+        # draw the grid and flip the window
+        self.draw_grid(opacity=self.full_grid_opacity)
+        self.task_bar.current_index = 0
```
lawhead, line 139: I would't expect this method to mutate the object state. It seems like this has the potential to lead to some unexpected bugs. Maybe have a temp variable here and set it back to its previous value after the screenshot?

bcipy/task/paradigm/matrix/calibration.py
lawhead: Thanks, I just noticed this missing. Looks right to me.

bcipy/helpers/save.py
```diff
@@ -169,3 +182,35 @@ def save_model(model: SignalModel, path: Union[Path, str]):
...
+    assert all(isinstance(key, str) for key in stimuli_position_info.keys()), \
+        'stimuli_position_info keys must be strings'
+    assert list(map(type, (stimuli_position_info[key] for key in stimuli_position_info)))[0] is tuple, \
+        'stimuli_position_info values must be a tuple'
+
+    # assert that screen_info is a dict with at least the key 'screen_resolution'
+    assert 'screen_size_pixels' in screen_info.keys(), \
+        'screen_size_pixels must be a key in screen_info'
+    assert isinstance(screen_info['screen_size_pixels'], list), \
+        'screen_size_pixels must be a list'
```
lawhead: Given the amount of input validation required would it make sense to have some stricter parameter types?

bcipy/display/paradigm/matrix/display.py
lawhead: Yes, the `visual.TextStim` is a psychopy display component, so the position information is expected to be a valid psychopy format, depending on the units. When no units are provided it uses the default 'norm' units.

lawhead (approved):

bcipy/helpers/save.py
tab-cmd: That's true with the new typing it should be mostly caught... I will remove some of this (minus key:value assert) and avoid a more explicit data structure for now.

bcipy/helpers/save.py
tab-cmd: Good idea, I'm happy to put those in the export as well. We will need to to some testing to see how these line up. I should have data to look at by EOD today.

PR #304, opened 2023-11-21
VEP calibration

lawhead:
# Overview

Created a new calibration task using the VEPDisplay.

## Ticket

https://www.pivotaltracker.com/story/show/184564388
https://www.pivotaltracker.com/story/show/186126657

## Contributions

- Added a new calibration task
- Updated the task registry
- Refactored the target prompting and symbol animations for the VEP Display
- Added functions for stimuli generation

## Test

- Ran unit tests, linting, and mypy
- Ran VEP Calibration task

## Discussion

For testing this locally, the following timing parameters worked well.
    time_prompt: 4
    time_fixation: 0.5
    time_flash: 4

I created two followup issues:

https://www.pivotaltracker.com/story/show/186522989
https://www.pivotaltracker.com/story/show/186522657


bcipy/parameters/parameters.json
```diff
@@ -367,6 +367,22 @@
...
   },
+  "time_vep_animation": {
+    "value": "1.0",
+    "section": "bci_config",
+    "readableName": "Time VEP Animation (sec)",
+    "helpTip": "The amount of time in seconds for the animation moving symbols to boxes.",
+    "recommended_values": "",
+    "type": "float"
+  },
+  "vep_flicker_rates": {
```
tab-cmd: I don't think we should support this kind of parameter configuration... We could do a range or discuss a different configuration strategy (task params). If this isn't a parameter we need to configure for an experiment, we can also set it ourselves. I'd support a code default and a separate PR to address configurability.

bcipy/display/paradigm/vep/codes.py
```diff
@@ -0,0 +1,78 @@
...
+    if flicker_rate > refresh_rate:
+        raise BciPyCoreException(
+            'flicker rate cannot be greater than refresh rate')
+    if flicker_rate <= 1:
+        raise BciPyCoreException('flicker rate must be greater than 1')
+
+    # get the number of frames per flicker
+    length_flicker = refresh_rate / flicker_rate
+
+    if length_flicker.is_integer():
```
tab-cmd, line 53: I am not sure you added this, so my apologies, but if this checks the type already, we could potentially do a `if not length_flicker.is_integer()` and avoid the cast?

bcipy/display/paradigm/vep/codes.py
```diff
@@ -0,0 +1,78 @@
...
+        length_flicker = int(length_flicker)
+    else:
+        err_message = f'flicker rate={flicker_rate} is not an integer multiple of refresh rate={refresh_rate}'
```
tab-cmd, line 56: We could put the calculated value here too

bcipy/display/paradigm/vep/layout.py
```diff
@@ -165,9 +165,35 @@ def positions(self) -> List[Tuple]:
         left = right_of(layout.left, width / 2)
         right = left_of(layout.right, width / 2)
 
-        positions = [(left, top), (right, top), (left, bottom),
-                     (right, bottom)]
-        if self.num_boxes == 6:
-            positions += [(layout.horizontal_middle, top),
-                          (layout.horizontal_middle, bottom)]
-        return positions
+        if self.num_boxes == 4:
```
tab-cmd: We might consider making these a type (4by4VEP, etc) to help set other properties as well.

bcipy/display/paradigm/vep/vep_stim.py
```diff
@@ -0,0 +1,93 @@
...
+            square_boundary = envelope(pos=square.pos, size=square.size)
+            # squares define the holes in the polygon
+            if square.color == colors[0]:
+                frame2_holes.append(square_boundary)
+            elif square.color == colors[1]:
+                frame1_holes.append(square_boundary)
+
+        # Checkerboard is represented as a polygon with holes, backed by a
+        # simple square with the alternating color.
+        # This technique renders more efficiently and scales better than using
```
tab-cmd, line 53: That's good to know. The only other VEP I could find used the GradientStim. Is the performance gain due to the decrease in objects drawn overall?

bcipy/helpers/list.py
```diff
@@ -34,3 +34,47 @@ def grouper(iterable, chunk_size, incomplete="fill", fillvalue=None):
         return zip(*chunks)
 
     raise ValueError("Expected fill or ignore")
+
+
+def find_index(iterable: List,
```
tab-cmd, line 39: `find_first_index`

bcipy/helpers/list.py
```diff
@@ -34,3 +34,47 @@ def grouper(iterable, chunk_size, incomplete="fill", fillvalue=None):
...
+        lst - list of items to copy
+        length - expands list to this length
+        fill - optional; used to determine which element to use for
+            the fill, given the list. Defaults to the last element.
+
+    >>> expand([1,2,3], length=5)
+    [1,2,3,3,3]
+    """
+    times = length - len(lst)
+    if lst and times > 0:
```
tab-cmd, line 77: Given the type, we can assume this is not None. Perhaps a lst length check?

bcipy/display/paradigm/vep/display.py
```diff
@@ -1,301 +1,364 @@
...
+                 trigger_type: str = 'text',
+                 symbol_set: Optional[List[str]] = None,
+                 codes: Optional[List[List[int]]] = None,
+                 should_prompt_target: bool = True):
         self.window = window
+        frame_rate = self.window.getActualFrameRate()
+        assert frame_rate, 'An accurate window frame rate could not be established'
         self.window_size = self.window.size  # [w, h]
-        self.refresh_rate = round(window.getActualFrameRate())
+        self.refresh_rate = round_refresh_rate(frame_rate)
```
tab-cmd, line 63: We could also support a better util to grab this a few times and average

bcipy/display/paradigm/vep/display.py
```diff
@@ -1,301 +1,364 @@
...
 
-        return timing
+        return self._timing
+
+    def starting_position(self, sym: str) -> Tuple[float, float]:
+        """Get the starting position for the given symbol"""
+        pos_index = self.sort_order(sym)
+        return self.starting_positions[pos_index]
+
+    def prompt_target(self,
```
tab-cmd, line 209: This should return a trigger

bcipy/display/paradigm/vep/display.py
```diff
@@ -1,301 +1,364 @@
...
+                self.set_stimuli_colors(stim)
+                self.prompt_target(target,
+                                   target_box_index=self.box_index(
+                                       stim, target.symbol))
+        else:
+            [_fixation, *stim] = self.stim_properties()
+            self.set_stimuli_colors(stim)
+
+        self.log_inquiry(stim)
+        self.animate_inquiry(stim)
```
tab-cmd, line 183: Another potential trigger return. I could see this being used in eye tracking

bcipy/display/paradigm/vep/display.py
```diff
@@ -1,301 +1,364 @@
...
+        """Emphasize the box at the given index"""
+        for i, box in enumerate(self.text_boxes):
+            if i == target_box_index:
+                box.borderWidth = self.box_border_width + 4
+                box.setOpacity(1.0)
+            else:
+                box.borderWidth = self.box_border_width - 2
+                box.setOpacity(0.8)
+
+    def do_fixation(self, fixation: StimProps) -> None:
```
tab-cmd, line 267: This should also return triggers

bcipy/task/paradigm/vep/calibration.py
```diff
@@ -0,0 +1,261 @@
...
+
+def trigger_type(symbol: str, target: str, _index: int) -> TriggerType:
+    """Trigger Type.
+
+    This method is passed to convert_timing_triggers to properly assign TriggerTypes
+        to the timing of stimuli presented.
+    """
+    if target == symbol:
+        return TriggerType.TARGET
+    return TriggerType.EVENT
```
tab-cmd, line 34: We may need to support different TriggerTypes for this. A good question for the group at demo. What triggers will be needed for reshaping / future analysis. I added a few comments in the display that I know we will want.

bcipy/task/paradigm/vep/calibration.py
```diff
@@ -0,0 +1,261 @@
...
+    win (PsychoPy Display Object)
+    daq (Data Acquisition Object)
+    parameters (Dictionary)
+    file_save (String)
+    """
+
+    def __init__(self, win: visual.Window, daq: ClientManager,
+                 parameters: Parameters, file_save: str):
+        super(VEPCalibrationTask, self).__init__()
+        self.num_boxes = 6
```
tab-cmd, line 56: We could deprecate 4x4 until it is used if we are hardcoding. I'm thinking about long term maintainability and it would prevent some branching logic.

bcipy/task/paradigm/vep/calibration.py
```diff
@@ -0,0 +1,261 @@
...
+
+        self.trigger_handler.add_triggers(triggers)
+        self.trigger_handler.close()
+
+    def name(self):
+        return 'VEP Calibration Task'
+
+
+def init_calibration_display(parameters: Parameters,
+                             window,
```
tab-cmd: typing

bcipy/task/paradigm/vep/stim_generation.py
```diff
@@ -0,0 +1,214 @@
...
+        is_txt(bool): whether the stimuli type is text. False would be an image stimuli.
+
+    Return
+    ------
+        schedule_inq(tuple(
+            samples[list[list[str]]]: list of inquiries
+            timing(list[list[float]]): list of timings
+            color(list(list[str])): list of colors)): scheduled inquiries
+    """
+    if timing is None:
```
tab-cmd, line 43: I think it would be okay to fail without these instead of making optional.

bcipy/task/paradigm/vep/stim_generation.py
```diff
@@ -0,0 +1,214 @@
...
+    inquiries = generate_vep_inquiries(alp, num_boxes, inquiry_count, is_txt)
+    times = [timing for _ in range(inquiry_count)]
+    colors = [color for _ in range(inquiry_count)]
+
+    return InquirySchedule(inquiries, times, colors)
+
+
+def generate_vep_inquiries(symbols: List[str],
+                           num_boxes: int = 6,
+                           inquiry_count: int = 20,
```
tab-cmd, line 65: same here on defaults, it would make sense to have the caller define.

tab-cmd (changes requested):
Huge amount of effort here! I added some comments. The configurability one is the only thing I am requesting changes on before accepting. 

It's ready to time test, I think? Please schedule a time to test it in the lab with me! I think we can use this with a photodiode as-is, but happy to wear the cap if that falls through.

bcipy/helpers/list.py
lawhead: This is checking for an empty list. See https://peps.python.org/pep-0008/.

bcipy/display/paradigm/vep/vep_stim.py
lawhead: Yes, I tried various stim for this and there wasn't a significant difference. The main factor was the number of stim  drawn.

bcipy/display/paradigm/vep/layout.py
lawhead: In this context the num_boxes refers to the number of display areas where a symbol can be sorted. The VEPStim is where the number of squares per checkerboard is defined. If we only ever want to fix this at 6 display areas, I can remove the conditional logic. The other idea here is that if we want a different layout we can provide another class with the same interface for generating these positions.

bcipy/parameters/parameters.json
lawhead: Okay, I'll do some refactoring here. I was mainly concerned about having consistent, known values between the calibration and a future copy phrase. I wasn't sure what values to use so I started with ones close the the Higger paper. I'm fine putting these elsewhere if we don't want them easily changeable.

tab-cmd: Another option is to add List of Int or Float support to our parameters. Perhaps we can check-in with the team during demo to see what the short-term experiment needs are?

bcipy/display/paradigm/vep/layout.py
tab-cmd: I like that last idea! Unless we are going to test 4 vs 6 in the short term, I think it would be easier to maintain cutting it out.

bcipy/display/paradigm/vep/codes.py
lawhead: The cast is needed below (line 62) for list generation: `codes += [t] * length_flicker`. If it's missing you get a TypeError that you can't multiply a sequence by a float.

bcipy/display/paradigm/vep/codes.py
```diff
@@ -0,0 +1,80 @@
...
+import logging
+from typing import List
+
+import numpy as np
+
+from bcipy.helpers.exceptions import BciPyCoreException
+
+log = logging.getLogger(__name__)
+
+DEFAULT_FLICKER_RATES = [4, 5, 6, 10, 12, 15]
```
tab-cmd, line 12: Having tested this now in lab, having these calculated or a default dictionary for different Hz would be helpful! This could live in config.py

bcipy/task/paradigm/vep/calibration.py
```diff
@@ -0,0 +1,262 @@
...
+    box_config = BoxConfiguration(layout, height_pct=0.30)
+
+    stim_props = VEPStimuliProperties(stim_font=parameters['font'],
+                                      stim_pos=box_config.positions,
+                                      stim_height=0.1,
+                                      timing=timing,
+                                      stim_color=colors,
+                                      inquiry=[],
+                                      stim_length=1,
+                                      animation_seconds=1.0)
```
tab-cmd: pull from parameters?

bcipy/task/paradigm/vep/calibration.py
```diff
@@ -0,0 +1,262 @@
...
+
+    task_bar = CalibrationTaskBar(window,
+                                  inquiry_count=parameters['stim_number'],
+                                  current_index=0,
+                                  colors=[parameters['task_color']],
+                                  font=parameters['font'],
+                                  height=parameters['task_height'])
+
+    # issue #186641183 ; determine a better configuration strategy
+    flicker_rates = DEFAULT_FLICKER_RATES
```
tab-cmd: We should log these

bcipy/task/paradigm/vep/stim_generation.py
```diff
@@ -0,0 +1,214 @@
...
+import math
+import random
+from typing import Any, List, Optional
+
+from bcipy.helpers.list import find_index, swapped
+from bcipy.helpers.stimuli import (InquirySchedule, get_fixation,
+                                   random_target_positions)
+
+
+def generate_vep_calibration_inquiries(alp: List[str],
```
tab-cmd, line 12: These could go in helpers with the others. I'd be okay breaking our old pattern but wanna be consistent or make some tickets to clean up.

tab-cmd (approved):
The task runs as expected! Left a few cleanup items. This still needs to be tested more fully in the lab, but I'd be okay merging with some clean up and making follow-up VEP PRs to the rc or closing this and making future VEP PRs to this branch.

bcipy/display/paradigm/vep/codes.py
lawhead: I added more details for this to https://www.pivotaltracker.com/story/show/186641183.

bcipy/task/paradigm/vep/stim_generation.py
lawhead: The stimuli module is getting pretty large and these functions seemed specific to this task. I also think they are more likely to change, so it made sense to me to split these out. I think it would be good to revisit the `stimuli` module and see if some of the things in there would make sense elsewhere. I'll create a ticket for that.

PR #305, opened 2023-12-01
Multimodal offline analysis updates, bug fixes

celikbasak:
# Overview

Updates and bug fixes after the first multimodal PR merge.

## Ticket

[Pivotal ticket](https://www.pivotaltracker.com/story/show/186593908)

## Contributions

- Ensured that the offline analysis runs without error for both EEG and Gaze data.
- Bug that prevents gaze data visualization is fixed.
- Clean-up of unnecessary code in offline analysis. All functions are moved into `bcipy/signal/model` module.
- Two gaze models (`individual` and `centralized`) are completed and made ready for analysis.

## Test

- Unit tests completed

## Documentation

- Documentation pending!

## Changelog

- CHANGELOG.md updated


bcipy/helpers/stimuli.py
```diff
@@ -288,13 +293,17 @@ def __call__(self,
         for symbol in symbol_set:
             data_by_targets[symbol] = []
 
-        window_length = 3  # seconds, total length of flickering after prompt for each inquiry
+        buffer = stimulus_duration / 5  # seconds, buffer for each inquiry
```
tab-cmd, line 296: Why `/5`? This will break when changing the duration I think.

bcipy/signal/model/gaussian_mixture/gaussian_mixture.py
```diff
@@ -61,12 +59,16 @@ def predict(self, test_data: np.ndarray, means: np.ndarray, covs: np.ndarray):
 
         return likelihoods, predictions
 
-    def calculate_acc(self, test_data: np.ndarray, test_labels: Optional[np.ndarray]):
+    def calculate_acc(self, predictions: int, counter: int):
```
tab-cmd, line 62: Are these lists?

bcipy/signal/model/gaussian_mixture/gaussian_mixture.py
```diff
@@ -104,24 +107,45 @@ def evaluate(self, test_data: np.ndarray, sym_pos: np.ndarray):
...
         Compute log-likelihood of each sample.
         Predict the labels for the test data.
         '''
         # Compute over log-likelihood scores
         # Get the argmax of the scores
 
-        scores = self.model.score_samples(test_data)
+        num_components = len(means)
+
+        N, D = test_data.shape
```
tab-cmd, line 120: No single letter variables if possible, that violates PEP

bcipy/signal/model/offline_analysis.py
```diff
@@ -333,115 +338,125 @@ def analyze_gaze(
...
-            means, covs = model.evaluate(test_re)
+            means, covs = model.evaluate()
 
             # Visualize the results:
             figure_handles = visualize_gaze_inquiries(
                 le, re,
                 means, covs,
                 save_path=None,
-                show=show_figures,
+                show=False,
```
tab-cmd, line 356: Why have this call if no saving or showing happens?

bcipy/signal/model/offline_analysis.py
```diff
@@ -333,115 +338,125 @@ def analyze_gaze(
...
-        # Convert test_dict to a list of arrays:
         accuracy = 0
         acc_all_symbols = {}
         counter = 0
         for sym in symbol_set:
-            # Skip if test_dict is empty for certain symbols:
+            # Continue if there is no test data for this symbol:
             if len(test_dict[sym]) == 0:
                 acc_all_symbols[sym] = 0
                 continue
```
tab-cmd, line 381: I don't think this continue is needed...

tab-cmd (approved):
Nice updates! Do the integration tests still pass? Should we write one for the Gaze Model? Don't worry about the Windows build failures. I found a fix.

bcipy/helpers/stimuli.py
```diff
@@ -314,24 +323,24 @@ def __call__(self,
...
             # each symbol. If a target symbol is prompted more than once, the data is appended to the dict as a list.
             # Which is why we need to convert it to a (np.ndarray) and flatten the dimensions.
-            # This is not ideal, but it works for now.
 
-        # return np.stack(reshaped_data, 0), labels
         return data_by_targets
 
     @staticmethod
-    def centralize_all_data(data, symbol_pos):
+    def centralize_all_data(data, symbol_pos) -> np.ndarray:
```
lawhead, line 330: Looks like we need types for the parameters too.

bcipy/helpers/stimuli.py
```diff
@@ -314,24 +323,24 @@ def __call__(self,
...
         """ Using the symbol locations in matrix, centralize all data (in Tobii units).
         This data will only be used in certain model types.
         Args:
-            data (np.ndarray): Data in shape of num_channels x num_samples
+            data (np.ndarray): Data in shape of num_samples x num_dimensions
             symbol_pos (np.ndarray(float)): Array of the current symbol posiiton in Tobii units
         Returns:
-            data (np.ndarray): Centralized data in shape of num_channels x num_samples
+            new_data (np.ndarray): Centralized data in shape of num_samples x num_dimensions
```
lawhead, line 337: I like that you documented that this method does not mutate the input.

lawhead (approved):

bcipy/helpers/stimuli.py
celikbasak: I added a note right under this line:
`This buffer is used to account for the screen downtime between each stimulus. There seems to be a duty cycle of 80% for the stimuli, so we add a buffer of 20% of the stimulus length.` 
I'm not sure why this "duty cycle" occurs, but it's something that we noticed in the Triggers file – the total time between symbol flashes were `0.30 sec` even though the flash time is set as `time_flash = 0.25 sec` in the parameters file of each experiment. So yes, if the duration changes this buffer length will change too (in that case I hope my 80/20 calculation is correct!). 

Fyi, I use this parameter just to calculate the inquiry length:
```python
window_length = (stimulus_duration + buffer) * num_stimuli_per_inquiry
``` 
Instead, I can maybe just calculate the longest inquiry length with a formula similar to the one in InquiryReshaper: 
```python
        def get_inquiry_len(inq_trigs):
            return inq_trigs[-1] - inq_trigs[0]

        longest_inquiry = max(grouper(triggers, trials_per_inquiry, fillvalue='x'), key=lambda xy: get_inquiry_len(xy))
```

bcipy/signal/model/gaussian_mixture/gaussian_mixture.py
celikbasak: These are integers. It was actually only used for plotting purposes (i.e. calculating the decision accuracy per symbol). It was a one liner accuracy calculation method that has nothing to do with the Gaussian Mixture class, so I will remove it.

bcipy/signal/model/gaussian_mixture/gaussian_mixture.py
celikbasak: Updated the variable names

bcipy/signal/model/offline_analysis.py
celikbasak: It actually crashes without it :(

bcipy/signal/model/offline_analysis.py
celikbasak: `visualize_gaze_inquiries` is called for each symbol in this specific loop for debugging purposes. You're right that we don't save or show them during a regular offline analysis, instead we show all symbols together at the end. So I will remove this loop.

bcipy/helpers/stimuli.py
celikbasak: Added types for the input parameters.

PR #306, opened 2023-12-02
Task bar bug

lawhead:
# Overview

Fixed bugs in matrix position calculations and added parameter for task bar spacing.

## Ticket

https://www.pivotaltracker.com/story/show/186539510

## Contributions

- Bug fix in matrix position calculations
- Added height calculation to limit matrix so it won't overlap with the task bar.
- Added parameter to specify the task bar padding.
- Updated demos to use Overpass Mono font.
- Added README for matrix including recommended parameters and information on troubleshooting layout issues.

## Test

- Ran all unit tests
- Ran demos related to matrix displays
- Ran Matrix calibration locally


bcipy/parameters/parameters.json
```diff
@@ -523,6 +523,16 @@
     "recommended_values": "",
     "type": "str"
   },
+  "task_padding": {
+    "value": "0.05",
+    "section": "bci_config",
+    "readableName": "Task Bar Padding",
+    "helpTip": "Specifies the padding around the task bar text. Default: 0.1",
```
tab-cmd: update the default!

tab-cmd (approved):
I like this solution! We can test the fix on the home use machine when we look at VEP together in the lab

PR #308, opened 2023-12-15
Data buffering issue

lawhead:
# Overview

Restructured the `lsl_client` to pull data from LSL for querying in a way that ensures the latest data is available in the buffer for querying.

## Ticket

https://www.pivotaltracker.com/story/show/186539529

## Contributions

- Additional logging in lsl_client to better track what's happening during data pulls.
- Increased buffer size of the LSL StreamInlet so it would not lose data during long pauses (app startup; after letter selection; etc).

## Test

- Ran Copy Phrase task and analyzed logs to ensure that sufficient data was queried to make a decision.

## Discussion

According to the LSL docs (https://labstreaminglayer.readthedocs.io/info/faqs.html), setting the `max_buflen` on a `StreamInlet` should ensure that it keeps only that amount of data (in seconds) and discards the rest. We were relying on this functionality to get data for a decision. However, in recent versions of LSL it seems to sometimes be discarding too much data. This seems to occur after long pauses. When we queried for the latest (2400 records), we would sometimes only get 200-300. I'm still working on creating a minimal example to replicate this problem.

Note also that I am still getting application crashes (Segmentation fault) if the application sits too long on the opening screen or if a task is paused for a while. I'm not sure what's causing this.

### DIFF
--- a/bcipy/acquisition/protocols/lsl/lsl_client.py
+++ b/bcipy/acquisition/protocols/lsl/lsl_client.py
@@ -222,12 +222,12 @@ class LslAcquisitionClient:
         """Pull a chunk of samples from LSL and record in the buffer.
         Returns the count of samples pulled.
         """
-        log.info(f"Pulling from LSL (max_samples: {self.max_samples})")
+        log.debug(f"Pulling from LSL (max_samples: {self.max_samples})")
         # A timeout of 0.0 gets currently available samples without blocking.
         samples, timestamps = self.inlet.pull_chunk(
             timeout=0.0, max_samples=self.max_samples)
         count = len(samples)
-        log.info(f"\tReceived {count} samples")
+        log.debug(f"\tReceived {count} samples")
         for sample, stamp in zip(samples, timestamps):
             self.buffer.append(Record(sample, stamp))
         return count
--- a/bcipy/acquisition/protocols/lsl/lsl_client.py
+++ b/bcipy/acquisition/protocols/lsl/lsl_client.py
@@ -1,7 +1,8 @@
 """DataAcquisitionClient for LabStreamingLayer data sources."""
 import logging
-from typing import Optional, List
+from typing import Dict, List, Optional
 
+import pandas as pd
 from pylsl import (StreamInfo, StreamInlet, local_clock, resolve_byprop,
                    resolve_stream)
 
@@ -54,6 +55,26 @@ class LslAcquisitionClient:
         self.max_buffer_len = max_buffer_len
         self.save_directory = save_directory
         self.raw_data_file_name = raw_data_file_name
+        self._max_samples = None
+
+    @property
+    def first_sample_time(self) -> float:
+        """Timestamp returned by the first sample. If the data is being
+        recorded this value reflects the timestamp of the first recorded sample"""
+        if self.recorder:
+            return self.recorder.first_sample_time
+        return self._first_sample_time
+
+    @property
+    def max_samples(self) -> int:
+        """Maximum number of samples available at any given time."""
+        if self._max_samples is None:
+            if self.device_spec.sample_rate == IRREGULAR_RATE:
+                self._max_samples = int(self.max_buffer_len)
+            else:
+                self._max_samples = int(self.max_buffer_len *
+                                        self.device_spec.sample_rate)
+        return self._max_samples
 
     def start_acquisition(self) -> bool:
         """Connect to the datasource and start acquiring data.
@@ -74,7 +95,7 @@ class LslAcquisitionClient:
 
         self.inlet = StreamInlet(
             stream_info,
-            max_buflen=self.max_buffer_len,
+            max_buflen=4 * self.max_buffer_len,  # TODO: revisit this value
             max_chunklen=1)
 
         if self.device_spec:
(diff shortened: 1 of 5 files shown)


bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -74,7 +95,7 @@ def start_acquisition(self) -> bool:
 
         self.inlet = StreamInlet(
             stream_info,
-            max_buflen=self.max_buffer_len,
```
tab-cmd, line 77: Let's set this as our max buffer length

bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -182,15 +210,28 @@ def get_latest_data(self) -> List[Record]:
...
+        while count == self.max_samples:
+            count = self._pull_chunk()
 
         return self.buffer.get()
 
+    def _pull_chunk(self) -> int:
+        """Pull a chunk of samples from LSL and record in the buffer.
+        Returns the count of samples pulled.
+        """
+        log.info(f"Pulling from LSL (max_samples: {self.max_samples})")
```
tab-cmd: Some of these could be debug level

bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -142,38 +177,31 @@ def get_data(self,
...
+
+        assert start >= data_start, 'Start time out of range'
+        assert end <= data_end, 'End time out of range'
 
         data_slice = [
             record for record in data if start <= record.timestamp <= end
         ][0:limit]
-        log.info(f'{len(data_slice)} records returned')
-        return data_slice
+        log.info(f"Filtered records: {self._data_stats(data_slice)}")
```
tab-cmd, line 202: These data stats seem to be the most helpful / informative !

bcipy/helpers/acquisition.py
```diff
@@ -165,7 +165,7 @@ def init_lsl_client(parameters: dict,
     """Initialize a client that acquires data from LabStreamingLayer."""
 
     data_buffer_seconds = round(max_inquiry_duration(parameters))
-
+    log.info(f"Setting an acquisition buffer of {data_buffer_seconds} seconds")
```
tab-cmd: It might be helpful to know the device this applies to. Right now this logs twice with the same value.

bcipy/acquisition/record.py
```diff
@@ -1,8 +1,12 @@
...
-information, where data is a single reading from a device and is a list
-of channel information (float)."""
+
+class Record(NamedTuple):
+    """Domain object used for storing data and timestamp
+    information, where data is a single reading from a device and is a list
+    of channel information (float)."""
+    data: List[Any]
+    timestamp: float
+    rownum: int = None
```
tab-cmd: Optional[int] = None

bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -126,12 +139,34 @@ def __exit__(self, _exc_type, _exc_value, _traceback):
...
+                'count': len(data),
+                'seconds': round(data_end - data_start, precision),
+                'from': round(data_start, precision),
+                'to': round(data_end, precision),
+                'expected_diff': round(1 / self.device_spec.sample_rate,
+                                       precision),
+                'mean_diff': round(diffs.mean(), precision),
+                'max_diff': round(diffs.max(), precision)
+            }
+        return {}
```
tab-cmd, line 159: Would it make sense to return none? Then a caller could check for that condition instead of failing on a key error.

tab-cmd (approved):
Tested this in the lab and, with a .95 AUC, I was able to type on Matrix and RSVP. This seems to fix the issue! Below are the logs from the session.

[Matrix Calibration Logs](https://github.com/CAMBI-tech/BciPy/files/13756327/bcipy_system_log.txt)
[Matrix Copy Phrase Logs](https://github.com/CAMBI-tech/BciPy/files/13756331/bcipy_system_log.txt)

There are some logging duplicates and potential to change log levels added. Thanks for the quick fix!

bcipy/helpers/acquisition.py
lawhead: Good suggestion

bcipy/acquisition/protocols/lsl/lsl_client.py
lawhead: It's a little confusing because there are two buffers to consider here. One is the buffer that LSL uses for the `StreamInlet`. The other is the size of our `RingBuffer` that the lsl_client uses internally. The `self.max_buffer_len` is the size used by the `RingBuffer`. These two buffers used to use the same `max_buffer_len` without issue, but for some reason that is no longer the case and we need to increase the value for the `StreamInlet` to have the data we're looking for.

The reason we need to maintain the secondary `RingBuffer` is so we can do subsequent queries for the same data. Queries to the `StreamInlet` will drain that buffer.

bcipy/acquisition/protocols/lsl/lsl_client.py
lawhead: My primary use case for the dict is to print out the whole thing. If it might be a `None` value I would need some additional checks. I think in this case if we're not sure that a key would be in the map it would be cleaner to use a `map.get(key, default_value)`.

PR #310, opened 2024-01-06
Device Status

lawhead:
# Overview

Added active/passive status to devices.

## Ticket

https://www.pivotaltracker.com/story/show/186164071

## Contributions

- Added a new field to `DeviceSpec` for the device status. Updated the loading / saving code to read/write the field. Updated `devices.json` to use a default value of 'active'.
- Updated acquisition.py module to allow for setting this field using the parameters.
- Updated offline_analysis to only train models for active devices.
- Updated Copy Phrase task to only create evidence_evaluators for active devices.
- Added unit tests

## Test

- Ran all unit tests
- Ran a calibration task with "acq_mode": "EEG+Eyetracker:passive", followed by offline analysis to confirm that only the EEG model was trained. Changed the `devices.json` in the calibration folder, ran the offline analysis, and confirmed that the code attempted to train both models.


tab-cmd (approved):
This looks good! 

I looked around and couldn't find any documentation on configuring devices. I'm not sure what the best location would be, but a small blurb about setting active/passive in the main README could be helpful (maybe near the core modules section).

lawhead:
Ugh, sorry about that. I forgot to push the commit with the documentation.

PR #314, opened 2024-02-07
Add Data Source

lawhead:
# Overview

Added the data source to the sampled trial data to allow us to confirm that the sampled likelihoods match the values in the session data.

## Contributions

- Created a named tuple for storing Trial data; documented the fields.
- Refactored data engine to simplify the needed data processing by moving some of this to computed fields in the ExtractedExperimentData class

## Test

- Ran the simulator and spot checked that the sampled values in the log matched the EEG likelihoods in the corresponding session.xlsx file.


sreekaroo:
thanks for refactor 👍

PR #319, opened 2024-03-18
Added new sampler to sample an Inquiry at a time

lawhead:
# Overview

Created a new Sampler for the Simulator to use an Inquiry at  a time.

## Ticket

https://www.pivotaltracker.com/story/show/187262169

## Contributions

- Created a new sampler to test our architecture for changing sampling strategies.
- Added a new logging utility to help with logging DataFrames.

## Test

- Ran the simulator using the new sampler. Inspected the log files to confirm that for each simulated inquiry, all of the EEG likelihoods were coming from the same source Inquiry. Also confirmed that the source was different for each simulated query. This works for both inquiries with a target and those without.

### DIFF
--- a/bcipy/simulator/helpers/sampler.py
+++ b/bcipy/simulator/helpers/sampler.py
@@ -1,6 +1,9 @@
 import logging
+import random
 from abc import ABC
-from typing import Callable, List
+from collections import defaultdict
+from pathlib import Path
+from typing import Callable, Dict, List, Tuple
 
 import numpy as np
 import pandas as pd
@@ -9,7 +12,8 @@ from bcipy.helpers.exceptions import TaskConfigurationException
 from bcipy.helpers.parameters import Parameters
 from bcipy.helpers.symbols import alphabet
 from bcipy.simulator.helpers.data_engine import RawDataEngine
-from bcipy.simulator.helpers.log_utils import format_sample_rows
+from bcipy.simulator.helpers.log_utils import (format_sample_df,
+                                               format_sample_rows)
 from bcipy.simulator.helpers.state_manager import SimState
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


bcipy/simulator/helpers/sampler.py
```diff
@@ -79,3 +87,118 @@ def __default_reshaper(self, eeg_responses: List[np.ndarray]) -> np.ndarray:
...
+        for c_i, channel_eeg in enumerate(trial_channels_eeg):
+            channels_eeg[c_i].append(channel_eeg)
+
+    return np.array(channels_eeg)
+
+
+class InquirySampler(Sampler):
+    """Samples an inquiry of data at a time."""
+
+    def __init__(self, data_engine: RawDataEngine, *args, **kwargs):
```
sreekaroo: Is it possible to remove the` *args` and `**kwargs` arguments? I'd like the samplers to have pretty explicit APIs if possible.

bcipy/simulator/helpers/sampler.py
```diff
@@ -79,3 +87,118 @@ def __default_reshaper(self, eeg_responses: List[np.ndarray]) -> np.ndarray:
...
+                                   & (self.data['inquiry_n'] == inquiry_n)]
+        assert len(inquiry_df) == len(
+            inquiry_letter_subset), f"Invalid data source {data_source}"
+
+        inq_target_pos = target_pos
+        if target_letter in inquiry_letter_subset:
+            inq_target_index = inquiry_df[inquiry_df['target'] == 1].index[0]
+            inq_target_pos = inquiry_df.index.tolist().index(inq_target_index)
+
+        if inq_target_pos == target_pos:
```
sreekaroo: Would you mind adding some brief comments here explaining logic

sreekaroo (commented):
Nice, it seems like the architecture was convenient to implement new sampling logic. This PR would serve as good reference for developing new future Samplers

PR #320, opened 2024-04-15
Orchestrator parameters

Carsonthemonkey:
# Overview

Added functions for parsing, serializing, and validating a series of Orchestrator actions

## Ticket

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/187263322)

## Contributions

- Adds function for parsing string of actions into a list of tasks (actions are not defined as anything beyond task types as of now)
- Adds function for validating that string of actions is in the correct format
- Adds function for converting a list of tasks into an action string

## Test

- Adds unit test for parsing a single action, multiple actions, and invalid actions
- Adds unit test for serializing a single action and multiple actions
- Adds task for validating valid and invalid action strings

## Documentation

- Functions are documented with docstrings

## Changelog

- changelog updated


bcipy/orchestrator/parameters/actions.py
```diff
@@ -0,0 +1,85 @@
...
+        action_sequence : str
+            A string of actions in the format of 'Action1 -> Action2 -> ... -> ActionN'.
+
+    Returns
+    -------
+        List[TaskType]
+            A list of TaskType objects that represent the actions in the input string.
+    """
+    try:
+        actions = [taskname_dict[action.strip()] for action in action_sequence.split('->')]
```
lawhead: You might want to have a parameter for the delimiter or use a constant for this. It's repeated several places and it would be easy to miss one if we decided to change this value.

Also, you can just return here. You don't need to create a variable for `actions`.

bcipy/orchestrator/parameters/actions.py
```diff
@@ -0,0 +1,85 @@
+from typing import List
+from bcipy.task import TaskType
+"""This file can define actions that can happen in a session orchestrator visit. To start these will be 1:1 with tasks, but later this can be extended to represent training sequences, GUI popups etc."""
+
+taskname_dict = {}
+for i, task in enumerate(TaskType.list()):
+    taskname_dict[task] = TaskType(i + 1)
+
+
+def parse_actions(action_sequence: str) -> List[TaskType]:
```
lawhead: As discussed in our meeting, Tasks and Actions are not going to be the same thing, so this name should change. It could just be `parse` and the parameter name `sequence`. The return type would need to a be a union of TaskType and Action.

bcipy/orchestrator/parameters/actions.py
```diff
@@ -0,0 +1,85 @@
+from typing import List
```
lawhead: You will probably want to use the module `actions` for defining the `Action` classes. The functionality in here is really about configuration, so I would recommend renaming this as `config.py` or `configuration.py`. I also don't think this needs to be nested in a parameters directory. It can be a top-level module within orchestrator.

bcipy/orchestrator/tests/test_actions.py
```diff
@@ -0,0 +1,43 @@
+import unittest
+from bcipy.orchestrator.parameters.actions import parse_actions, serialize_actions, validate_action_string
+from bcipy.task import TaskType
+
+
+class TestActions(unittest.TestCase):
```
lawhead, line 6: This is a nice set of tests. 👍

lawhead (changes requested):
The code looks good. I would just like some things named a bit differently to clarify the intent.

bcipy/orchestrator/config.py
```diff
@@ -0,0 +1,87 @@
...
+        action_sequence : str
+            A string of actions in the format of 'Action1 -> Action2 -> ... -> ActionN'.
+
+    Raises
+    ------
+        ValueError
+            If the string of actions is invalid.
+    """
+    try:
+        for action in action_sequence.split('->'):
```
lawhead: Here is another place to use the separator constant.

bcipy/orchestrator/config.py
```diff
@@ -0,0 +1,87 @@
...
+    Returns
+    -------
+        List[TaskType]
+            A list of TaskType objects that represent the actions in the input string.
+    """
+    action_str = ''
+    for i, task in enumerate(action_sequence):
+        action_str += task.label
+        if i != len(action_sequence) - 1:
+            action_str += ' ' + ACTION_SEPARATOR + ' '
```
lawhead: This could be more concise by using the string `join` method.

`return  f" {ACTION_SEPARATOR} ".join([task.label for task in action_sequence])`

lawhead (approved):

lawhead:
Try to get the checks passing before merging. I think there are some typing issues to address.

PR #321, opened 2024-04-25
Pause error

lawhead:
# Overview

- Fixed an issue where BciPy would crash after a long pause. 
- Fixed a bug with segfaults when running BciPy on Macs.

## Ticket

https://www.pivotaltracker.com/story/show/187261791

## Contributions

- Refactored the acquisition module to use multiprocessing, rather than threading (see discussion below). 
- Added a configuration for the max number of seconds allowed for pausing.
- UI handling to end a task after a max pause duration has been reached.

### Segfault issue

The LSL library has methods which must be run on the main thread of execution on linux/Mac computers. Running them in a separate thread was causing segfaults to occur. By refactoring to use multiprocessing, each of the objects that use LSL are essentially acting as the main thread, which resolves this issue. Note that we should do a timing test to confirm that the static offsets are still the same.

### Pause bug

Previously we calculated the max buffer for LSL based on the maximum possible duration of an inquiry. However, we did not account for pause length. I was making the assumption that the LSL buffer would continue to accumulate data and only retain the latest (like what would happen in a ring buffer data structure). However, LSL has undefined behavior when the StreamInlet `max_buflen` is exceeded. When you attempt to retrieve data after this maximum is exceeded some of the data is missing and what is returned may not even be in monotonic order. I was able to replicate this behavior without using BciPy in the `lsl_buffer_test.py` script. This script will allow us to test future versions of LSL in case we want to change our approach. Currently, the resolution is to set a fixed maximum buffer size in the configuration that accounts for the maximum amount of time we may want to pause during a session. If this is exceeded the task gracefully shuts down.

## Test

- Ran a calibration task and paused on the wait screen (before starting the task) for different durations to confirm the behavior is correct.
- Ran the calibration task and paused after one or more inquiries. Confirmed that the task was able to resume if the wait time was less than the configured max.

### DIFF
--- a/bcipy/acquisition/protocols/lsl/lsl_client.py
+++ b/bcipy/acquisition/protocols/lsl/lsl_client.py
@@ -19,6 +19,8 @@ from bcipy.helpers.clock import Clock
 
 log = logging.getLogger(__name__)
 
+LSL_TIMEOUT = 5.0  # seconds
+
 
 def time_range(stamps: List[float],
                precision: int = 3,
@@ -108,7 +110,7 @@ class LslAcquisitionClient:
         stream_info = resolve_device_stream(self.device_spec)
         self.inlet = StreamInlet(
             stream_info,
-            max_buflen=MAX_PAUSE_SECONDS + 5,
+            max_buflen=MAX_PAUSE_SECONDS,
             max_chunklen=1)
         log.info("Acquiring data from data stream:")
         log.info(self.inlet.info().as_xml())
@@ -127,10 +129,11 @@ class LslAcquisitionClient:
                 queue=msg_queue)
             self.recorder.start()
             log.info("Waiting for first sample from lsl_recorder")
-            self._first_sample_time = msg_queue.get(block=True, timeout=5.0)
+            self._first_sample_time = msg_queue.get(block=True,
+                                                    timeout=LSL_TIMEOUT)
             log.info(f"First sample time: {self.first_sample_time}")
 
-        self.inlet.open_stream(timeout=5.0)
+        self.inlet.open_stream(timeout=LSL_TIMEOUT)
         if self.max_buffer_len and self.max_buffer_len > 0:
             self.buffer = RingBuffer(size_max=self.max_samples)
         if not self._first_sample_time:
@@ -369,7 +372,7 @@ def discover_device_spec(content_type: str) -> DeviceSpec:
     """Finds the first LSL stream with the given content type and creates a
     device spec from the stream's metadata."""
     log.info(f"Waiting for {content_type} data to be streamed over LSL.")
-    streams = resolve_byprop('type', content_type, timeout=5.0)
+    streams = resolve_byprop('type', content_type, timeout=LSL_TIMEOUT)
     if not streams:
         raise Exception(
             f'LSL Stream not found for content type {content_type}')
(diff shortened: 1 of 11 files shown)


bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -86,34 +105,36 @@ def start_acquisition(self) -> bool:
...
-        if not streams:
-            raise Exception(
-                f'LSL Stream not found for content type {content_type}')
-        stream_info = streams[0]
-
+        stream_info = resolve_device_stream(self.device_spec)
         self.inlet = StreamInlet(
             stream_info,
-            max_buflen=4 * self.max_buffer_len,  # TODO: revisit this value
+            max_buflen=MAX_PAUSE_SECONDS + 5,
```
tab-cmd: To help with later debugging, I would advocate adding the `5` to the max pause seconds or making it a class variable `_max_buffer_padding`or similar.

bcipy/acquisition/protocols/lsl/lsl_client.py
```diff
@@ -86,34 +105,36 @@ def start_acquisition(self) -> bool:
...
+                directory=self.save_directory,
+                filename=self.raw_data_file_name,
+                device_spec=self.device_spec,
+                queue=msg_queue)
             self.recorder.start()
+            log.info("Waiting for first sample from lsl_recorder")
+            self._first_sample_time = msg_queue.get(block=True, timeout=5.0)
+            log.info(f"First sample time: {self.first_sample_time}")
 
+        self.inlet.open_stream(timeout=5.0)
```
tab-cmd: or perhaps a `timeout` variable

bcipy/config.py
```diff
@@ -51,5 +51,6 @@
 
 # misc configuration
 WAIT_SCREEN_MESSAGE = 'Press Space to start or Esc to exit'
+MAX_PAUSE_SECONDS = 360
```
tab-cmd: Is this a variable we would ever want to configure in the parameters? I can imagine it being an experiment control along with max # of pauses. This pattern works well if not.

bcipy/helpers/task.py
```diff
@@ -314,38 +315,45 @@ def get_user_input(window, message, color, first_run=False):
...
+    -------
+        True to resume; False to exit.
+    """
+    pause_start = time.time()
+    window.wait_screen(message, color)
+    keys = event.waitKeys(keyList=['space', 'escape'])
+
+    elapsed_seconds = time.time() - pause_start
+    if elapsed_seconds >= MAX_PAUSE_SECONDS:
+        log.info("Pause exceeded the allowed time. Ending task.")
```
tab-cmd: Let's add the MAX_PAUSE_SECONDS to the error message to help with debugging

tab-cmd (approved):
Nice catch here! LSL keeping us on our toes! This should be a more resilient way to use it overall

tab-cmd:
> Nice catch here! LSL keeping us on our toes! This should be a more resilient way to use it overall

I added a bug ticket for the failing MacOS builds: https://www.pivotaltracker.com/story/show/187517298

bcipy/config.py
lawhead: Good question. I think my preference is to leave it here until it comes up.

PR #322, opened 2024-05-10
VEP calibration session data

lawhead:
# Overview

Added session.json output for VEP Calibration task to capture data needed for creating a model.

## Ticket

https://www.pivotaltracker.com/story/show/186522989

## Contributions

- Updated task data module to allow for task-specific Inquiry and Session data
- Added properties to display elements useful for recording in the session data.
- Added functionality to VEP Calibration task for outputting session data. General task data includes information about each box (coordinates, colors, flicker rate). Each Inquiry has information about which stimuli were presented in each box, which box contained the target symbol, as well as the flicker rate of that box.

## Test

- Ran all unit tests
- Ran a VEP calibration task and inspected the resulting session.json file.

### DIFF
--- a/bcipy/task/paradigm/vep/calibration.py
+++ b/bcipy/task/paradigm/vep/calibration.py
@@ -1,5 +1,5 @@
 """VEP Calibration task-related code"""
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 from psychopy import core, visual  # type: ignore
 
@@ -35,23 +35,6 @@ def trigger_type(symbol: str, target: str, _index: int) -> TriggerType:
     if target == symbol:
         return TriggerType.TARGET
     return TriggerType.EVENT
-
-
-def target_index(inquiry: List[Union[List[str], str]]) -> Optional[int]:
-    """Given a VEP calibration inquiry, determine which box holds the target.
-
-    Parameters
-    ----------
-        inquiry - list of target, fixation, and sublists representing each
-            vep box and its content symbols.
-
-    Returns the box index (0-based).
-    """
-    target, _fixation, *boxes = inquiry
-    for i, box in enumerate(boxes):
-        if target in box:
-            return i
-    return None
 
 
 class VEPCalibrationTask(Task):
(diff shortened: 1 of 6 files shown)


bcipy/task/paradigm/vep/calibration.py
```diff
@@ -226,14 +302,15 @@ def init_calibration_display(parameters: Parameters,
...
-                                      stim_height=0.1,
-                                      timing=timing,
-                                      stim_color=colors,
-                                      inquiry=[],
-                                      stim_length=1,
-                                      animation_seconds=parameters['time_vep_animation'])
+    stim_props = VEPStimuliProperties(
+        stim_font=parameters['font'],
+        stim_pos=box_config.positions,
+        stim_height=0.1,
```
tab-cmd, line 306: This may be a good class variable so we can change it with task parameters as needed

tab-cmd (approved):
This looks good to me! I sent the output to @celikbasak to check it had all the information needed for our models.

Here are my notes for things to address in future PRs:
- I wasn't able to get the time spent output in the session.json on my machine. 
-  I encountered a weird double animation during inquiries. I'll send you a screen capture on Discord (too large for GitHub). 
- Fake data capability. I ended up hardcoding my frame rate and hacking in the fake parameter from start_task. We might consider adding that feature to help with development. This may be related to the double animation I experienced though!

lawhead:
The double animation happens if the frame rate is incorrect. There is a VEP Display parameter for manually setting the frame rate, which overrides the psychopy computed value, which doesn't seem to work on my M2 Mac. You can try setting this value and see if that fixes it for you. I had to set this to 120hz even though I have my Display settings set to 60hz. I set this in the VEP calibration module, but I agree that exposing this as a parameter might be useful.

I'm not sure I'm computing the time_spent. I'll double-check that and add it to this PR.

tab-cmd:
> The double animation happens if the frame rate is incorrect. There is a VEP Display parameter for manually setting the frame rate, which overrides the psychopy computed value, which doesn't seem to work on my M2 Mac. You can try setting this value and see if that fixes it for you. I had to set this to 120hz even though I have my Display settings set to 60hz. I set this in the VEP calibration module, but I agree that exposing this as a parameter might be useful.
> 
> I'm not sure I'm computing the time_spent. I'll double-check that and add it to this PR.

Okay! @celikbasak said this should be enough to train VEP models. We may find better ways to organize/make this available online, but I think it's worth waiting to test out / encounter pain points.

PR #325, opened 2024-05-27
Add BciPy Report 

tab-cmd:
# Overview

Add BciPy reporting functionality using [reportlab](https://docs.reportlab.com/). This PR implements extendable Report and ReportSection classes that can be used to generate a PDF report. Two report sections are tested,

- SignalReportSection (MNE ERP Plots used in offline analysis)
- SessionReportSection (Any Dict to be displayed in the report)

## Ticket

https://www.pivotaltracker.com/story/show/185944301

## Contributions

- `report.py`: Add Report, ReportSection, SignalReportSection, SessionReportSection. 

## Test

- `make test-all`
- Ran `demo_visualization.py` adding a breakpoint at the end. Importing the needed classes (`from bcipy.helpers.report import Report, SignalReport, SessionReport`) and running:
```python
    report = Report('.')
    session = {'session': 1, 'date': '2021-10-01'}
    session_text = SessionReportSection(session)
    report.add(session_text)
    sr = SignalReportSection(figure_handles)
    report.add(sr)
    report.compile()
    report.save()
```
This will output the following: 
[BciPyReport.pdf](https://github.com/CAMBI-tech/BciPy/files/15462241/BciPyReport.pdf)

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes.

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? done!


bcipy/helpers/report.py
```diff
@@ -0,0 +1,220 @@
...
+    """Report Section.
+
+    An abstract class to handle the creation of a section in a BciPy Report.
+    """
+
+    def _compile(self) -> Flowable:
+        ...
+
+    def _create_header(self) -> Flowable:
+        ...
```
lawhead, line 32: Given that the `_compile` and `_create_header` methods are intended to be called by an external class (`Report`), should this still have a leading underscore?

bcipy/helpers/report.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+class SignalReport(ReportSection):
```
lawhead: Naming is always hard. It would be more consistent to name this `SignalReportSection`, since it subclasses `ReportSection` rather than `Report`. Alternatively we could rename `Report` to `ReportBuilder` or something like that.

lawhead (approved):

bcipy/helpers/report.py
tab-cmd: The intention is that it should not be called individually, but only on a final build. The create header is something that should only be called internally. I will remove the underscore to compile and add a note in the docstring.

PR #328, opened 2024-06-13
Orchestrator actions

Carsonthemonkey:
# Overview

Adds actions. These allow previously non task procedures to be run through the orchestrator

## Ticket

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/187693597)

## Contributions

- Add `CodeHookAction`
- Add `CallbackAction`
- Add `OfflineAnalysisAction`
- Add orchestrator run save structure
- Refactor task registry slightly to allow for action parsing

## Test

- Ensure existing parsing tests are passing

## Documentation

- Comments have been added to necessary code


Carsonthemonkey:
Task parsing tests are not working at the moment. I am refactoring this area of the code to hopefully be more extensible on `task_registry_refactor`

bcipy/orchestrator/actions.py
```diff
@@ -0,0 +1,124 @@
...
+import subprocess
+
+
+class CallbackAction(Task):
+    """
+    Action for running a callback.
+    """
+
+    def __init__(self, callback: callable, *args, **kwargs) -> None:
+        super(CallbackAction, self).__init__()
```
lawhead: We're a little inconsistent about this but the `super` calls can be simplified to `super().__init__()`

bcipy/orchestrator/actions.py
```diff
@@ -0,0 +1,124 @@
...
+        self.callback = callback
```
lawhead, line 23: So the other way to do this would be to have a `callback` method in this class, and then this class could be subclassed for each new action. Both ways would work and I'm not sure which is better but this does feel a bit like mixing paradigms.

bcipy/orchestrator/actions.py
```diff
@@ -0,0 +1,124 @@
...
+        return self.data_directory
+
+    def construct_command(self):
+        command = 'python -m bcipy.signal.model.offline_analysis'
+        command += ' --data_folder ' + self.data_directory
+        if self.parameters_path:
+            command += ' --parameters_file ' + self.parameters_path
+        if self.alert:
+            command += ' --alert'
+        return command
```
lawhead, line 85: Any reason not to put this all in the constructor?

bcipy/orchestrator/actions.py
```diff
@@ -0,0 +1,124 @@
...
+from bcipy.task.paradigm.rsvp.calibration.calibration import \
+    RSVPCalibrationTask
+from bcipy.task.paradigm.rsvp.calibration.timing_verification import \
+    RSVPTimingVerificationCalibration
+from bcipy.task.paradigm.rsvp.copy_phrase import RSVPCopyPhraseTask
+from bcipy.task.paradigm.vep.calibration import VEPCalibrationTask
+from bcipy.orchestrator.actions import OfflineAnalysisAction
+from bcipy.orchestrator.actions import CodeHookAction
+from bcipy.orchestrator.actions import CallbackAction
+
```
lawhead: Imports should be at the top.

bcipy/orchestrator/config.py
```diff
@@ -27,13 +34,13 @@ def parse_sequence(sequence: str) -> List[TaskType]:
             A list of TaskType objects that represent the actions in the input string.
     """
     try:
-        actions = [taskname_dict[action.strip()] for action in sequence.split(ACTION_SEPARATOR)]
+        sequence = [task_registry_dict[action.strip()] for action in sequence.split(ACTION_SEPARATOR)]
```
lawhead: Renaming 'action' to 'task' might be clearer.

bcipy/orchestrator/orchestrator.py
```diff
@@ -0,0 +1,102 @@
...
+        self.experiment_id = experiment_id
+        self.log = logging.getLogger(__name__)
+        self.sys_info = get_system_info()
+        self.tasks = []
+        self.session_data = []
+        self.init_orchestrator_save_folder(self.parameters["data_save_loc"])
+
+        self.ready_to_execute = False
+
+    def add_task(self, task) -> None:
```
lawhead: The 'task' parameter could use a type annotation.

lawhead (approved):
This is some good progress. I like the decision to make an Action a subclass of Task. As you mentioned in the comments, TaskRegistry will likely need a bit of additional work. In particular we may want some way to distinguish between spelling tasks and actions for various reasons (for instance, in main.py we provide a list of available tasks). For this PR most of my comments are fairly minor cleanup.

bcipy/orchestrator/actions.py
Carsonthemonkey: Nice, I figured that would work but I wasn't sure if there was a reason to do it the longer way

bcipy/orchestrator/actions.py
Carsonthemonkey: In this pattern, how would the argument passing to the function be handled? It makes sense to me to use inheritance, but I do like the simplicity of just passing it a function and some args at the moment

bcipy/orchestrator/actions.py
```diff
@@ -0,0 +1,120 @@
...
+# TODO: Refactor this, and TaskInfo. This is currently redundant.
```
tab-cmd: Let's pull this into a ticket and remove the TODO from code

bcipy/orchestrator/config.py
```diff
@@ -1,16 +1,23 @@
...
 To start these will be 1:1 with tasks, but later this can be extended to represent training sequences, GUI popups etc"""
 
-ACTION_SEPARATOR = '->'
+from typing import List
+from bcipy.task import Task, TaskType
+from bcipy.orchestrator.actions import task_registry_dict
+
+TASK_SEPERATOR = '->'
+
+# task_registry_dict = {}
```
tab-cmd: remove unused code

bcipy/orchestrator/config.py
```diff
@@ -1,16 +1,23 @@
...
-from bcipy.task import TaskType
 """This file can define actions that can happen in a session orchestrator visit.
 To start these will be 1:1 with tasks, but later this can be extended to represent training sequences, GUI popups etc"""
 
-ACTION_SEPARATOR = '->'
+from typing import List
+from bcipy.task import Task, TaskType
+from bcipy.orchestrator.actions import task_registry_dict
+
+TASK_SEPERATOR = '->'
```
tab-cmd: Move to config

bcipy/orchestrator/orchestrator.py
```diff
@@ -0,0 +1,102 @@
...
+from bcipy.helpers.parameters import Parameters
+from bcipy.helpers.validate import validate_experiment
+from bcipy.helpers.system_utils import get_system_info
+from bcipy.task import Task
+from bcipy.config import DEFAULT_EXPERIMENT_ID, DEFAULT_PARAMETERS_PATH, DEFAULT_USER_ID
+from bcipy.signal.model import SignalModel
+from bcipy.language.main import LanguageModel
+from bcipy.helpers.load import load_json_parameters
+
+# Session Orchestrator Needs:
```
tab-cmd, line 26: Add helpful information to class docstring

bcipy/orchestrator/orchestrator.py
```diff
@@ -0,0 +1,102 @@
...
+# - A way to initialize the session (user, experiment, tasks, parameters, models, system info, log, save folder)
+#   - save folder is not used in execute method and could be from a provided argument or from the parameters?
+# - A way to save the session data
+
+
+# Test SessionOrchestrator using Actions
```
tab-cmd: Remove

bcipy/orchestrator/orchestrator.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def __init__(
+        self,
+        experiment_id: str = DEFAULT_EXPERIMENT_ID,
+        user: str = DEFAULT_USER_ID,
+        parameters_path: str = DEFAULT_PARAMETERS_PATH,
+    ) -> None:
+        validate_experiment(experiment_id)
+        self.parameters_path = (
+            parameters_path  # TODO: load parameters and cast them to the correct values
```
tab-cmd: Is this TODO still relevant? I see a load below

bcipy/orchestrator/tests/test_config.py
```diff
@@ -0,0 +1,73 @@
+import unittest
+from bcipy.orchestrator.config import parse_sequence, serialize_sequence, validate_sequence_string
+from bcipy.task import TaskType
+from bcipy.orchestrator.actions import OfflineAnalysisAction
+
+
+class TestConfig(unittest.TestCase):
```
tab-cmd: This could be more explicit to what we are testing.

tab-cmd (approved):
Looks good! Let's clean this up (docstrings/todo) and get @lawhead's response to your question before merging

PR #331, opened 2024-07-11
Static Offset per Device

lawhead:
# Overview

Refactored code to allow configuration of a static_offset value for each device, rather than a global value set in the parameters.json.

## Ticket

https://www.pivotaltracker.com/story/show/187737714

## Contributions

- Added a static_offset to the DeviceSpec; added a new default value for each device in devices.json
- Updated any code using previous offset from the parameters file to read the offset for the device of interest (in particular, see offline_analysis and copy phrase)
- Fixed unit tests

## Test

- Ran all unit tests
- Ran RSVP Calibration -> Offline Analysis -> Copy Phrase to ensure that typing still worked.

### DIFF
--- a/bcipy/acquisition/tests/test_client_manager.py
+++ b/bcipy/acquisition/tests/test_client_manager.py
@@ -15,6 +15,7 @@ class TestClientManager(unittest.TestCase):
         self.eeg_device_mock.content_type = 'EEG'
         self.eeg_device_mock.sample_rate = 300
         self.eeg_device_mock.is_active = True
+        self.eeg_device_mock.static_offset = 0.1
 
         self.eeg_client_mock = Mock()
         self.eeg_client_mock.device_spec = self.eeg_device_mock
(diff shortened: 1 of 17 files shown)


tab-cmd (approved):
Thanks for this update! This looks exactly as I would have envisioned it. In offline analysis, if wanting to use an offset for a different device, we'd only need to pull it from the spec. @celikbasak will need to review but this should work well for us.

bcipy/acquisition/tests/test_client_manager.py
```diff
@@ -26,6 +27,7 @@ def setUp(self):
         self.gaze_device_mock.content_type = 'Eyetracker'
         self.gaze_device_mock.sample_rate = 60
         self.gaze_device_mock.is_active = False
+        self.gaze_device_mock.static_offset = 0.0
```
celikbasak, line 30: It is currently assumed 0.0 for my analyses as well, but I will think of a way to estimate the correct value

bcipy/parameters/devices.json
```diff
@@ -120,6 +124,7 @@
         "sample_rate": 60.0,
         "description": "Tobii Nano. For use with the Tobii Pro SDK.",
         "excluded_from_analysis": ["device_ts", "system_ts", "left_pupil", "right_pupil"],
-        "status": "active"
+        "status": "active",
+        "static_offset": 0.0
```
celikbasak, line 128: Let's keep it 0 for now

celikbasak (approved):

PR #333, opened 2024-07-19
Experiment Protocol Gui

Carsonthemonkey:
# Overview

Add ability to define task protocols in the experiment gui

## Ticket

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/187701295)

## Contributions

- Add additional dropdown that displays all available tasks
- Add UI elements for reordering and removing tasks
- Add protocol saving to `experiment.json`
- Refactor formatting for create experiment button

## Test

- No testing for GUI besides manually inspecting app.

## Documentation

- Buttons in the new interface are clearly labeled and code is commented.

## Changelog

- Changelog updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-# Experiment Action
-
+# Experiment Refactor
 - Refactors the Experiment Field Collection GUI to be an action
+- Allows task protocol to be defined in the orchestrator
 
 # Orchestrator Parameters
 Adds parsing functions for action sequences required for the Session Orchestrator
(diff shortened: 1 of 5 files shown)


tab-cmd (approved):
I like the up/down ordering. Thanks for the addition! What PR do we want to update BCInterface? Currently, it fails due to TaskType import. Looking at the tickets in pivotal it is not clear to me!

Carsonthemonkey:
> I like the up/down ordering. Thanks for the addition! What PR do we want to update BCInterface? Currently, it fails due to TaskType import. Looking at the tickets in pivotal it is not clear to me!

Since this the main entry point to starting a session it would make sense to me to wait until orchestrator is a little more fleshed out since it is required to run the task protocol. I think It would also make sense for me to refactor the GUI code before this since it will be affected by it.

PR #336, opened 2024-08-06
Artifact Detection and BCI-Fit Artifact Report

tab-cmd:
# Overview

Added an Artifact Detection class to the `signal.evaluate` module. Updated the artifact report section to include information about artifacts detected.

## Ticket

https://www.pivotaltracker.com/story/show/187526878
https://www.pivotaltracker.com/story/show/187833646

## Contributions

- `artifact.py`: module added to detect and report on artifacts in EEG data. This replaces the old evaluator and rules modules.
- `report.py`: updated the artifact report to extract relevant information and add to the PDF report.

## Test

- `make test-all` & using the `demo_report.py` to produce reports. 

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Added docstrings and extensive documentation to the Evaluate README. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? TODO


bcipy/signal/evaluate/artifact.py
```diff
@@ -0,0 +1,595 @@
...
+            raw_data: RawData,
+            parameters: dict,
+            device_spec: DeviceSpec,
+            save_path: Optional[str] = None,
+            semi_automatic: bool = False,
+            session_triggers: Optional[Tuple[list, list, list]] = None,
+            percent_bad: float = 50.0,
+            detect_eog: bool = True,
+            eye_channels: Optional[List[str]] = None,
+            detect_voltage: bool = True) -> None:
```
lawhead, line 156: Rather than having a of boolean flag for each artifact type it might be more flexible here to have a single parameter that is a list of ArtifactTypes of interest.

bcipy/signal/evaluate/artifact.py
```diff
@@ -0,0 +1,595 @@
...
+        return labels, percent_dropped
+
+    def label_artifacts(
+            self,
+            detect_voltage: bool = True,
+            detect_eog: bool = True,
+            extra_labels: Optional[Annotations] = None) -> Annotations:
+        """Label the artifacts in the raw data."""
+        # Create an empty annotations object to store all the annotations
+        annotations = mne.Annotations(0, 0, 'start')
```
lawhead, line 234: This class is very large and it might make sense to split up its responsibilities. One option is to introduce the concept of an `ArtifactLabeler` with a subclass for each type (ex. `VoltageLabeler`, `EOGEventLabeler`). Then this method could iterate through the `ArtifactTypes` of interest, matching each one to a labeler, and combine the results.

lawhead (approved):
This will add some really useful functionality! My comments are optional but would make it easier to extend this with more options.

PR #337, opened 2024-08-09
Bci gui rewrite

Carsonthemonkey:
# Overview

Rewrite BCIGui code to be simpler and more maintainable

## Ticket

[GUI Refactor](https://www.pivotaltracker.com/story/show/188031090)
[Intertask Action](https://www.pivotaltracker.com/story/show/187994028)

## Contributions

- Rewrite base GUI class
- Rewrite Experiment registry to extend base gui class
- Create intertask action with new ui code

## Test

- Manual testing (may want to look into Qt Test in the future for automated GUI testing)

## Documentation

- Inline documentation written for GUI methods

## Changelog

- Changelog updated


bcipy/gui/bciui.py
```diff
@@ -0,0 +1,324 @@
...
+
+    def list(self):
+        return [widget.data for widget in self.widgets]
+
+    def list_property(self, prop: str):
+        return [widget.data[prop] for widget in self.widgets]
+
+
+# --- Experiment registry code ---
+from bcipy.helpers.load import load_fields, load_experiments
```
tab-cmd: These imports should be at the top of the file (PEP). They can still be separated with a comment as you have done here if that's helpful.

bcipy/gui/bciui.py
```diff
@@ -0,0 +1,324 @@
+import os
```
tab-cmd, line 1: `gui.py` or `main.py` are possible filenames. This one works though. We should deprecate older files in this PR. If the filename changes, we'll also need to update the Makefile!

bcipy/gui/bcipy_stylesheet.css
```diff
@@ -0,0 +1,45 @@
+/* This stylesheet uses the QSS syntax, but is named as a CSS file to take advantage of IDE CSS tooling */
```
tab-cmd, line 1: I really like this change!

bcipy/gui/bciui.py
```diff
@@ -0,0 +1,324 @@
...
+    QMessageBox,
+)
+from typing import Optional, List
+
+
+class BCIUI(QWidget):
+    contents: QVBoxLayout
+    center_content_vertically: bool = False
+
+    def __init__(self, title="BCIUI", default_width=500, default_height=600):
```
tab-cmd: types/ returns needed

bcipy/gui/bciui.py
```diff
@@ -0,0 +1,324 @@
...
+        super().__init__()
+        self.resize(default_width, default_height)
+        self.setWindowTitle(title)
+        self.contents = QVBoxLayout()
+        self.setLayout(self.contents)
+
+    def app(self):
```
tab-cmd: Needed?

bcipy/gui/bciui.py
tab-cmd: This could also be a separate file.

bcipy/gui/intertask_action.py
```diff
@@ -0,0 +1,57 @@
...
+from PyQt6.QtWidgets import (
+    QApplication,
+    QLabel,
+    QHBoxLayout,
+    QPushButton,
+    QProgressBar,
+)
+
+
+class IntertaskAction(BCIUI):
```
tab-cmd: This should likely be InterTask GUI, the Action itself will need to inherit Task.

bcipy/gui/bciui.py
Carsonthemonkey: Planning to split this into a seperate file before merging. Should I delete the original `ExperimentRegistry` on this branch before merging?

bcipy/gui/bciui.py
Carsonthemonkey: Maybe I should make this an abstract method

Carsonthemonkey: Seems like I can't inherit from both `QWidget` and `ABC` so leaving this in for now

bcipy/gui/bciui.py
tab-cmd: That sounds good to me, let's deprecate it!

bcipy/gui/intertask_gui.py
```diff
@@ -0,0 +1,57 @@
...
+from PyQt6.QtWidgets import (
+    QApplication,
+    QLabel,
+    QHBoxLayout,
+    QPushButton,
+    QProgressBar,
+)
+
+
+class IntertaskGUI(BCIUI):
```
tab-cmd, line 15: Wouldn't this need to be a task to use this in orchestration? Or is this going to be a feature that is either on or off? If that's the case, we may want to add a boolean to protocol/experiment registry to configure this.

PR #340, opened 2024-08-21
Parameter guardrails and cleanup

tab-cmd:
# Overview

Add `editable` boolean field to parameters.json entries, update GUI, and make small adjustments to the parameter form for readability. Refactored file names in Task to have more consistency across modules. This field will eventually be used by the Tasks and Session Orchestrator to prevent any parameter updates if editable is false. 

## Ticket

https://www.pivotaltracker.com/story/show/187727881
https://www.pivotaltracker.com/story/show/188144688

## Contributions

- Bullet points about your contribution here

## Test

Screenshot 2024-08-21 at 1 00 29 PM

## Documentation

## Changelog


bcipy/helpers/parameters.py
```diff
@@ -218,9 +220,10 @@ def check_valid_entry(self, entry_name: str, entry: dict) -> None:
         "fake_data": {
             "value": "true",
             "section": "bci_config",
-            "readableName": "Fake Data Sessions",
+            "name": "Fake Data Sessions",
             "helpTip": "If true, fake data server used",
-            "recommended_values": "",
+            "recommended": "",
+            "editable": "true",
```
lawhead, line 226: I wasn't sure where exactly to leave this feedback but the value for `editable` should just be `true` or `false`, rather than the string values (`"true"`, `"false"`). The `value` field has to be a string, since it might be cast to any python type, but this one doesn't need casting and JSON can represent booleans.

bcipy/task/paradigm/rsvp/copy_phrase.py
```diff
@@ -128,7 +128,7 @@ class RSVPCopyPhraseTask(Task):
         "max_inq_per_series",
         "max_minutes",
         "max_selections",
-        "min_inq_len",
+        "min_inq_per_series",
         "show_feedback",
```
lawhead, line 132: This came up in a discussion in February and I created a different ticket for it: https://www.pivotaltracker.com/n/projects/2460065/stories/187147945.

bcipy/helpers/parameters.py
```diff
@@ -13,9 +13,10 @@ class Parameter(NamedTuple):
     """Represents a single parameter"""
     value: Any
     section: str
-    readableName: str
+    name: str
     helpTip: str
-    recommended_values: list
+    recommended: list
+    editable: bool
```
lawhead, line 19: I think this improves readability but it's kind of a major change. These parameters files tend to get used for various things long after data is collected (training models, directing the simulator in how to parse the raw data, etc.). We may want to provide a script to upgrade existing parameters files or add a compatibility layer which can read the old format.

lawhead (approved):

bcipy/helpers/parameters.py
tab-cmd: I am leaning toward a compatibility layer since this will only be used online and we haven't released a new version in awhile. It can be safely ignored like helpTip in many cases.

bcipy/task/paradigm/rsvp/copy_phrase.py
tab-cmd: We did duplicate this! We went with a different solution, but it still addresses the need for a single parameter.

PR #346, opened 2024-09-12
Bcipy client refactor

tab-cmd:
# Overview

Updated bci_main to work with orchestrator instead of old task execution methods. Update load_users and bci-gui to reflect changes. Refactor logging practices were applicable to support saving in task logs etc. 

## Ticket

https://www.pivotaltracker.com/story/show/187744879
https://www.pivotaltracker.com/story/show/187886760
https://www.pivotaltracker.com/story/show/188151705

## Contributions

- Client Refactor: Update main.py to use orchestrator—Deprecate task initialization from start_task. 
- GUI: update commands that are executed from the GUI. A Task or Experiment may be run from the GUI.
- Logging: Update major BciPy modules to accept a custom logger, allowing the logs to end up in the correct file location even with multiple tasks run in sequence. By default, most will still grab a root logger. 

## Test

- `make test-all`
- `bcipy`: tested command with fake and real data. Additionally, verified both a single Task and Experiment could be executed correctly
- `Time Test`: both matrix and rsvp were time tested. All are within normal limits.

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? In line documentation were updated. README etc. in progress. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? TODO


bcipy/gui/BCInterface.py
```diff
@@ -407,10 +408,16 @@ def start_experiment(self) -> None:
...
+            if self.task != BCInterface.default_text:
+                cmd = (
+                    f'bcipy '
+                    f'-u "{self.user}" -t "{self.task}" -p "{self.parameter_location}"'
+                )
+            else:
+                cmd = (
+                    f'bcipy '
+                    f'-u "{self.user}" -e "{self.experiment}" -p "{self.parameter_location}"'
+                )
```
lawhead, line 420: `subprocess.Popen` can take either a string or a sequence of args. In this case most of these values are the same, so it might make more sense for the conditional to just add the things that are different.

```
cmd = ['bcipy', '-u', self.user, '-p', self.parameter_location]
if self.task != BCInterface.default_text:
    cmd.extend(['-t', self.task])
else:
    cmd.extend(['-e', self.experiment])
```

bcipy/helpers/system_utils.py
```diff
@@ -208,21 +208,21 @@ def configure_logger(
     logfile = os.path.join(save_folder, 'logs', log_name)
 
     # configure it
-    root_logger = logging.getLogger()
-    root_logger.setLevel(log_level)
+    custom_logger = logging.getLogger(name=save_folder)
```
lawhead: I think it's good to specify which logger we are configuring, but since you can define more than one handler per logger, consider making the name of the logger different than the save_folder name, which only affects the FileHandler. We may need an additional parameter for that.

`custom_logger = logging.getLogger(name=logger_name)`

lawhead (approved):
The orchestrator and task-related code looks good. We discussed logging changes that I would like to see. I don't want to block if that needs to happen later.

lawhead:
Thanks for working on the logging code. 👍

PR #349, opened 2024-09-19
Task specific params and Matrix Layout

tab-cmd:
# Overview

Updated parameters.json and downstream code to pull Task-relevant parameters. I've updated the code where applicable and added a layout option for Matrix. 

## Ticket

https://www.pivotaltracker.com/story/show/188172524
https://www.pivotaltracker.com/story/show/188188217

## Contributions

- `parameters.json`: split matrix and rsvp parameters. Add layout.
- `MatrixDisplay`: add a method to set the layout of the grid

## Test

- `make test-all`
- Ran with fake data through the client

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? docstrings and other docs updated. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? partial


bcipy/helpers/symbols.py
```diff
@@ -45,18 +45,18 @@ def qwerty_order(is_txt_stim: bool = True,
 
     row1 = "QWERTYUIOP"
     row2 = f"ASDFGHJKL{backspace}"
-    row3 = f" ZXCV{space}BNM "
-    return f"{row1}{row2}{row3}".index
+    row3 = f"ZXCV{space}BNM"
+    return f"{row1}{row2}{row3}"
```
lawhead: This was being used as a function to select a position on a grid and the spaces were important. For example, if I want to represent the first 10 letters (A through J) on two rows (2x6) with the items on the second row centered, the grid positions would be:

```
. . . . . .
. . . . . .
```
with this arrangement:
```
A  B  C  D  E  F
.  G  H  I  J .
```

I would represent this as:
```python
def my_order():
  row1="ABCDEF"
  row2=" GHIJ "
  return f"{row1}{row2}".index

>>> sort_fn = my_order()
>>> sort_fn("G")
7
```
Note that this is different than:
```python
>>> "ABCDEFGHIJ".index("G")
6
```
So this does need to return a function, not just a list of symbols. This could have been better documented.

bcipy/display/paradigm/matrix/display.py
```diff
@@ -76,16 +78,14 @@ def __init__(self,
...
         self.stimuli_inquiry = []
         self.stimuli_timing = []
         self.stimuli_colors = []
         self.stimuli_font = stimuli.stim_font
 
         assert stimuli.is_txt_stim, "Matrix display is a text only display"
 
-        self.symbol_set = symbol_set or alphabet()
+        self.symbol_set = self.build_symbol_set(stimuli)
         self.sort_order = sort_order or self.symbol_set.index
```
lawhead: As I mentioned elsewhere we do need to separate out the concept of a symbol set from how these symbols are positioned in the grid. I used the term `sort_order` here but that's a bit misleading, since it's not just re-ordering symbols but it's choosing a position on the grid. The `symbol_set` code should not change, it's the `sort_order` (or something better named, like `symbol_positioner`?) that should be set.

bcipy/display/paradigm/matrix/display.py
```diff
@@ -114,10 +114,22 @@ def __init__(self,
...
+    def build_symbol_set(self, stimuli: StimuliProperties) -> List[str]:
+        """Build the symbol set for the display."""
+        # TODO pass space and other symbols as parameters
+        if stimuli.layout == 'ALP':
+            return alphabet()
+        elif stimuli.layout == 'QWERTY':
+            return qwerty_order()
+        elif stimuli.layout == 'FREQ':
+            return frequency_order()
+        else:
```
lawhead, line 130: Would it make sense for `stimuli.layout` to be a `Callable` that could be used directly?

```python
   def init_symbol_positioner(self, stimuli: StimuliProperties) -> List[str]:
        """Build the symbol set for the display."""
        assert stimuli.layout, "Layout is required"
        return stimuli.layout()
```

It would probably be good to separate out the mapping of the string to the function into another place. An enum would be a good option here since it would also provide a discrete set of values.

```python
def alphabet_order(parameters=None, include_path=True):
  return alphabet(parameters, include_path).index


class MatrixLayout(Enum):
    ALP = ('alp', alphabet_order)
    QWERTY = ('qwerty', qwerty_order)
    FREQ = ('freq', frequency_order)

    def __new__(cls, *args):
      """Autoincrements the value of each item added to the enum."""
      value = len(cls.__members__) + 1
      obj = object.__new__(cls)
      obj._value_ = value
      return obj

    def __init__(self, label, fn):
      self.label = label
      self.symbol_positioner = fn
```

Initialized from parameters.json:
```python
layout = MatrixLayout[parameters.get('matrix_keyboard_layout', 'ALP')]
StimuliProperties(layout=layout.symbol_positioner)
```

lawhead (changes requested):

bcipy/helpers/symbols.py
tab-cmd: I see what you're saying. I'll refactor to use sort_order from the stimuli properties. The Matrix Display configuration could use some refactoring. We might remove row/col controls altogether and allow keyboard selections only.

bcipy/display/paradigm/matrix/display.py
tab-cmd: Implementing these concepts is a little confusing right now: a symbol set and an "order" that redefines the alphabet symbol set. What we want with QWERTY is to define positions and order, but the rows/cols parameters also control this, so in effect, it adds some padding values to the grid and enforces the order.  We may want to separate the concepts of Keyboard, Symbol Set, and Ordering. I'm content with the hacky approach I pushed for now since we have other tickets to complete before the next experiment begins, but happy to discuss more before we release 2.0.

lawhead: I agree that it would be useful to have a higher level interface/API for interacting with these items. I like the idea of passing in a Keyboard, for instance. This would encapsulate the concepts of rows, columns, ordering, spacing. However, I think the underlying low level mechanism will be the same.

lawhead (approved):

PR #352, opened 2024-09-27
Orchestrator - Load a JSON of copy phrases to use during execution 

tab-cmd:
# Overview

This PR adds new functionality to load a JSON of copy phrases. I have also fixed several task font bugs and VEPCalibration issues stemming from the PR for separated parameters. This adds an official TaskMode enum to support this feature.

## Ticket

https://www.pivotaltracker.com/story/show/188152705 (slight divergence from the discussed approach, using parameters instead of an experiment) 

## Contributions

- Add support for phrases.json
- Update READMEs
- Bug fix font issues
- Add TaskMode and fix paradigm definitions in Tasks
- Add VEP parameters (failed due to redefinition to support matrix and rsvp separation)
- Add missing unittests
- Move exceptions to root 
- Deprecate add_missing_parameters functionality (for now)

## Test

- make test-all
- demo_orchestrator.py with all Calibrations and Multiple Copy Phrases
     - ensured presentation was as expected
     - ensured parameters.json saved correctly with phrases.json was used

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? README and docstrings updated. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Will do in main orchestrator PR


bcipy/parameters/parameters.json
```diff
@@ -410,31 +410,41 @@
     "type": "float"
   },
   "font": {
-    "value": "Consolas",
```
tab-cmd, line 413: I removed this since it is not available without installing a new font.

bcipy/display/paradigm/vep/display.py
```diff
@@ -336,10 +339,10 @@ def draw_animation(self, stimuli: List[StimProps]) -> None:
         while self.static_clock.getTime() < self.timing_animation:
             for frame in range(frames):
                 self.draw_boxes()
+                self.draw_static()
                 for sym, sti in self.sti.items():
                     sti.pos = animation_paths[sym][frame]
                     sti.draw()
-                self.draw_static()
```
lawhead: The initial reason the draw_static was after the stimuli was to meet performance requirements. Generally the more objects that are drawn the longer it takes. It's possible that with some of the refactoring I did a while back the performance is within spec if static is drawn first, but this would be worth measuring at some point.

bcipy/task/__init__.py
```diff
@@ -1,13 +1,14 @@
...
 
 # Makes the following classes available to the task registry
 from .registry import TaskRegistry
 
 __all__ = [
     'Task',
     'TaskRegistry',
-    'TaskData'
+    'TaskData',
+    "TaskMode"
```
lawhead: Looks like TaskMode is the only one here in double quotes.

bcipy/task/orchestrator/orchestrator.py
```diff
@@ -69,28 +77,71 @@ def __init__(
...
+        """Add a task to the orchestrator"""
         self.tasks.append(task)
         self.task_names.append(task.name)
+        self.ready_to_execute = True
 
     def add_tasks(self, tasks: List[Type[Task]]) -> None:
+        """Add a list of tasks to the orchestrator"""
         for task in tasks:
             self.add_task(task)
+        self.ready_to_execute = True
```
lawhead, line 93: I don't think this `self.ready_to_execute` is needed since this method is doing a self send to a method that already sets this.

bcipy/task/orchestrator/orchestrator.py
```diff
@@ -69,28 +77,71 @@ def __init__(
...
+        {
+            "Phrases": [
+                [string, int],
+                [string, int],
+                ...
+            ]
+        }
+        """
+        # load copy phrases from json file or take the task text if no file is provided
+        if self.parameters.get('copy_phrases_location') != "":
```
lawhead: Empty strings are treated as False so this can be shortened to: 

`if self.parameters.get('copy_phrases_location'):`

lawhead (approved):

bcipy/task/orchestrator/orchestrator.py
tab-cmd: This is to prevent execution without any Tasks defined; before execution has occurred, more tasks can be added via a list of Tasks or a single Task, and then all tasks will be reset. This allows a caller to continue using the orchestrator after completing some protocol. I will rethink this as it would technically constitute a new/modified protocol.

PR #353, opened 2024-10-04
Matrix Inquiry Preview

lawhead:
# Overview

Added Inquiry Preview functionality for Matrix Calibration and Matrix Copy Phrase. Fixed a bug in RSVP Copy Phrase when using Inquiry Preview.

## Ticket

https://www.pivotaltracker.com/story/show/188181571

## Contributions

- Added ButtonPressHandler classes for managing button press interactions. There is a different handler for each mode (press to accept, press to reject, preview only).
- Updated display demos.
- Added Inquiry Preview functionality to Matrix display.
- Initialized Matrix Calibration and Copy Phrase tasks with preview parameters.
- Refactored rsvp calibration to use button press handler classes.

## Test

- Added unit tests for ButtonPressHandlers.
- Added unit tests for Inquiry Preview functionality in RSVP and Matrix displays.

### DIFF
--- a/bcipy/display/components/button_press_handler.py
+++ b/bcipy/display/components/button_press_handler.py
@@ -4,7 +4,6 @@ from typing import List, Optional, Type
 
 from psychopy import core, event
 
-from bcipy.display.main import ButtonPressMode
 from bcipy.helpers.clock import Clock
 from bcipy.helpers.task import get_key_press
 
@@ -107,13 +106,3 @@ class PreviewOnlyButtonPressHandler(ButtonPressHandler):
     def accept_result(self) -> bool:
         """Should the result of a button press be affirmative"""
         return True
-
-
-def get_button_handler_class(
-        mode: ButtonPressMode) -> Type[ButtonPressHandler]:
-    """Get the appropriate handler constructor for the given button press mode."""
-    return {
-        ButtonPressMode.NOTHING: PreviewOnlyButtonPressHandler,
-        ButtonPressMode.ACCEPT: AcceptButtonPressHandler,
-        ButtonPressMode.REJECT: RejectButtonPressHandler
-    }.get(mode)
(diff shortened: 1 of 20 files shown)


bcipy/display/components/button_press_handler.py
```diff
@@ -0,0 +1,109 @@
...
+        Parameters
+        ----------
+            wait_length - maximum number of seconds to wait for a key press
+            key_input - key that we are listening for.
+            clock - clock used to associate the key event with a timestamp
+        """
+        self.max_wait = max_wait
+        self.key_input = key_input
+        self.clock = clock or Clock()
+        self.response: Optional[List] = None
```
tab-cmd, line 30: List[str, float]

bcipy/display/main.py
```diff
@@ -269,6 +280,44 @@ def __init__(
...
+    def button_press_mode(self):
+        """Mode indicated by the inquiry progress method."""
+        return ButtonPressMode(self.preview_inquiry_progress_method)
+
+
+def get_button_handler_class(
+        mode: ButtonPressMode) -> Type[ButtonPressHandler]:
+    """Get the appropriate handler constructor for the given button press mode."""
+    mapping = {
+        ButtonPressMode.NOTHING: PreviewOnlyButtonPressHandler,
```
tab-cmd, line 305: I like this approach!

bcipy/display/paradigm/rsvp/display.py
```diff
@@ -128,6 +135,15 @@ def schedule_to(self,
         self.stimuli_timing = timing or []
         self.stimuli_colors = colors or []
 
+    @property
```
tab-cmd, line 138: This property is also set on the task class?

tab-cmd (approved):

bcipy/display/components/button_press_handler.py
lawhead: I think it would actually need to be something like `Union[List[str], List[int]]`, which isn't quite right either. The problem is that a list is actually not a good representation of this data structure. It would make more sense to be a `Tuple[str, float]`, which we do in some places. I made a cleanup ticket to address this across the code base: https://www.pivotaltracker.com/story/show/188409248.

PR #355, opened 2024-10-19
Model Loading

lawhead:
# Overview

 Changed model loading behavior to prompt the user using a separate file dialog for each model.

## Ticket

https://www.pivotaltracker.com/story/show/188038501

## Contributions

- Updated the ask_filename utility to optionally take a prompt. Also updated to use the PyQt file dialog, rather than the native dialog; this provides consistency between platforms as they don't always display the prompt.
- Added utility functions to the acquisition helper for getting the active device content types based on the configured acq_mode.
- Added load functions for models.

## Test

- Created unit tests for new functions.
- Initialized Copy Phrase task with different `acq_mode` configurations to confirm that the file prompts were working correctly.


bcipy/gui/file_dialog.py
```diff
@@ -25,18 +27,21 @@ def __init__(self):
 
         # The native dialog may prevent the selection from closing after a
         # directory is selected.
-        self.options = QFileDialog.Option(3)
+        self.options = QFileDialog.Option.DontUseNativeDialog
```
lawhead, line 30: Note that this was previously broken because the Option enum changed between pyqt5 and 6.

tab-cmd (commented):
Let's wait on the orchestrator merge; this will change where it needs to be implemented (task vs bci_main). The PR is ready.

lawhead:
> Let's wait on the orchestrator merge; this will change where it needs to be implemented (task vs bci_main). The PR is ready.

The orchestrator changes have been merged into this branch and the model loading code was moved to the task. This PR should be ready for review.

tab-cmd (approved):
It works well for me! Thanks for making the updates after the orchestrator merge.

PR #357, opened 2024-10-25
Report artifact enhancements, README, Python 3.10.6 support

tab-cmd:
# Overview

After meeting with the neurophysiology and clinical team, several changes were needed to support the new BciPy report. This fixes some errors found, sets new defaults, adds a demo, and cleans up some of the documentation. Upgrades MNE.

## Ticket

(https://www.pivotaltracker.com/story/show/188110245)

## Contributions

- `demo_calibration_report.py`: A demo to show how to run the action and allow the report to be generated offline in case of failures in the field
- `bcipy/helpers/stimuli.py`: fix baseline designation 
- `bcipy/helpers/convert.py`: change the default to mV (most common). 
- `bcipy/gui/file_dialog.py`: allow for prompt in loading of directories 
- `bcipy/signal/evaluate/artifact.py`: WIP, change defaults, add more logging
- `bcipy/task/actions.py`: Update the BciPyCalibrationReport to prompt for the protocol directory if not provided. Pass the correct labels to visualize_erp. Add some logging. 
- `setup.py`: to include python 3.10.6. Note this breaks editable installs and some console scripting. However, much of the library would still work as a package. This is expected, even if undesired see [1](https://github.com/python/importlib_metadata/issues/364) and [2](https://github.com/pypa/setuptools/issues/2816). A ticket was created to refactor our builds to support future versions of python [here](https://www.pivotaltracker.com/story/show/188483553). When upgrading to 3.11, we may have to lose support for 3.8. However, other dependencies can then be upgraded (mne, etc.). 

## Test

- `make test-all`
- Verified reports generated with @danielpklee 

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Yes! Updates READMES and typing where applicable. 

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? N/a


bcipy/gui/file_dialog.py
```diff
@@ -15,7 +15,7 @@ class FileDialog(QWidget):
 
     def __init__(self):
         super().__init__()
-        self.title = 'PyQt6 file dialogs - pythonspot.com'
```
lawhead, line 18: Good catch. Looks like that was left over from a copy-paste.

bcipy/signal/evaluate/artifact.py
```diff
@@ -40,17 +42,19 @@ class DefaultArtifactParameters(Enum):
     """
 
     # Voltage
-    PEAK_THRESHOLD = 75e-7
-    PEAK_MIN_DURATION = 0.005
+    PEAK_THRESHOLD = 15e-6
+    PEAK_MIN_DURATION = 0.001
```
lawhead, line 48: It might be useful to have a comment or documentation somewhere about how these values were determined.

lawhead (approved):

bcipy/signal/evaluate/artifact.py
tab-cmd: I will add a comment; before we release 2.0, I am proposing a few enhancements to this module that should help with configuration and use. https://www.pivotaltracker.com/story/show/188483451

lawhead (approved):

PR #362, opened 2024-11-07
BIDS export and refactor modules

tab-cmd:
# Overview

Add BIDS export support and deprecate old BDF/EDF exporting and visualization. Refactor the helpers module into corresponding core and io modules to make the library more straightforward for writing the 2.0 paper. 

## Ticket

https://www.pivotaltracker.com/story/show/186801441

## Contributions

- Refactor Helpers into io and core modules
- Add ConvertFormat & convert_to_bids method for converting BciPy data into a BIDS-compatible file/structure. 

## Test

- `make test-all`
- `python bcipy/io/demo/demo_convert.py`
I confirmed with @danielpklee that the output can be loaded in BrainVision and is in good resolution! The triggers will need some mapping when loading the data. 

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? READMEs updated.


bcipy/io/convert.py
```diff
@@ -0,0 +1,400 @@
...
+    assert norm_units[0] >= -1 and norm_units[0] <= 1, "X coordinate must be between -1 and 1"
+    assert norm_units[1] >= -1 and norm_units[1] <= 1, "Y coordinate must be between -1 and 1"
+
+    # convert PsychoPy norm units to Tobii units
+    tobii_x = (norm_units[0] / 2) + 0.5
+    tobii_y = ((norm_units[1] * -1) / 2) + 0.5
+    return (tobii_x, tobii_y)
+
+
+if __name__ == "__main__":
```
tab-cmd, line 391: TODO remove

bcipy/io/demo/demo_convert.py
```diff
@@ -0,0 +1,61 @@
...
+        help='Path to the directory with raw_data to be converted',
+        required=False)
+
+    args = parser.parse_args()
+
+    path = args.directory
+    if not path:
+        path = load_experimental_data()
+
+    convert_to_bids(
```
tab-cmd, line 54: TODO call method above in a comment at least to demo usage

bcipy/io/convert.py
```diff
@@ -0,0 +1,402 @@
...
+
+def convert_to_bids(
+        data_dir: str,
+        participant_id: str,
+        session_id: str,
+        run_id: str,
+        output_dir: str,
+        task_name: Optional[str] = None,
+        line_frequency: float = 60,
+        format: ConvertFormat = ConvertFormat.BV) -> str:
```
lawhead, line 55: Would it make sense to return a `Path`?

bcipy/io/convert.py
```diff
@@ -0,0 +1,402 @@
...
+    # load the raw data
+    raw_data = load_raw_data(raw_data_file)
+    device_spec = preconfigured_device(raw_data.daq_type)
+    volts = True if device_spec.channel_specs[0].units == 'volts' else False
+
+    # load the parameters
+    parameters = load_json_parameters(parameters_file, value_cast=True)
+    trial_window = parameters.get("trial_window")
+    if trial_window is None:
+        trial_window = (0.0, 0.5)
```
lawhead:
```suggestion
trial_window = parameters.get('trial_window', (0.0, 0.5))
```

bcipy/io/demo/demo_convert.py
```diff
@@ -0,0 +1,71 @@
...
+                            convert_to_bids(
+                                data_dir=os.path.join(directory, participant, experiment, session, run),
+                                participant_id=f'0{participant_counter}',
+                                session_id=f'0{session_counter}',
+                                run_id=run,
+                                output_dir=f'./bids/{experiment}/',
+                                format=format.value
+                            )
+                    session_counter += 1
+        participant_counter += 1
```
lawhead: It might be useful to generalize this as a utility function that accepts a function as an argument. Something like:
```
walk_experiment_data(directory: str, experiment_id: str, fn: Callable[...])
```
where the function accepts arguments for data_dir, participant_id, etc.

bcipy/io/demo/demo_convert.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # uncomment to convert a single session
+    # convert_to_bids(
+    #     data_dir=path,
+    #     participant_id='01',
+    #     session_id='01',
+    #     run_id='01',
+    #     output_dir='./bids/',
+    #     format=ConvertFormat.BV
+    # )
```
lawhead: This single session behavior could be called when a different flag is set.

lawhead (approved):

bcipy/io/demo/demo_convert.py
tab-cmd: I like that idea. I will implement it!

bcipy/io/load.py
```diff
@@ -318,3 +302,297 @@ def fast_scandir(directory_name: str, return_path: bool = True) -> List[str]:
...
+                        experiment_id=experiment_id,
+                        date_time=date_time,
+                        run=run,
+                        task=task
+                    )
+                )
+                run += 1
+
+        self.task_paths = [task.path for task in experiment_task_data]
+        self.tasks = [task.task_name for task in experiment_task_data]
```
lawhead: `self.task_paths` and `self.tasks` could be computed properties.

bcipy/io/load.py
```diff
@@ -318,3 +302,297 @@ def fast_scandir(directory_name: str, return_path: bool = True) -> List[str]:
...
+
+    def __init__(
+            self,
+            data_directory: str,
+            experiment_id: Optional[str] = None,
+            user_id: Optional[str] = None,
+            date: Optional[str] = None,
+            date_time: Optional[str] = None,
+            excluded_tasks: Optional[List[str]] = None,
+            anonymize: bool = False) -> None:
```
lawhead, line 380: This class could use documentation on its parameters. It seems like many of these are used for filtering operations. Maybe they should be renamed with a 'filter' suffix (ex. `date_filter`, `user_id_filter`).

bcipy/io/load.py
```diff
@@ -318,3 +302,297 @@ def fast_scandir(directory_name: str, return_path: bool = True) -> List[str]:
...
+        users = []
+        for user in user_paths:
+            if self.user_id:
+                if self.user_id in user:
+                    users.append(user)
+            else:
+                users.append(user)
+
+        self.user_paths = users
+        self.users = [user.split('/')[-1] for user in users]
```
lawhead: These `if` statements could be collapsed to `if self.user_id and self.user_id in user`. However, I think the whole `for` loop can be skipped if there is no `self.user_id` filter defined.

```
if self.user_id:
    self.user_paths = [user for user in user_paths if self.user_id in user]
else:
    self.user_paths = user_paths
self.users = [user.split('/')[-1] for user in self.user_paths
```
Or alternatively self.users could be a computed property. The same pattern applies to most of the filtering methods in the class.

bcipy/io/load.py
```diff
@@ -318,3 +302,297 @@ def fast_scandir(directory_name: str, return_path: bool = True) -> List[str]:
...
+        self.user_id = user_id
+        self.date = date
+        self.experiment_id = experiment_id.replace('_', '')
+        self.date_time = date_time.replace("_", "").replace("-", "")
+        self.session_id = f'{self.experiment_id}{self.date_time}'
+        self.date_time = date_time
+        self.task = task
+        self.run = str(run)
+        self.path = path
+        self.parameters = self.get_parameters()
```
lawhead, line 340: You might consider whether parameters should always be loaded here. It makes it a heavy object to pass around if it's not being used. It might make sense to have parameters be a computed property that's only loaded when needed.
```
@property
def parameters(self):
    if not self._parameters:
        self._parameters = load_json_parameters(
            f'{self.path}/{DEFAULT_PARAMETERS_FILENAME}',
            value_cast=True)
    return self._parameters

lawhead (approved):

bcipy/io/load.py
tab-cmd: It only needs to grab the task name, but it can be used in the future to persist with some other relevant parameters when bundling. I will consider how to remove/standardize from the file structure.

lawhead (approved):
Nice cleanup!

PR #366, opened 2024-11-15
Fusion model patch for 2.0.1rc4

celikbasak:
# Overview

Offline analysis workflow change, fusion model training-testing added, bugfix in gaze reshaper & gaze model.

## Ticket

- https://www.pivotaltracker.com/story/show/186593908

## Contributions

- Offline analysis workflow updated to include train-test split for fusion analysis if there are multiple active devices.
- `Fusion.py` added for generating model performance measures. 
- Bugfix in GazeReshaper where all NaN values are removed. Further tests addded.
- Unused gaze models removed. Gaze model name updates.
- All models are updated in accordance with scikit-learn estimators.

## Test

- Unit tests to be implemented.

## Documentation

- Model documentation will be updated.

## Changelog

- CHANGELOG.md updated.


bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+    trials_per_inquiry = parameters.get("stim_length")
+    # The task buffer length defines the min time between two inquiries
+    # We use half of that time here to buffer during transforms
+    buffer = int(parameters.get("task_buffer_length") / 2)
+
+    # Get signal filtering information
+    transform_params = parameters.instantiate(ERPTransformParams)
+    downsample_rate = transform_params.down_sampling_rate
+    # static_offset = parameters.get("static_trigger_offset")
+    static_offset = 0.0
```
tab-cmd: We need to pull this from the device spec. 

eeg_static_offset = eeg_device_spec.static_offset

Then use it when reshaping triggers below.

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+        device_type='EYETRACKER',
+        apply_starting_offset=False
+    )
+    ''' Trigger_timing includes PROMPT and excludes FIXATION '''
+
+    target_symbols = [trigger_symbols[idx]
+                      for idx, targetness in enumerate(trigger_targetness_gaze) if targetness == 'prompt']
+    inq_start = trigger_timing_gaze[1::11]  # inquiry start times, exluding prompt and fixation
+
+    symbol_set = alphabet()
```
tab-cmd: We already pass a symbol set, and it makes sense to have it as an arg.

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+    # Size = Inquiries x Channels x Samples
+
+    # Reshaping gaze data:
+    gaze_inquiries_dict, gaze_inquiries_list, _ = gaze_model.reshaper(
+        inq_start_times=inq_start,
+        target_symbols=target_symbols,
+        gaze_data=trajectory_data,
+        sample_rate=gaze_sample_rate,
+        stimulus_duration=flash_time,
+        num_stimuli_per_inquiry=10,
```
tab-cmd: trials_per_inquiry

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+
+    # Channel map can be checked from raw_data.csv file or the devices.json located in the acquisition module
+    # The timestamp column [0] is already excluded.
+    eeg_channels_used = [eeg_channels[i] for i, keep in enumerate(eeg_channel_map) if keep == 1]
+
+    gaze_channels_used = [gaze_channels[i] for i, keep in enumerate(gaze_channel_map) if keep == 1]
+
+    # Define the model object before reshaping the data
+    k_folds = parameters.get("k_folds")
+    eeg_model = PcaRdaKdeModel(k_folds=k_folds)
```
tab-cmd: I'd make these optional arguments that default to the EEG and ET models you have listed here.

def calculate_eeg_gaze_fusion_acc(
        eeg_data,
        gaze_data,
        device_spec_eeg,
        device_spec_gaze,
        symbol_set,
        parameters,
        data_folder,
        eeg_model: SignalModel = PcaRdaKdeModel,
        eye_model: SignalModel = GaussianProcess,
):

Then, here we would:

eeg_model = eeg_model(k_folds=k_folds)
gaze_model = gaze_model()

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+from bcipy.helpers.triggers import TriggerType, trigger_decoder
+from bcipy.preferences import preferences
+from bcipy.signal.model.base_model import SignalModelMetadata
+from bcipy.signal.model.gaussian_mixture import (GaussianProcess)
+from bcipy.signal.model.pca_rda_kde import PcaRdaKdeModel
+from bcipy.signal.process import (ERPTransformParams, extract_eye_info,
+                                  filter_inquiries, get_default_transform)
+
+
+def calculate_eeg_gaze_fusion_acc(
```
tab-cmd: Non-blocking comment: I suggest splitting some of this into other methods (perhaps pull out EEG and ET processing?) It will help us write unittests later.

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+
+        train_data_eeg = eeg_inquiries[:, train_indices, :]
+        test_data_eeg = eeg_inquiries[:, test_indices, :]
+        eeg_inquiry_timing = np.array(eeg_inquiry_timing)
+        train_eeg_inquiry_timing = eeg_inquiry_timing[train_indices]
+        test_eeg_inquiry_timing = eeg_inquiry_timing[test_indices]
+        inquiry_symbols_test = np.array([])
+        for t_i in test_indices:
+            inquiry_symbols_test = np.append(inquiry_symbols_test, inquiry_symbols[t_i * 10:(t_i + 1) * 10])
+        inquiry_symbols_test = inquiry_symbols_test.tolist()
```
tab-cmd: Why x 10? Should this be trials_per_inquiry?

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+                for time_0 in range(inquiry_length):
+                    for time_1 in range(inquiry_length):
+                        l_ind = eye_coord_0 * inquiry_length + time_0
+                        m_ind = eye_coord_1 * inquiry_length + time_1
+                        if np.abs(time_1 - time_0) > time_horizon:
+                            cov_matrix[l_ind, m_ind] = 0
+
+        reshaped_mean = np.mean(reshaped_data, axis=0)
+
+        # eps = 5e+3
```
tab-cmd: ?

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+        erp_train_labels = eeg_inquiry_labels[train_indices].flatten().tolist()
+        # erp_test_labels = eeg_inquiry_labels[test_indices].flatten().tolist()
+        eeg_model.fit(preprocessed_train_eeg, erp_train_labels)
+        eeg_model.metadata = SignalModelMetadata(device_spec=device_spec_eeg,
+                                                 transform=default_transform)
+        # save_model(eeg_model, Path(data_folder, f"model_{eeg_model.auc:0.4f}.pkl"))
+        preferences.signal_model_directory = data_folder
+
+        # extract train and test indices for gaze data:
+        centralized_gaze_data_train = centralized_gaze_data[train_indices]
```
tab-cmd: Should some of this live in the Gaze Model itself?

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,370 @@
...
+
+
+def calculate_eeg_gaze_fusion_acc(
+        eeg_data,
+        gaze_data,
+        device_spec_eeg,
+        device_spec_gaze,
+        symbol_set,
+        parameters,
+        data_folder):
```
tab-cmd: Return statement `-> Tuple[list, list, list]`? We may want the average or max acc returned. `-> Tuple[str, str, str]`

bcipy/signal/model/offline_analysis.py
```diff
@@ -311,8 +309,14 @@ def analyze_gaze(
         symbol_set=alphabet()
     )
 
-    # Extract the data for each target label and each eye separately.
     # Apply preprocessing:
+    inquiry_length = inquiries_list[0].shape[1]  # number of time samples in each inquiry
+    preprocessed_array = np.zeros((len(inquiries_list), 4, inquiry_length))
```
tab-cmd: 4?

bcipy/signal/model/offline_analysis.py
```diff
@@ -369,87 +373,29 @@ def analyze_gaze(
...
         time_horizon = 9
-        # Accuracy vs time horizon
 
         for eye_coord_0 in range(4):
             for eye_coord_1 in range(4):
-                for time_0 in range(180):
-                    for time_1 in range(180):
-                        l_ind = eye_coord_0 * 180 + time_0
-                        m_ind = eye_coord_1 * 180 + time_1
+                for time_0 in range(inquiry_length):
```
tab-cmd, line 383: 👍🏼

bcipy/signal/model/offline_analysis.py
```diff
@@ -528,31 +474,42 @@ def offline_analysis(
...
+        device_spec_gaze = devices_by_name.get(gaze_data.daq_type)
+        symbol_set = alphabet()
+        eeg_acc, gaze_acc, fusion_acc = calculate_eeg_gaze_fusion_acc(
+            eeg_data, gaze_data, device_spec_eeg, device_spec_gaze, symbol_set, parameters, data_folder)
+        log.info(f"EEG Accuracy: {eeg_acc}, Gaze Accuracy: {gaze_acc}, Fusion Accuracy: {fusion_acc}")
+
+    # Ask the user if they want to proceed with full dataset model training
+    if confirm("Would you like to proceed with model training?"):
+        models = []
+        log.info(f"Starting offline analysis for {data_file_paths}")
```
tab-cmd: Non-blocking comment: We could add the data or device types in the confirmation message!

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ Our final release candidate before the official 2.0 release!
     - Offline analysis to support multimodal fusion. Initial release of GazeModel, GazeReshaper, and Gaze Visualization #294
     - Updates to ensure seamless offline analysis for both EEG and Gaze data #305
     - Offline analysis support for EEG and (multiple) gaze models. Updates to support Eye Tracker Evidence class #360
+    - Fusion model analysis and performance metrics support. Bugfixes in gaze model #366
```
tab-cmd: We should make a new section for the patch! 

```
# 2.0.1-rc.4

Patch on final release candidate

## Contributions

- Fusion model analysis and performance metrics support. Bugfixes in gaze model #366

```

tab-cmd (changes requested):
Some more minor changes and questions. Thanks again for getting this through!

bcipy/signal/model/offline_analysis.py
lawhead: Also non-blocking, but it seems like functionality regarding interaction (confirmation) should be separated out from the functionality of training a model.

bcipy/signal/model/evaluate/fusion.py
lawhead: It would also make this more readable to make this a NamedTuple.

bcipy/signal/model/evaluate/fusion.py
celikbasak: I forgot about the arg, thank you!

bcipy/signal/model/evaluate/fusion.py
celikbasak: I didn't specify a default model arg for the fusion function because we haven't defined a model to save yet at this point in `offline_analysis` (and I don't save the models trained in `fusion` in any case). Therefore I'm a bit confused which way it would make sense the most

bcipy/signal/model/evaluate/fusion.py
celikbasak: Does `Tuple[float, float, float]` work?

bcipy/signal/model/offline_analysis.py
celikbasak: It is a built-in size for the dimension of `gaze_data` (left eye x, left eye y, right eye x, right eye y). I will make it more pythonic

bcipy/signal/model/offline_analysis.py
celikbasak: Could you elaborate on that @lawhead? I thought based on the flowchart, we wanted to confirm with the user if they would like to still train and save the models after seeing fusion results with a subset of the training data.

bcipy/signal/model/evaluate/fusion.py
celikbasak: I agree. It also currently only supports one or two gaze models, I will try to write it more generic

bcipy/signal/model/evaluate/fusion.py
celikbasak: If you are asking about the `reshaped_mean`,  it is the sample average of reshaped inquiry data. As for `eps`, that was a small epsilon value added to the covariance matrix, as a workaround to non-invertibility. But I think we solved that problem.

bcipy/signal/model/evaluate/fusion.py
celikbasak: `centralized_gaze_data` is the output of `subtract_mean` function which lies in `gaze_model`. Line 244 only picks a subset from that dataset (i.e. `train_indices`) for quick model training

bcipy/signal/model/offline_analysis.py
```diff
@@ -339,15 +344,15 @@ def analyze_gaze(
...
                 (preprocessed_data[sym].shape[0] *
                  preprocessed_data[sym].shape[2],
                  preprocessed_data[sym].shape[1]))
             model.fit(reshaped_data)
 
-        if model_type == "Centralized":
+        if model_type == "Centralized_GM":
             # Centralize the data using symbol positions and fit a single Gaussian.
             # Load json file.
             with open(f"{data_folder}/stimuli_positions.json", 'r') as params_file:
```
tab-cmd: TODO: Pull from config

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,367 @@
...
+        trigger_path=f"{data_folder}/{TRIGGER_FILENAME}",
+        remove_pre_fixation=False,
+        exclusion=[
+            TriggerType.PREVIEW,
+            TriggerType.EVENT,
+            TriggerType.FIXATION,
+            TriggerType.SYSTEM,
+            TriggerType.OFFSET],
+        device_type='EYETRACKER',
+        apply_starting_offset=False
```
tab-cmd: Why not apply this offset?

bcipy/signal/model/evaluate/fusion.py
```diff
@@ -0,0 +1,367 @@
...
+    )
+    ''' Trigger_timing includes PROMPT and excludes FIXATION '''
+
+    target_symbols = [trigger_symbols[idx]
+                      for idx, targetness in enumerate(trigger_targetness_gaze) if targetness == 'prompt']
+    inq_start = trigger_timing_gaze[1::11]  # inquiry start times, exluding prompt and fixation
```
tab-cmd: Is this a result of inquiry length? Should it be calculated?

bcipy/signal/model/offline_analysis.py
lawhead: This can be a later refactor, but in general these large methods should be split up into smaller re-usable pieces and steps requiring user interaction can be split out. For instance, if we want to script some of this in another context  we do not want to require a user to hit the confirmation button.

bcipy/signal/model/evaluate/fusion.py
celikbasak: Because I need the original timing from the triggers file to match it with csv file. The way I reshape Gaze and EEG are fairly different in that sense. For more clarity please take a look at the following:
1. Getting `inq_start` times, excluding prompt and fixation: https://github.com/CAMBI-tech/BciPy/blob/gaze_model_bugfix/bcipy/signal/model/evaluate/fusion.py#L115 
2. Feeding `inq_start` into reshaper and finding the corresponding indices in the csv file (i.e. `start_times`): https://github.com/CAMBI-tech/BciPy/blob/gaze_model_bugfix/bcipy/helpers/stimuli.py#L291 
3. Using the stimulus durations, calculate the inquiry length and obtain a `stop time`. The values between `start` & `stop` indices become the reshaped inquiry data. https://github.com/CAMBI-tech/BciPy/blob/gaze_model_bugfix/bcipy/helpers/stimuli.py#L315

bcipy/signal/model/evaluate/fusion.py
celikbasak: This is a result of `trials_per_inquiry + 1` indicating inquiry length + the prompt symbol. I will update the line

celikbasak:
Thank you so much for your updates/bug fixes Tab!!

tab-cmd:
@celikbasak This needs documentation (README and docstring) and an integration test. Then, it can be merged! If you can tackle the documentation, I will add some tests.

bcipy/signal/model/gaussian_mixture/gaussian_mixture.py
```diff
@@ -12,43 +12,16 @@
...
-class KernelGP(SignalModel):
-    def __init__(self):
-        reshaper = GazeReshaper()
-
-    def fit(self, training_data: np.ndarray, training_labels: np.ndarray):
-        training_data = np.asarray(training_data)
-
-    def evaluate(self, test_data: np.ndarray, test_labels: np.ndarray):
-        ...
+class GaussianProcess(SignalModel):
```
tab-cmd, line 61: This is missing several important methods. It doesn't seem like it could be used online as-is

PR #368, opened 2024-11-26
Simulator CLI/GUI

lawhead:
# Overview

Created a command line interface and GUI to assist with configuring parameters for a simulation.

## Ticket

https://www.pivotaltracker.com/story/show/188497024

## Contributions

- Added an interactive mode to the existing command line interface. `bcipy-sim -i`.
- Added a GUI to run the simulator: `bcipy-sim-gui`
- Both modes produce similar results but the GUI has better support for selecting data directories from more loosely structured data folders.
- Added an argument to customize the output folder of the simulation.
- Simulator bug fixes

## Test

- Ran all unit tests
- Ran the simulator CLI and GUI code to confirm output as expected.


scripts/python/update_params.py
```diff
@@ -0,0 +1,65 @@
+"""Update parameters files to latest."""
```
tab-cmd, line 1: What versions of BciPy would this support?

setup.py
```diff
@@ -110,6 +110,7 @@ def run(self):
             'bcipy = bcipy.main:bcipy_main',
             'bcipy-erp-viz = bcipy.helpers.visualization:erp',
             'bcipy-sim = bcipy.simulator:main',
+            'bcipy-sim-gui = bcipy.simulator.ui.gui:main',
```
tab-cmd: This could be a flag on `bcipy-sim`? I was considering collapsing some of these scripts in the future.

bcipy/simulator/ui/gui.py
```diff
@@ -0,0 +1,679 @@
...
+        return tree
+
+    def skip_path(self, directory: Path) -> bool:
+        """Should the given directory be skipped"""
+        if directory in self.paths or is_ancestor(directory, self.paths):
+            return False
+        return True
+
+
+class ChooseFileInput(QWidget):
```
tab-cmd, line 116: We have similar code in the `GUI` module for getting files and directories. Could we leverage that here? We will be collapsing some of the GUI code (and likely passing around a single QApplication Instance) to package in the future, which is why I ask.

bcipy/simulator/ui/gui.py
```diff
@@ -0,0 +1,679 @@
...
+
+def init(title='BCI Simulator', size=(750, 600)) -> str:
+    """Set up the GUI components and start the main loop."""
+    app = QApplication(sys.argv)
+    panel = MainPanel(title, size)
+    app.exec()
+    command = panel.cli_output()
+    app.quit()
+    if command:
+        print(command, file=sys.stdout)
```
tab-cmd, line 676: use the logger?

bcipy/simulator/task/copy_phrase.py
```diff
@@ -133,7 +133,7 @@ def compute_device_evidence(
             # This assumes that sampling is independent. Changes to the sampler API are needed if
             # we need to provide the trial context of the last sample.
             sampled_data = sampler.sample_data(current_state)
-            evidence = model.compute_class_probabilities(
+            evidence = model.compute_likelihood_ratio(
```
tab-cmd, line 136: 👍🏼

bcipy/gui/main.py
```diff
@@ -273,8 +273,10 @@ def init_control(self, value) -> QWidget:
         # Default is a text input
         return QLineEdit(value)
 
-    def init_editable(self, value: bool) -> QWidget:
+    def init_editable(self, value: bool) -> Optional[QWidget]:
         "Override. Another checkbox is needed for editable"
+        if value is None:
```
tab-cmd, line 278: Update type to show it can be more than bool

tab-cmd (approved):
Looks good! I haven't pulled it locally to try the new GUI yet, but will soon. We should demo with Dan after the next visit to get some feedback.

tab-cmd:
I think this can go to the base branch `2.0` unless you were hoping to release it with the patch.

bcipy/simulator/ui/gui.py
lawhead: These widgets are a bit different in that they represent just the selection input and not the input, label, checkbox, and separator, all of which would affect the layout. These also have hooks for a change action, which is important for this form, but not the parameters GUI.

I do think our existing widgets could be updated to use these components as the underlying input, which would eliminate the duplicate code between the two. That would be a good refinement and when we do that work these could move to a common location.

bcipy/simulator/ui/gui.py
lawhead: Since this starts executing a subprocess immediately in the console it's confusing to the user if this doesn't print to stdout. Also, this gives users the opportunity to run straight from the command line once the configuration is set, which can be faster than the GUI if you're doing this multiple times with only small variation.

scripts/python/update_params.py
lawhead: I think this works for all versions.

setup.py
lawhead: I refactored to remove the setup script and include a command line flag instead.

PR #369, opened 2024-12-09
BIDS DSI-24, ET Data, and 1020 support

tab-cmd:
# Overview

Cleanup convert methods. Add DSI-24 support. Add 1020 coordinate methods. Add the ability to output eye tracker data. Minor bugfix on install dependencies. 

## Ticket

https://www.pivotaltracker.com/story/show/186801441
https://www.pivotaltracker.com/story/show/188668716 (data shared in Google Drive)

## Contributions

- `convert`: add convert et data to BIDS. Modify the demo to handle older data structures. 
- `load`: cleanup BciPySessionTaskData
- `raw_data`: add 1020 methods
- `requirements.txt` define `PyQt6-sip` to permit 3.8 builds

## Test

- `make test-all`
- Ran the demo_convert on Matrix Multimodal data and uploaded for NEU. 

## Documentation

- Are documentation updates required? In-line, README, or [documentation](https://github.com/BciPy/bcipy.github.io)? Done.

## Changelog

### DIFF
--- a/bcipy/gui/file_dialog.py
+++ b/bcipy/gui/file_dialog.py
@@ -64,7 +64,7 @@ def ask_filename(
         file_types: str = DEFAULT_FILE_TYPES,
         directory: str = "",
         prompt: str = "Select File",
-        strict=False) -> Union[str, BciPyCoreException]:
+        strict: bool = False) -> Union[str, BciPyCoreException]:
     """Prompt for a file using a GUI.
 
     Parameters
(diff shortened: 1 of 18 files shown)


bcipy/gui/file_dialog.py
```diff
@@ -72,7 +74,7 @@ def ask_filename(
 
     Returns
     -------
-    path to file or None if the user cancelled the dialog.
+    path to file or raises an exception if the user cancels the dialog.
```
lawhead, line 81: I don't think we always want an exception to be thrown on a dialog cancel. For instance, in GUI applications, a user might initially want to change a path, but then decide to stick with the original value. This would result in a lot of extra error handling. 

What about using an additional parameter to specify the behavior? It could either be a boolean indicating whether or not an exception should be thrown (ex. `ask_filename(..., strict=True)`), or we could provide a default value like a dict.get (ex. `ask_filename(..., default='')`).

tab-cmd: That's a good point. In most of my use cases, this was the fallback GUI if no path was given. It's helpful to have an error when exiting without a choice in those cases. I'll use the boolean, as you suggest.

lawhead (approved):

PR #370, opened 2024-12-11
Fix toml bug

lawhead:
# Overview

Fix an issue with the toml file caused by an inadvertent backspace prior to commit.


tab-cmd (approved):
Thank you for fixing it!

lawhead:
> Thank you for fixing it!

Yeah, sorry about breaking it.

PR #371, opened 2024-12-17
Simulator Metrics

lawhead:
# Overview

Compile metrics summarizing all of the runs in a simulation.
* Outputs summary_data.json which collects the metrics from each session.json file in the run.
* Computes descriptive statistics for each field and outputs the results to the log files and console.
* Generates box plots for several statistics of interest.
* Outputs the mean duration of the runs.

## Ticket

https://www.pivotaltracker.com/story/show/188335185

## Contributions

- Also updated sim logging to use INFO, rather than DEBUG to clean up the log files.

## Test

- Unit tests
- Ran several simulations and inspected the output.

## Screenshots
Screenshot 2024-12-17 at 1 37 33 PM

Screenshot 2024-12-17 at 1 37 57 PM

### DIFF
--- a/bcipy/simulator/metrics.py
+++ b/bcipy/simulator/metrics.py
@@ -4,7 +4,7 @@ import logging
 from collections import Counter
 from json import load
 from pathlib import Path
-from typing import Dict, List
+from typing import Any, Dict, List
 
 import pandas as pd
 
@@ -18,7 +18,7 @@ SUMMARY_DATA_FILE_NAME = "summary_data.json"
 TASK_SUMMARY_PREFIX = "task_summary__"
 
 
-def add_item(container: Dict[str, List], key: str, value: float):
+def add_item(container: Dict[str, List], key: str, value: Any):
     """Add or append value"""
     if key in container:
         container[key].append(value)
@@ -63,12 +63,11 @@ def session_paths(sim_dir: str) -> List[Path]:
     ]
 
 
-def summarize(sim_dir: str) -> Dict:
+def summarize(sim_dir: str) -> Dict[str, List[Any]]:
     """Summarize all session runs."""
-    combined_metrics = {}
+    combined_metrics: Dict[str, List[Any]] = {}
     for session_path in session_paths(sim_dir):
         with open(session_path, 'r', encoding=DEFAULT_ENCODING) as json_file:
-            print("Loading: " + session_path)
             session_dict = load(json_file)
             add_items(combined_metrics, session_dict)
     return combined_metrics
--- a/bcipy/simulator/metrics.py
+++ b/bcipy/simulator/metrics.py
@@ -144,13 +144,13 @@ def report(sim_dir: str, show_plots: bool = False) -> None:
     """Summarize the data, write as a JSON file, and output a summary to
     the top level log file."""
     summary = summarize(sim_dir)
-    save_json_data(summarize(sim_dir), sim_dir, SUMMARY_DATA_FILE_NAME)
+    save_json_data(summary, sim_dir, SUMMARY_DATA_FILE_NAME)
 
     df = pd.DataFrame(summary)
     log_descriptive_stats(df)
 
     logger.info("Typed:")
-    logger.info(Counter(summary['typed']))
+    logger.info(Counter(summary.get('typed')))
     ave_minutes = round((df['total_seconds'] / 60).mean(), 2)
     logger.info(f"Average duration: {ave_minutes} minutes")
     plot_results(df, save_path=plot_save_path(sim_dir), show=show_plots)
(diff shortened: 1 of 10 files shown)


bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+
+from bcipy.config import DEFAULT_ENCODING
+from bcipy.helpers.acquisition import max_inquiry_duration
+from bcipy.io.load import load_json_parameters
+from bcipy.io.save import save_json_data
+from bcipy.simulator.util.artifact import TOP_LEVEL_LOGGER_NAME
+
+logger = logging.getLogger(TOP_LEVEL_LOGGER_NAME)
+
+SUMMARY_DATA_FILE_NAME = "summary_data.json"
```
tab-cmd: These could go in the top-level config.

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+    """List of paths to the session.json files"""
+    return [
+        Path(run_dir, "session.json")
+        for run_dir in sorted(Path(sim_dir).glob("run*/"))
+    ]
+
+
+def sim_parameters(sim_dir: str) -> Dict[str, Any]:
+    """Simulation parameters"""
+    return load_json_parameters(str(Path(sim_dir, "parameters.json")), True)
```
tab-cmd: Pull default name from config? Return is `Parameters`

bcipy/simulator/util/artifact.py
```diff
@@ -30,14 +30,14 @@ def configure_logger(log_path: str,
     """
```
tab-cmd, line 32: It is not required for this PR, but we should talk about the name and configuration of this module. There are some top-level constants here, too. If we want to separate the simulator, we could implement a simulator/config.py.

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+    """Add all items of interest from the session data."""
+    keys = [
+        'total_number_series', 'total_inquiries', 'total_selections',
+        'inquiries_per_selection'
+    ]
+    for key in keys:
+        add_item(combined_metrics, key, session_dict[key])
+
+    for key in session_dict['task_summary'].keys():
+        if 'switch' not in key and 'rate' not in key:
```
tab-cmd: some context here would be helpful

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+    ]
+
+
+def sim_parameters(sim_dir: str) -> Dict[str, Any]:
+    """Simulation parameters"""
+    return load_json_parameters(str(Path(sim_dir, "parameters.json")), True)
+
+
+def summarize(sim_dir: str) -> Dict[str, List[Any]]:
+    """Summarize all session runs."""
```
tab-cmd: We should document what we expect the data/dir to look like

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+    add_item(
+        combined_metrics, 'total_seconds',
+        calculate_duration(session_dict['total_inquiries'], inquiry_seconds))
+
+
+def session_paths(sim_dir: str) -> List[Path]:
+    """List of paths to the session.json files"""
+    return [
+        Path(run_dir, "session.json")
+        for run_dir in sorted(Path(sim_dir).glob("run*/"))
```
tab-cmd: Is this a dir name we can pull from elsewhere?

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
...
+
+def add_item(container: Dict[str, List], key: str, value: Any):
+    """Add or append value"""
+    if key in container:
+        container[key].append(value)
+    else:
+        container[key] = [value]
+
+
+def get_final_typed(session_dict: Dict) -> str:
```
tab-cmd: What do you expect session dict to look like?

bcipy/simulator/metrics.py
```diff
@@ -0,0 +1,160 @@
+"""Summarize metrics across simulator runs."""
```
tab-cmd: This could go in data or util?

tab-cmd (approved):
Great addition! The README needs to be updated a bit to reflect the change in expected output. 

Suggestions:
What do you think about having a run-name argument? We could use it in the sim dir name and/or metric png title to help with multiple comparisons. 

When outputting the command generated from the GUI, we should add quotations around string paths to prevent issues with spaces in dir names.

bcipy/simulator/metrics.py
lawhead: Yes, I can add that to the artifact module.

bcipy/simulator/util/artifact.py
lawhead: Sounds good. Let's discuss.

PR #374, opened 2025-01-21
Inquiry Range Sampler

lawhead:
# Overview

Added a Sampler for filtering the data by inquiry range and series range.

## Ticket

https://www.pivotaltracker.com/story/show/188673150

## Contributions

- Refactored samplers into their own directory.
- New sampler for filtering data.
- Added sampler arguments to the Task Factory so users could configure the parameters for a Sampler.
- Command line support to the simulator to accept sampler and sampler args parameters.
- Added a sampler pulldown menu to the simulator GUI.
- Added a new GUI component to generate input fields for a given object and used it on the simulator GUI to dynamically configure the selected sampler.

## Test

- Created and ran unit tests.
- Ran the simulator using various configurations of samplers and inputs to confirm that the correct values were being used.

### DIFF
--- a/bcipy/simulator/README.md
+++ b/bcipy/simulator/README.md
@@ -43,6 +43,7 @@ For example,
 - `m`: Path to a pickled (.pkl) signal model. One or more models can be provided.
 - `n`: Number of simulation runs
 - `o`: Output directory for all simulation artifacts.
+- `s`: Sampling strategy to use; by default the TargetNonTargetSampler is used. The value provided should be the class name of a Sampler.
 
 #### Sim Output Details
 
--- a/bcipy/simulator/README.md
+++ b/bcipy/simulator/README.md
@@ -25,6 +25,8 @@ optional arguments:
   -n N                  Number of times to run the simulation
   -s SAMPLER, --sampler SAMPLER
                         Sampling strategy
+  --sampler_args SAMPLER_ARGS
+                        Sampler args structured as a JSON string.
   -o OUTPUT, --output OUTPUT
                         Sim output path
 ```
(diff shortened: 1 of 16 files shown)


bcipy/simulator/task/task_runner.py
```diff
@@ -126,12 +131,15 @@ def main():
         task_factory = TaskFactory(params_path=sim_args['parameters'],
                                    source_dirs=sim_args['data_folder'],
                                    signal_model_paths=sim_args['model_path'],
-                                   sampling_strategy=classify(sim_args['sampler']),
-                                   task=SimulatorCopyPhraseTask)
+                                   sampling_strategy=classify(
+                                       sim_args['sampler']),
+                                   task=SimulatorCopyPhraseTask,
+                                   sampler_args=json.loads(sim_args['sampler_args']))
```
tab-cmd: I'd make this a method. We could add some parameter validation or helpful logging to it.

bcipy/simulator/README.md
```diff
@@ -43,6 +45,8 @@ For example,
 - `m`: Path to a pickled (.pkl) signal model. One or more models can be provided.
 - `n`: Number of simulation runs
 - `o`: Output directory for all simulation artifacts.
+- `s`: Sampling strategy to use; by default the TargetNonTargetSampler is used. The value provided should be the class name of a Sampler.
+- `sampler_args`: Arguments to pass in to the selected Sampler. Some samplers can be customized with further parameters. These should be structured as a JSON dictionary mapping keys to values. For example: `--sampler_args='{"inquiry_end": 4}'`
```
tab-cmd, line 49: Is this the only parameter that could change in sampling? It may be helpful to have a `custom_args=` that could be used to modify other parameters in the future, too.

tab-cmd (approved):
Codacy made a few minor suggestions (typing, etc.) that I agree with; otherwise, it looks great. I'm excited to try it with the group later today.

bcipy/simulator/README.md
lawhead: Any of the sampler arguments can be provided. You can inspect the output from the GUI to see what this looks like. 

In regards to making this more generic, I would like to keep this pretty specific to samplers for now and see how things evolve. I think if there are a number of simulator objects that require more configurability we may want a different solution (such as a sim_config file).

lawhead:
I ran mypy on the gui modules and fixed all the issues, which included the things caught by codacity and a number of other issues. I also refactored the parameters for the InquiryRangeSampler to make it clearer to use.

PR #375, opened 2025-02-06
Simulator Verbose Mode

lawhead:
# Overview

Added simulator flag to control the verbosity of the output.

## Ticket

https://www.pivotaltracker.com/story/show/188328498

## Contributions

- Changed log level for many statements.
- Added a verbose parameter used to control the log level.
- Deleted generated triggers.txt

## Test

- Ran the simulator using both modes and compared the outputs

## Discussion

These changes still output all of the files, but the file contents are much shorter. I opted to keep log files for each run in case any errors were encountered. If we did not have logs at all then these would be missed.

The largest output files are the individual session.json files for each run. These are used for generating the overall metrics. If the simulation artifacts are still too large we can do some further refactoring to accumulate metrics values as we go and delete each run's session.json after that run's data has been accumulated.

### DIFF
--- a/bcipy/helpers/copy_phrase_wrapper.py
+++ b/bcipy/helpers/copy_phrase_wrapper.py
@@ -6,10 +6,10 @@ from typing import List, Tuple
 import numpy as np
 
 from bcipy.config import SESSION_LOG_FILENAME
-from bcipy.exceptions import BciPyCoreException
-from bcipy.helpers.language_model import histogram, with_min_prob
 from bcipy.core.stimuli import InquirySchedule, StimuliOrder
 from bcipy.core.symbols import BACKSPACE_CHAR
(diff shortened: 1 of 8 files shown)


tab-cmd (approved):
This works well for me, and I like the processing indicator, too. Do we need the XLSX files? Would it be easy to prevent them from exporting?

lawhead:
The xlsx file output can be configured in the parameters.json file by setting "summarize_session" to false.

tab-cmd:
> The xlsx file output can be configured in the parameters.json file by setting "summarize_session" to false.

That may be a parameter we override in a simulation. I don't know if it would be helpful unless we wanted more verbose output. This is not a blocking comment. Since there is a workaround, let's get feedback from the group and address it later.

PR #376, opened 2025-02-12
Simulator Replay Script

lawhead:
# Overview

Integrate replay script into simulator

## Ticket

https://www.pivotaltracker.com/story/show/188805610

## Contributions

- Refactored the replay_session script to work with the simulator framework.
- Created a ReplayTask to iterate through a previously recorded session.
- Created a ReplaySampler
- Added a module to compute the metrics specific to this

## Test

- Added unit tests
- Ran the replay session with a single data directory and with multiple.


bcipy/simulator/README.md
```diff
@@ -102,6 +102,28 @@ A simulation can be started using a graphical user interface.
...
 This provides a way to explore the file system when providing the parameters.json file, simulation model, and input data sources. After all required inputs have been provided the user can initiate the simulation from the GUI interface. The command line used to run the simulation are output to the console prior to the run making it easier to start subsequent simulations with the same set of arguments.
 
+## Replay Session
+
+The simulator also includes a Task for replaying a recorded session using a different signal model. This is used for testing if changes to a model result in more easily differentiated signals.
+
+This functionality currently has a different entry point.
+
+```
+(venv) $ python bcipy/simulator/task/replay_session.py -h
```
tab-cmd: What do you think about adding it to the console scripts? 

```
bcipy-replay 
```

Or it could use the one registered for sim with a different flag.

```
bcipy-sim --replay 
```

My concern is that for users who install BciPy (not clone) via pip, the path may be more challenging to access than a CLI interface.

tab-cmd (approved):
Thanks for integrating this!

bcipy/simulator/README.md
lawhead: I agree, I do think we need a better way of running it. I'm still considering whether I can add it to the GUI, but I think that would be a separate PR. Maybe the quicker fix for this PR would be your second suggestion. I'll work on that.

PR #377, opened 2025-02-24
Causal LM Integration 

tab-cmd:
# Overview

This PR adds the Causal language model and provides a demo for running simulations on a data set with multiple users.

## Contributions

- remove broken integration tests and poorly defined decomposition test.
- @dcgaines added causal language model to `language.model.causal`
- Updated simulator READE.md with instructions on how to run the group processing demo + markdown linting
- Add `NullDAQ` and `close` methods to `NullDisplay.` This was important when trying to run different modes in simulation.
- Added the ability to pass in a full parameter set to `TaskFactory`
- Added the ability to provide a prefix to `init_simulation_dir`. This allows for multiple runs with different parameters to be distinguished. 

## Test

- Used the demo code to publish results for the BCI Meeting 2025
- `Make test-all`

## Documentation

- Are documentation updates required?  Yes

## Changelog

- Is the CHANGELOG.md updated with your detailed changes? Not yet.


dcgaines (approved):
Looks good on the LM side!

bcipy/simulator/task/copy_phrase.py
```diff
@@ -95,6 +98,10 @@ def init_evidence_types(
     def init_display(self) -> Display:
         return NullDisplay()
 
+    def init_acquisition(self) -> ClientManager:
+        """Override to do nothing"""
```
lawhead, line 102: Technically I don't think the parent class has this method.

bcipy/simulator/task/task_factory.py
```diff
@@ -32,13 +32,15 @@ def update_latest_params(parameters: Parameters) -> None:
...
-                 sampler_args: Optional[Dict[str, Any]] = None):
+    def __init__(
+            self,
+            params_path: str,
+            source_dirs: List[str],
+            signal_model_paths: List[str],
+            sampling_strategy: Type[Sampler] = TargetNontargetSampler,
+            task: Type[SimulatorCopyPhraseTask] = SimulatorCopyPhraseTask,
+            parameters: Optional[Parameters] = None,
+            sampler_args: Optional[Dict[str, Any]] = None):
```
lawhead, line 60: I think it's confusing supporting both a `params_path` and a `parameters` object. We should only do one of these. Since we now have a use case that the parameters object may be manipulated before being passed in, we should just use that and not worry about `params_path`. All the code that's currently using `params_path` can be updated to `load_json_parameters` before passing it in.

lawhead (approved):
Seems like there are some Codacy type checks that we may want to address. Otherwise, just the change to the TaskFactory parameters and I think this is good to merge.

bcipy/simulator/task/copy_phrase.py
tab-cmd: This method was only used to set the NullDAQ. I cannot recall why this was needed. I'm addressing the other point and will test it; I believe there were some calls to it somewhere, and I was getting a None has no attribute x type error.

PR #385, opened 2025-03-17
Simulator multimodal

lawhead:
# Overview

Proof of concept to demonstrate simulator support for multimodal. 

## Ticket

https://www.pivotaltracker.com/story/show/188263392

## Contributions

- Created a demo module in the simulator with a detailed README on how to use different aspects of the simulator, including multimodal
- Added a sample Button Press model, as well as a corresponding data processor.
- Refactored data processor code to include a registry to make it easier to add support for new processors.

## Test

- Ran the simulator with multiple models as described in the demo README. Confirmed that the results fused evidence from both models.
- Ran all unit tests.


lawhead:
We may need to look into our Codacity settings. The duplication in unit tests is intentional for better readability and understanding. Also, the security concern on using random.uniform is bogus.

bcipy/simulator/data/processor_registry.py
```diff
@@ -0,0 +1,75 @@
...
+# flake8: noqa
+from bcipy.simulator.data.data_process import (EegRawDataProcessor,
+                                               RawDataProcessor)
+from bcipy.simulator.demo.button_press_data_processor import \
+    ButtonPressDataProcessor
+from bcipy.task.data import EvidenceType
+
+log = logger.getLogger(__name__)
+
+PROCESSOR_TYPES = [EegRawDataProcessor, ButtonPressDataProcessor]
```
tab-cmd: I prefer the PEP8 recommendation of `EEGDataProcessor` when handling acronyms. I'm not sure the raw is needed when examining the emerging pattern, either.

bcipy/parameters/devices.json
```diff
@@ -120,5 +120,17 @@
         "excluded_from_analysis": ["device_ts", "system_ts", "left_pupil", "right_pupil"],
         "status": "active",
         "static_offset": 0.0
+    },
+    {
+        "name": "Switch",
+        "content_type": "MARKERS",
```
tab-cmd, line 126: I had missed this before, but should this be `MARKER`? We would register one switch or marker stream at a time, similar to the other devices. 

If this is not a continuous stream, like would be the case for LSL markers, I'm not sure it makes sense to be registered here. If it were, we would like to know the origin (ex. psychopy_switch, lsl_marker_stream) to help understand how to register any changes configured here.

bcipy/simulator/demo/button_press_data_processor.py
```diff
@@ -0,0 +1,139 @@
...
+                                               RawDataProcessor, ReshapedData,
+                                               TimingParams)
+from bcipy.simulator.demo.button_press_utils import (should_press_button,
+                                                     simulate_raw_data,
+                                                     switch_device)
+from bcipy.simulator.exceptions import IncompatibleParameters
+from bcipy.task.data import EvidenceType
+
+
+class ButtonPressDataProcessor(RawDataProcessor):
```
tab-cmd: This could perhaps be split into a generator and processor. We do have button press data that could be loaded or mocked.

bcipy/simulator/demo/button_press_model.py
```diff
@@ -0,0 +1,78 @@
+"""Signal model for button presses."""
+from pathlib import Path
+from typing import List
```
tab-cmd: This can go to the signal module. I think this is a nicer implementation of what's happening than simply using the `compute_probs_after_preview`

bcipy/simulator/demo/README.md
```diff
@@ -0,0 +1,73 @@
+# Simulator Demo
```
tab-cmd: Given the patterns of other submodules, I would prefer a `demo` submodule to show how to use existing functionality. These methods can be rolled into other location safely and many would be quite useful.

bcipy/simulator/demo/button_press_data_processor.py
tab-cmd: As well, it can be moved to the existing data_process module

bcipy/simulator/demo/button_press_utils.py
```diff
@@ -0,0 +1,119 @@
...
+from bcipy import config
+from bcipy.acquisition.datastream.mock.switch import switch_device
+from bcipy.core.parameters import Parameters
+from bcipy.core.raw_data import RawDataWriter
+from bcipy.core.triggers import Trigger, TriggerType, trigger_decoder
+from bcipy.display.main import ButtonPressMode, PreviewParams
+from bcipy.helpers.acquisition import raw_data_filename
+
+
+def pairwise(iterable):
```
tab-cmd: These can be moved to `simulator/util`

bcipy/simulator/tests/test_button_press_model.py
```diff
@@ -0,0 +1,71 @@
+"""Tests for Button Press Model"""
```
tab-cmd: We can do this in another PR, but there are enough tests for this module now that we should follow the pattern set elsewhere.

`tests/*submodule_name*/test_*name_of_file_tested*.py`

bcipy/simulator/demo/README.md
```diff
@@ -0,0 +1,73 @@
...
+For inquiries in which the target not shown:
+
+- evidence values for symbols in inquiry should be degraded
+
+Note that the progress method (`preview_inquiry_progress_method` parameter) doesn't matter if it is set to "press to accept" or "press to skip", since the ButtonDataProcessor interprets this and outputs a 1.0 for inquiries that should be supported and 0.0 for those that shouldn't.
+
+### Limitations
+
+- A `preview_inquiry_progress_method` of 0 is currently not supported and an exception will be thrown. Ideally, all inquiries should get an evidence value of 1.0 (no change) with this mode.
+- Button evidence only works correctly with the InquirySampler. This is due to all trials in the same inquiry receiving the same value.
```
tab-cmd: We could try loading / processing from an actual data file for this. The inquiry preview and SCRD studies have this data available from the triggers.txt and session.json.

tab-cmd (approved):
I'm hesitant to approve this one, although I know some extra was needed to fully test the multimodal functionality. This PR gets us closer to a workable "Button" or "Switch" integration, but I think the flipping between them in name and configuration overall is a bit off. Using a SignalModel was a really clever idea. I'd advocate for moving it into the signal module. It could (and likely should) be used outside of the simulator for other purposes. Please consider the comments left and let's discuss either here or in Discord.

tab-cmd:
> We may need to look into our Codacity settings. The duplication in unit tests is intentional for better readability and understanding. Also, the security concern on using random.uniform is bogus.

We can bypass the check and mute others. Are you able to log in to bypass this check? If not, I will make sure you can in the future! 

 ```The Standard pseudo-random generators are not suitable for security/cryptographic purposes. (B311)``` is correct in a literal sense. It is not entirely random, as we occasionally saw occur in other tests of randomness by the default Python generators in this codebase. There is no harm where you use it, though, and this is a limitation with static code analysis in general (not so smart), so the warning can be safely ignored. It should be a warning imo, though. We have discussed using Monte Carlo and bootstrapping algorithms, which this warning would be correct to catch.

lawhead:
As per our offline discussion, I made a number of refactoring changes to more formally develop the `SwitchModel` and allow it to be used in a Copy Phrase task, rather than just in a simulation.

- Renamed Button classes to Switch for consistency.
- Moved model to signal module.
- Created a `SwitchEvidenceEvaluator` for use within tasks.
- Several updates to the acquisition module (`lsl_client`) to correctly handle irregular data streams.

For the simulator:

- Updated the `SwitchDataProcessor` to use raw marker data.
- Created a script for generating marker data from Inquiry Preview triggers, as well as being able to mock the data for Copy Phrase sessions that did not use Inquiry Preview.
- Moved Switch-related functionality out of the demo folder. Added multimodal section to the main simulator README.

Testing
- Confirmed that all unit tests passed.
- Ran a Copy Phrase task with acq_mode set to 'EEG+MARKERS'. The EEG data was streamed using a mock lsl server and the switch data was streamed from the mock GUI switch in bcipy.acquisition.mock.switch. Confirmed multimodal data acquisition and fusion.

bcipy/task/tests/core/test_switch_evaluator.py
```diff
@@ -0,0 +1,61 @@
+"""Tests for EEG evidence evaluator"""
```
tab-cmd: cleanup comment

bcipy/simulator/data/processor_registry.py
```diff
@@ -0,0 +1,74 @@
+"""Registry for data processors."""
+import logging as logger
+from typing import List, Optional, Type
+
+from bcipy.acquisition.multimodal import ContentType
+from bcipy.signal.model.base_model import SignalModel
+# flake8: noqa
+from bcipy.simulator.data.data_process import (EegRawDataProcessor,
```
tab-cmd: `EEGRawDataProcessor`

bcipy/simulator/README.md
```diff
@@ -140,3 +141,93 @@ python bcipy/simulator/demo/demo_group_simulation.py
 ```
 
 See the demo file for more details on how to configure the simulation.
+
+## Multimodal
```
tab-cmd, line 145: Nice documentation 👍🏼

tab-cmd (approved):

bcipy/simulator/data/processor_registry.py
lawhead: Thanks, I forgot to do that.

PR #386, opened 2025-03-20
Support multi run experiments and additional 10-20 channels

tab-cmd:
# Overview

Update to allow for multirun BIDS conversion, add 10-20 channels, add BIDS to MNE feature with demo, and update the macos builds.

## Contributions

- Add 10-20 channels to `raw_data.py`. Update the tests.
- `BIDS_to_MNE` method added to `convert.py`. This allows loading of entire BIDS collections for processing.
- Updated the demo to show how to package multirun experiments. Add a load bids demo with some example plotting / processing.
- Update m2_install and workflow files to fix mac os build issue.

## Test

- `make test-all`
- loaded converted BIDS data using the demo

## Documentation

- Documentation added where needed

## Changelog


.github/workflows/main.yml
```diff
@@ -99,7 +99,7 @@ jobs:
 
   build-macos:
 
-    runs-on: macos-latest
+    runs-on: macos-14
```
lawhead, line 102: While 'latest' can help us detect issues earlier, I think pinning all of our CI tests to versions might be a better fit moving forward so we don't get random breakage when we're not ready to deal with it.

lawhead (approved):

PR #390, opened 2025-04-22
LM Toolkit Refinements

lawhead:
# Overview

Refactored Language Model protocols; fixed type issues, linting, and import sort order.

## Discussion

Our protocols were not correct. It was possible to create a LanguageModel just by implementing the `set_symbol_set` method, which was not the intention. I changed the hierarchy here to make `LanguageModel` a Union type of `CharacterLanguageModel` and `WordLanguageModel`. This also necessitated changes to the types in the helper module.

I also removed the symbol_set property from the Protocol. That property is an implementation detail since it's not used by any code that uses a Language Model. Also, it potentially introduces a subtle bug since it's possible to set that attribute through either the property or the `set_symbol_set` method.

### DIFF
--- a/bcipy/exceptions.py
+++ b/bcipy/exceptions.py
@@ -1,4 +1,3 @@
-
 class BciPyCoreException(Exception):
     """BciPy Core Exception.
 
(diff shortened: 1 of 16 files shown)


dcgaines (approved):
Looks good to me, thanks for putting this together!

